PR #74, opened 2020-10-08
Fix Storybook incompatibility with Tailwindcss (fixes #62)

### LINKED ISSUE #62: Storybook doesn't render components with styling

We have a single story for the schema viewer, but it doesn't look great:

image

This is because the Tailwind css components aren't being imported directly into the story. It should be possible to fix this with some configuration changes. This repository may have some hints on how to configure things (my suggestion would be to look at the postcss configuration):

https://github.com/jerriclynsjohn/svelte-storybook-tailwind

Iinh:
This PR fixes issue #62 by adding the correct Tailwind CSS components, using https://github.com/jerriclynsjohn/svelte-storybook-tailwind as template.

Screen Shot 2020-10-08 at 7 32 16 AM


wlach (changes requested):
Hi @linh! Thank you, this looks much better. However it looks like the heading isn't getting styled as I'd expect (it should be larger). I think this is because the styles in https://github.com/mozilla/glean-dictionary/blob/main/src/Tailwindcss.svelte aren't getting imported. Could you see if there's a way for them to be imported by default for all stories? You'll see that the main application includes them like this:

https://github.com/mozilla/glean-dictionary/blob/5e787003843d662079e8c047078c108de8b6d976/src/App.svelte#L36

Iinh:
**What I’ve tried**

I imported Tailwind to the story file itself to use as a styling wrapper with `decorators`. However, the problem is custom decorator syntax requires a div with literal styling that wraps around the component, like so:

```
export const Basic = () => {
  Component: SchemaViewer,
  decorators:  [(Story) => <div style={{ margin: '3em' }}><Story/></div>]
};
```

`Tailwindcss.svelt` is a Svelte component, so it can’t be properly used inside the callback function. Something like this wouldn’t work:

```
Import Tailwindcss from “../Tailwind”
export const Basic = () => {
  Component: SchemaViewer,
  decorators:  [(Story) => <div style={{ Tailwindcss }}><Story/></div>]
};
```

**My solution**

Currently, Tailwindcss is compiled to a separate CSS file at run time. The output of this file, according to the rollup configuration, is at `/public/build/bundle.css`. Since this CSS file always has to be compiled to be used in the main app, I thought why not use it for Storybook too. This also ensures that any additional styling, once processed, will apply to all stories. Attached image shows how the story looks with this solution. 

Looking forward to feedback. Thank you!

story

wlach:
@linh Would this not work?

```svelte
Import Tailwindcss from “../Tailwind”
export const Basic = () => {
  Component: SchemaViewer,
  decorators:  [(Story) => <div><Tailwindcss /><Story/></div>]
};
```

Using the bundle css directly might be ok, though I'd worry about it being kept up to date (I'm not sure if storybook will recompile it when it changes?). 

Regardless, thanks for your persistence on this!

Iinh:
@wlach It'd break the storybook file, at least on my local machine. Also Eslint rejected it. I've tried different variations of the above syntax, even what you suggested, but nothing worked: they all broke the file. Are you able to use it locally? If yes then there is something wrong with my linter.

I agree that the caveat to using bundle css is that it needs to be compiled with `npm run dev` first for storybook to get the changes (although we can write a [post-css](https://github.com/postcss/postcss-cli) script to exclusively watch any changes to tailwind if you prefer not using `npm run dev`)

I appreciate your time @wlach!

wlach:
> Hi @wlach, your code would cause an error in storybook. I believe because you used JSX (`<div><Tailwindcss/><Story/></div>`) inside decorators, and JSX syntax is invalid in Javascript without a compiler. That’s why the code sample used in the document for decorators would not work in our case because they all use JSX. However it would work in a React application.

You are of course totally right on this. Apologies for the misdirection.

> Set up Postcss configuration inside storybook so we don’t have to run `npm run dev` and rely on development build for Tailwind processing, something similar to [this](https://github.com/jerriclynsjohn/svelte-storybook-tailwind/tree/master/.storybook).

I think this sounds like the way to go! The main piece of functionality we're looking for here are the storybook components reloading when people modify the top-level tailwind component. If that works, I think we're good to merge.

Thanks so much for your persistence on this.

Iinh:
This update is looking to fix 2 problems.
### 1) Tailwind incompatibility with Storybook
Currently Tailwind doesn't work properly in Storybook. This is fixed by [updating the webpack configuration](https://storybook.js.org/docs/svelte/configure/webpack) to include `svelte-preprocess`. However this fix alone doesn't completely resolve the issue of storybook not rendering the top-level tailwind component because of the problem below.

### 2) Global styles not available globally

`Tailwindcss.svelte` has all css global rules, including the Tailwind utilities packages. The problem with this file is it's placed inside `App.svelte`, which means that global styles only get compiled and become available for other components after App is called. This is fine in development as App is always the first one called. However, in storybook, each component is called in isolation. For example, when `SchemaViewer` loads in Storybook the only styling available is from SchemaViewer itself. If App was never called, SchemaViewer can't inherit any global styles. 

To fix this problem, we need to keep global styles in a different file. 3 main requirements:

- It can’t be a svelte component because we can’t use component syntax for styling in storybook 
- It needs to be accessible to all components: it can’t belong to just one component, like how inaccessible Tailwindcss.svelte was because it's only available through App
- Changes to this file need to reload the components, both in development and storybook

I believe the best option is to replace Tailwindcss.svelte with a css file for global styles, because:  
-  A css file can be easily imported to use with storybook. Any changes will directly reload the stories
- In development: `svelte-preprocess` doesn’t compile css files outside of Svelte components, we can use [postcss-cli](https://github.com/postcss/postcss-cli) to compile it instead. This css file writes to `public/build/global.css`, which is then used in `index.html` -- global styles will apply to all components. An npm script was added to watch this css file so that it'll recompile if changes are added. 

With this change, all stories will independently inherit global styles. We also don't have to rely on development build as the css file is imported directly to storybook.

Looking forward to feedback @wlach. Thank you!

package.json
```diff
@@ -4,7 +4,7 @@
   "scripts": {
     "build": "rollup --config",
     "build-storybook": "build-storybook --static-dir ./public --config-dir .storybook",
-    "dev": "rollup --config --watch",
+    "dev": "rollup --config --watch & postcss src/tailwind.css -o public/build/global.css -w",
```
wlach, line 7: Should not be necessary (see review)

Iinh:
Hi @wlach thank you for your feedback! I've updated my code with the hybrid approach 😄 Looks like CircleCI failed my PR because of some code format error in `Tailwindcss.svelte`. I tried changing it around but it didn't seem to work. Locally my linter is not complaining anything so it's not clear what I need to fix either, I probably need another pair of eyes to help find this format error. What usually is your approach to solve issues like this? Thank you.

wlach:
@linh try running `npm run format`! It should autofix things for you

Iinh:
@wlach YES that fixed it, thank you so much! I learned so many new things this week. This PR is ready for review.

wlach (approved):
Awesome!

PR #75, opened 2020-10-08
Add filter box to apps page (fixes #73)

### LINKED ISSUE #73: Add a filter box to the applications page

If you look at the bigquery table view, you'll see that there's a filter box that lets you easily search for the subset of columns that you're interested in:

image

We should have a similar widget on the main page to filter through the list of applications where e.g. putting "firefox" into the box will only show applications with Firefox in the name.

DawoodShahat:
This PR fixes issue #73 Where Search Box is added to the main Application's Page

Screenshot 2020-10-08 at 8 26 07 PM

### DIFF
--- a/src/components/FilterBox.svelte
+++ b/src/components/FilterBox.svelte
@@ -4,7 +4,7 @@
   export let onChangeText = null;
   let filterText = "";
 
-  const changeText = (e) => {
+  const changeText = () => {
     onChangeText(filterText);
   };
 </script>
--- a/src/components/FilterBox.svelte
+++ b/src/components/FilterBox.svelte
@@ -1,0 +1,20 @@
+<script>
+  import throttle from "just-throttle";
+
+  export let onChangeText = null;
+  let filterText = "";
+
+  const changeText = (e) => {
+    onChangeText(filterText);
+  };
+</script>
+
+<div class="container py-4 mx-auto">
+  <input
+    class="shadow appearance-none border rounded w-full p-2 text-gray-700
+    leading-tight focus:outline-none focus:shadow-outline"
+    type="text"
+    bind:value={filterText}
+    on:input={throttle(changeText, 200)}
+    placeholder="filter terms" />
+</div>
--- a/src/components/SchemaViewer.svelte
+++ b/src/components/SchemaViewer.svelte
@@ -1,12 +1,11 @@
 <script>
-  import throttle from "just-throttle";
   import SchemaNode from "./SchemaNode.svelte";
+  import FilterBox from "./FilterBox.svelte";
 
   export let app;
   export let nodes = [];
-  let filterText = "";
   let nodesWithVisibility;
-  const filterTextChanged = () => {
+  const filterTextChanged = (filterText = "") => {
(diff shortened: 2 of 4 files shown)


src/pages/AppList.svelte
```diff
@@ -1,16 +1,32 @@
...
+  };
 </script>
 
 <h2>Applications</h2>
+<input
+  class="p-2 mb-4"
+  type="text"
+  bind:value={searchTerm}
+  on:input={searchTextChange}
+  placeholder="Search" />
```
wlach, line 26: The styling we have in the table view component looks a little nicer:

https://github.com/mozilla/glean-dictionary/blob/5e787003843d662079e8c047078c108de8b6d976/src/components/SchemaViewer.svelte#L50

The easiest way to make sure we get this (and that it stays consistent if we ever want to change it) is to refactor the filter box into a seperate component. I'll leave how to do so as an exercise for you to figure out based on the repository layout. 😄 If you need some help, the svelte tutorial is very good: https://svelte.dev/tutorial/basics

wlach (changes requested):
Hi @DawoodShahat! Great start, but let's try to refine the solution a bit. Some suggestions in line.

src/components/FilterBox.svelte
```diff
@@ -0,0 +1,20 @@
+<script>
+  import throttle from "just-throttle";
+
+  export let onChangeText = null;
```
wlach, line 4: Pretty sure this doesn't need to be set to null.

```suggestion
  export let onChangeText;
```

src/components/SchemaViewer.svelte
```diff
@@ -1,12 +1,11 @@
...
-  import throttle from "just-throttle";
   import SchemaNode from "./SchemaNode.svelte";
+  import FilterBox from "./FilterBox.svelte";
 
   export let app;
   export let nodes = [];
-  let filterText = "";
   let nodesWithVisibility;
-  const filterTextChanged = () => {
+  const filterTextChanged = (filterText = "") => {
```
wlach, line 8: I don't think this function needs a default value?

```suggestion
  const filterTextChanged = (filterText) => {
```

src/pages/AppList.svelte
```diff
@@ -1,16 +1,25 @@
 <script>
+  import FilterBox from "../components/FilterBox.svelte";
```
wlach, line 2: Let's actually call this a FilterInput, as FilterBox is a bit ambiguous (you'll also need to rename the component).

```suggestion
  import FilterInput from "../components/FilterInput.svelte";
```

wlach (changes requested):
This is looking great! Just a few more changes then I think we'll be good to merge.

src/components/SchemaViewer.svelte
DawoodShahat: This function is called inside SchemaViewer for the initial render. If I remove the default value it throws variable undefined error.

wlach:
Thank you @DawoodShahat! This looks and works great.

wlach (approved):

Issue #53, opened 2020-10-01
Hide applications that are deprecated by default


wlach:
*This issue is intended as an onboarding task for potential outreachy applicants. Please do not work on it unless you have completed the initial qualification task and it has been assigned to you.*

Currently we list all Glean applications in the UI by default on the home page (https://github.com/mozilla/glean-dictionary/blob/12ff6ac8f603f1245f6c6dcb9e4be9b85e28b135/src/pages/AppList.svelte), regardless of deprecation status. Instead we should have a checkbox that allows you to show/hide the applications that are deprecated. Something like this:

- [ ] Show deprecated applications

You can use the `deprecated` property in the apps json file to accomplish this task. Assuming you have the application running, have a look at this JSON payload:

http://localhost:5000/data/apps.json

wlach:
Fixed by #78

PR #78, opened 2020-10-09
Hid applications that are deprecated by default

mhmohona:
This PR is submitted for issue #53. 
Here I have added a checkbox and if the check box is unchecked, deprecated apps will be hidden. If it is checked, all applications will appear.
image
image

### DIFF
--- a/src/pages/AppList.svelte
+++ b/src/pages/AppList.svelte
@@ -1,7 +1,6 @@
 <script>
-	import { fade } from 'svelte/transition';
-	let visible = true;
-
+  import { fade } from "svelte/transition";
+  let showDeprecated = true;
   const URL = "data/apps.json";
   let apps;
   fetch(URL)
@@ -14,32 +13,26 @@
 <h2>Applications</h2>
 
 <label>
-	<input type="checkbox" bind:checked={visible}>
-	Show deprecated applications
+  <input type="checkbox" bind:checked={showDeprecated} />
+  Show deprecated applications
 </label>
 
-{#if visible}
-	<p transition:fade>
-		{#if apps}
-     {#each apps as app}
-       <p>
-       <a href="/apps/{app.name}">{app.name}</a>
-       {#if app.description}<i>{app.description}</i>{/if}
-       </p>
-      {/each}
-    {/if}   
-	</p>
-
-{:else}
+{#if showDeprecated}
   {#if apps}
-  {#each apps as app}    
-    {#if !app.deprecated}
-      <p> 
+    {#each apps as app}
+      <p>
         <a href="/apps/{app.name}">{app.name}</a>
         {#if app.description}<i>{app.description}</i>{/if}
-     </p>
+      </p>
+    {/each}
   {/if}
+{:else if apps}
+  {#each apps as app}
+    {#if !app.deprecated}
+      <p>
+        <a href="/apps/{app.name}">{app.name}</a>
+        {#if app.description}<i>{app.description}</i>{/if}
+      </p>
+    {/if}
   {/each}
-{/if}
-   
 {/if}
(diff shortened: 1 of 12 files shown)


mhmohona:
Hello @wlach! This PR is ready for review.

src/pages/AppList.svelte
```diff
@@ -1,4 +1,7 @@
 <script>
+	import { fade } from 'svelte/transition';
+	let visible = true;
```
wlach, line 3: We should use a more descriptive variable name here.

```suggestion
	let showDeprecated = true;
```

src/pages/AppList.svelte
```diff
@@ -1,4 +1,7 @@
 <script>
+	import { fade } from 'svelte/transition';
```
wlach, line 2: I appreciate the enthusiasm and desire to experiment, but let's leave the animated transitions to a seperate PR / issue. :) This helps describe why: https://homes.cs.washington.edu/~mernst/advice/version-control.html#logical-unit

For now I'd like to mostly concentrate on making sure that all the required UI elements are there, rather than on how they look.

wlach (changes requested):
The logic looks good! Aside from the review comments, it looks like there are some minor formatting issues (which you can fix by running `npm run format`) and there is a minor merge conflict with a PR that just landed. Can you take a look?

mhmohona:
This is the updated UI. 
image

src/pages/AppList.svelte
mhmohona: I thought it would make the boring UI looks nicer :sweat_smile: 
Will be careful about it in future.

wlach (changes requested):
Hi! Will take a closer look later, however in the mean time can you fix the merge conflict with upstream? You'll need to pull from the upstream repository and fix the conflicts (should not be too hard, let us know if you run into problems)

src/pages/AppList.svelte
```diff
@@ -14,15 +14,34 @@
...
+    {/each}
+  {/if}
+{:else if apps}
+  <FilterInput onChangeText={filterApps} />
   {#each filteredApps as app}
-    <p>
-      <a href="/apps/{app.name}">{app.name}</a>
-      {#if app.description}<i>{app.description}</i>{/if}
-    </p>
+    {#if !app.deprecated}
```
wlach, line 40: You can skip the if/else (and the duplicated logic for display which violates the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle) by doing the following on display:

```suggestion
    {#if showDeprecated || !app.deprecated}
```

Alternately, we could probably filter the applications in the filterApps function above. I'd accept either solution for now.

wlach (changes requested):
Getting close! Have some more feedback that I'd like to see addressed before this lands.

src/pages/AppList.svelte
mhmohona: Edited as per your suggestion @wlach . :smile:

wlach: Hi @mhmohona -- it looks like the logic is still duplicated unfortunately. To be clear, you should only *one* block that iterates through the app list and no higher level if/else (i.e. the logic should be substantially the same as before, just with that extra piece of logic that I showed above).

mhmohona: Sorry sorry @wlach, at that time I was confused. :sweat_smile: 
I hope now I could do what you asked.

I did not know about DRY principle before. I really appriciate the way you are helping to learn, to improve.

src/pages/AppList.svelte
```diff
@@ -14,15 +14,23 @@
   function filterApps(filterText) {
     filteredApps = apps.filter((appItem) => appItem.name.includes(filterText));
   }
+  let showDeprecated = true;
```
wlach, line 17:
```suggestion
  let showDeprecated = false;
```

wlach (changes requested):
Code looks good minus one suggestion. It looks like there's also a minor conflict still -- can you resolve it? You can probably do this entirely from the GitHub UI.

wlach (approved):
Thank you!

PR #82, opened 2020-10-09
Improve ping details (fixes #77)

### LINKED ISSUE #77: Ping page has almost no info

We show only a tiny subset of the data related to the ping on the ping page right now. You can see an example of markdown documentation which more fully represents the metadata we have here:

https://github.com/mozilla/mozregression/blob/master/docs/glean/metrics.md#pings

The code that needs to be modified is here:

https://github.com/mozilla/glean-dictionary/blob/main/src/pages/PingDetail.svelte

All the information we want to display *should* already be extracted. If you have the server running locally, go to e.g.:

http://localhost:5000/data/mozregression/pings/usage.json

This is the dataset corresponding to:

http://localhost:5000/#!/apps/mozregression/pings/usage

Don't worry about styling the component too much-- can just use a table for now (like we do in the table view already: http://localhost:5000/#!/apps/mach/tables/usage)

DawoodShahat:
Here is an improvement to PingDetails.svelte page

ping_page

### DIFF
--- a/src/pages/PingDetail.svelte
+++ b/src/pages/PingDetail.svelte
@@ -3,14 +3,6 @@
 
   export let params;
   const pingDataPromise = getPingData(params.app, params.ping);
-  const URL = `data/${params.app}/index.json`;
-  let app;
-
-  fetch(URL)
-    .then((r) => r.json())
-    .then((ret) => {
-      app = ret;
-    });
 </script>
 
 <style>
@@ -60,16 +52,14 @@
       </td>
     </tr>
   </table>
-{/await}
 
-{#if app}
   <h2>Metrics</h2>
   <ul>
-    {#each app.metrics as metric}
+    {#each ping.metrics as metric}
       <li>
         <a href={`/apps/${params.app}/metrics/${metric.name}`}>{metric.name}</a>
         <i>{metric.description}</i>
       </li>
     {/each}
   </ul>
-{/if}
+{/await}
--- a/src/pages/PingDetail.svelte
+++ b/src/pages/PingDetail.svelte
@@ -3,6 +3,14 @@
 
   export let params;
   const pingDataPromise = getPingData(params.app, params.ping);
+  const URL = `data/${params.app}/index.json`;
+  let app;
+
+  fetch(URL)
+    .then((r) => r.json())
+    .then((ret) => {
+      app = ret;
+    });
 </script>
 
 <style>
@@ -53,3 +61,15 @@
     </tr>
   </table>
 {/await}
+
+{#if app}
+  <h2>Metrics</h2>
+  <ul>
+    {#each app.metrics as metric}
+      <li>
+        <a href={`/apps/${params.app}/metrics/${metric.name}`}>{metric.name}</a>
+        <i>{metric.description}</i>
+      </li>
+    {/each}
+  </ul>
+{/if}
--- a/src/pages/PingDetail.svelte
+++ b/src/pages/PingDetail.svelte
@@ -18,8 +18,9 @@
 </style>
 
 {#await pingDataPromise then ping}
+  <p><a href={`/apps/${params.app}/tables/${ping.name}`}>BigQuery table</a></p>
   <h1>{ping.name}</h1>
-  <table class="bg-gray-100 table-header">
+  <table class="table-header">
     <tr>
       <td>Description</td>
       <td>{ping.description}</td>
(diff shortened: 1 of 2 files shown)


DawoodShahat:
@wlach Hey, Please take a look into this when possible.

src/pages/PingDetail.svelte
```diff
@@ -5,8 +5,52 @@
...
+    <tr>
+      <td>Reviews Data</td>
+      <td>
+        {#each ping.data_reviews as review}
+          <a class="mr-2" href={review}>{review}</a>
+        {/each}
+      </td>
+    </tr>
+    <tr>
+      <td>Client Id</td>
```
wlach, line 42:
```suggestion
      <td>Includes Client Identifier</td>
```

src/pages/PingDetail.svelte
```diff
@@ -5,8 +5,52 @@
...
+  .table-header td {
+    @apply border;
+    @apply p-2;
+  }
+</style>
+
 {#await pingDataPromise then ping}
   <h1>{ping.name}</h1>
-  <p>{ping.description}</p>
+  <table class="bg-gray-100 table-header">
```
wlach, line 22: Let's skip the styling for now.

```suggestion
  <table class="table-header">
```

src/pages/PingDetail.svelte
```diff
@@ -5,8 +5,52 @@
...
+        {/each}
+      </td>
+    </tr>
+    <tr>
+      <td>Client Id</td>
+      <td>{ping.client_id ? 'Yes' : 'No'}</td>
+    </tr>
+    <tr>
+      <td>
+        notification_email{ping.notification_emails.length > 1 ? 's' : ''}
```
wlach, line 47:
```suggestion
        Notification Email{ping.notification_emails.length > 1 ? 's' : ''}
```

src/pages/PingDetail.svelte
```diff
@@ -5,8 +5,52 @@
...
+      </td>
+      <td>
+        {#each ping.notification_emails as email}<span>{email}</span>{/each}
+      </td>
+    </tr>
+  </table>
+
   <p><a href={`/apps/${params.app}/tables/${ping.name}`}>BigQuery table</a></p>
```
wlach, line 55: Let's make this part of the table above.

wlach (changes requested):
Great start! I have some initial feedback inline. The main thing missing from this is the list of metrics (along with links to their metrics page) that are sent part of this ping. Basically I would add a "Metrics" section just below the table that you added.

Refer back to the documentation linked in #77 for more detail on this (and some example documentation).

src/pages/PingDetail.svelte
```diff
@@ -3,10 +3,73 @@
...
   export let params;
   const pingDataPromise = getPingData(params.app, params.ping);
+  const URL = `data/${params.app}/index.json`;
+  let app;
+
+  fetch(URL)
+    .then((r) => r.json())
+    .then((ret) => {
+      app = ret;
+    });
```
wlach, line 13: This should not be necessary-- a short metric summary should also be sent as part of the ping.

I just realized that this is not currently the case though, which is a good opportunity to get some experience with the python parts of the codebase.

wlach (changes requested):
Thanks for this @DawoodShahat! The Svelte code looks good, but I realized (too late) that we're actually going to need to modify our ETL code (in Python) a bit to handle this well. Redownloading the application metadata is not the right solution.

This is actually good news, as I was going to ask you to look into something in this area next. 😄 

If you look at our script right now, there's some code to write out the ping json file that you're using in this view:

https://github.com/mozilla/glean-dictionary/blob/5499eeb17deaef22a3620c101f53c3792dcebbf4/scripts/build-glean-metadata.py#L54

Basically what we need to change is gather the metrics data *first*, figure out which pieces of metric metadata apply to which pings via the `send_in_pings` field, then add a short summary of the metrics to each pings blob. This is analagous to what we already do for applications here: https://github.com/mozilla/glean-dictionary/blob/5499eeb17deaef22a3620c101f53c3792dcebbf4/scripts/build-glean-metadata.py#L50

Let me know if you have any questions!

src/pages/PingDetail.svelte
DawoodShahat: Thanks for the review, As far as I understand I have to modify build-glean-metadata.py so that it writes pings data in a way that each ping should include 

- A list of metrics associated with the ping.
- A short description of each metric

This will help us avoid re-downloading the whole app data.

DawoodShahat: Should update the build-glean-metadata.py in a separate PR?

wlach: go ahead and do it in this one!

DawoodShahat:
Here is what it looks like after the suggestions.

Screenshot 2020-10-13 at 8 28 25 PM

wlach (changes requested):
Looks good, one last requested change and then we can land.

wlach (approved):

Issue #80, opened 2020-10-09
Table search only searches in last part of the component


wlach:
*This issue is intended as an onboarding task for potential outreachy applicants. Please do not work on it unless you have completed the initial qualification task and it has been assigned to you.*

Currently the filter search in the table view (e.g. http://localhost:5000/#!/apps/mach/tables/usage) only filters for content in the last "node" in the structure. e.g. if you have

```
client_info.app_version
client.info.app_display_version
client_info.client_id
...
```

and you search for "client", it will filter out all of the above except for `client_info.client_id`. Ideally the search would include all entries whose parent elements have the term in them (in the example above, this would be everything that includes `client`).

To fix this, you'll need to edit the schemaviewer component: https://github.com/mozilla/glean-dictionary/blob/0a54d6fee06c9031852fb96b19657c598ddbd1e9/src/components/SchemaViewer.svelte#L8

wlach:
Going to assign this to @hngerebara

UdokaVrede:
Hi wlach how can I locate the initial qualification task

mhmohona:
Hello @UdokaVrede! I am another outreachy applicant like you but as @wlach is currently on vacation and will be available from Monday, I am replying you.
 
In project Outreachy [website](https://www.outreachy.org/outreachy-december-2020-internship-round/communities/mozilla/#extend-glean-dictionary) you will find details about initial qualification task. 


These are steps mentioned there in website - 

The first step of the application process is to:    
 
* If you haven't used Svelte before (which is ok!), run through at least some of their interactive tutorial: https://svelte.dev/tutorial/basics
* Checkout the glean dictionary project (https://github.com/mozilla/glean-dictionary)
* Following the instructions in the README, extract and process the dataset information and get a local copy of the site running
* Find the "usage ping" BigQuery table for the "mach" application and find the client_info.app_channel column. Using your text editor and the locally checked out source, make it so that the description of the column is shown in RED rather than the current colour of GREY
* Take a screenshot and create a diff of the changes you made to make this change, and DM these to `wlach` on chat.mozilla.org (see the project communications channels) along with your GitHub user id

UdokaVrede:
Thanks

PR #84, opened 2020-10-10
Fix table search

hngerebara:
This PR fixes the issue stated in this [PR](https://github.com/mozilla/glean-dictionary/issues/80)

It now displays all search results that match the search term

Screenshot 2020-10-10 at 2 41 19 PM


DawoodShahat:
@hngerebara Hey, make sure to run ```npm run format``` before pushing to avoid build fails based on formatting.

mhmohona:
@DawoodShahat is right! Also [this](https://github.com/mozilla/glean-dictionary/blob/main/docs/development.md#code-quality) might be helpful.

hngerebara:
Hi @wlach, this is ready for review. Please review it. Thanks

meghajain-1711:
One other thing that helped me was **npm test** after npm format and build, before committing. It runs the same circleCI and gives the error before hand if any :)

acmiyaguchi (changes requested):
Thanks for taking time to work through this. I've left some comments inline. There are cases where it would be useful to include results that match the search key that are more than a parent away. One particular case would be a key-value pair in the schema, like `ping_info.experiments`, where we'd want to see more than just `ping_info.experiments.key` and `ping_info.experiments.value`.

Here's a patch file with the storybook test case that you can use to explore your patch with more control over the variables. 

https://gist.githubusercontent.com/acmiyaguchi/084f52020fe557697ca6d1f78f6f028c/raw/fdd0ee57703a9abb8f64659f034387f3ceef6b8d/0001-Add-storybook-test-for-nested-search-with-css-fix-fr.patch

To apply this, download it to the project root and run `git apply 0001-Add-storybook-test-for-nested-search-with-css-fix-fr.patch`. Start storybook with `npm run storybook` and navigate to the `SchemaViewer > Nested` page.  Feel free to use/modify as you see fit. If you do check it in, omit `story.css` which was take from #74.

src/components/SchemaViewer.svelte
```diff
@@ -14,13 +14,24 @@
...
     const addVisibility = (node) => {
       let modifiedNode = node;
+      let parentName = "";
 
       if (modifiedNode.fields) {
-        modifiedNode.fields.forEach(addVisibility);
+        parentName = modifiedNode.name;
+        modifiedNode.fields.forEach((field) => {
+          let modifiedNodeField = field;
+          modifiedNodeField.parentName = parentName;
```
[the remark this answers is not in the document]
hngerebara: Okay

hngerebara:
@acmiyaguchi Thanks for reviewing this. I didn't know there were more nodes with such depth. I have fixed it to cater for such situations. You can review again @acmiyaguchi @wlach

src/components/SchemaViewer.svelte
```diff
@@ -11,15 +11,32 @@
...
+          let modifiedNodeField = field;
+          modifiedNodeField.parentNames = parentNames;
+          return addVisibility(modifiedNodeField, parentNames);
+        });
       }
       modifiedNode.visible =
         filterTerms.length === 0 ||
-        filterTerms.every((term) => modifiedNode.name.includes(term));
+        filterTerms.every((term) =>
+          modifiedNode.parentNames && modifiedNode.parentNames.length > 0
```
acmiyaguchi, line 35: This could use the `parentNames` variable instead.

src/components/SchemaViewer.svelte
```diff
@@ -11,15 +11,32 @@
...
-        modifiedNode.fields.forEach(addVisibility);
+        if (parentNames.length > 0) {
+          parentNames.push(`${parentNames.slice(-1)}.${modifiedNode.name}`);
+        } else {
+          parentNames.push(`${modifiedNode.name}`);
+        }
+
+        modifiedNode.fields.forEach((field) => {
+          let modifiedNodeField = field;
+          modifiedNodeField.parentNames = parentNames;
```
acmiyaguchi, line 28: We can have same results without resorting to side effects on the node.

src/components/SchemaViewer.svelte
```diff
@@ -11,15 +11,32 @@
...
+        });
       }
       modifiedNode.visible =
         filterTerms.length === 0 ||
-        filterTerms.every((term) => modifiedNode.name.includes(term));
+        filterTerms.every((term) =>
+          modifiedNode.parentNames && modifiedNode.parentNames.length > 0
+            ? modifiedNode.parentNames.some((val) => val.includes(term)) ||
+              modifiedNode.name.includes(term)
+            : modifiedNode.name.includes(term)
```
acmiyaguchi, line 38: If I'm reading this correctly, `parentNames` will always be greater than 0 and include the name of the current node, so this logic can be simplified.

src/components/SchemaViewer.svelte
```diff
@@ -11,15 +11,32 @@
       .split(" ")
       .filter((t) => t.length > 0);
 
-    const addVisibility = (node) => {
+    const addVisibility = (node, parentNodeNames) => {
```
acmiyaguchi, line 14: If you wanted to reduce the amount of conditional branches in this function, I think you could set a default value to `parentNodeNames=["root"]` and remove a few of the `parentName.length > 0` checks.

acmiyaguchi (changes requested):
Great, searching for `experiments` returns useful results. Inline are some ways to simplify your patch.

src/components/SchemaViewer.svelte
hngerebara: Okay

src/components/SchemaViewer.svelte
hngerebara: @acmiyaguchi Please can you elaborate on this? not sure I understand

acmiyaguchi: So it's not actually a [side-effect](https://en.wikipedia.org/wiki/Side_effect_(computer_science)) since the modified node is never being returned, but the idea is that the variable should be used directly instead of being set in the object.

src/components/SchemaViewer.svelte
```diff
@@ -11,24 +11,38 @@
       .split(" ")
       .filter((t) => t.length > 0);
 
-    const addVisibility = (node) => {
+    const addVisibility = (node, parentNodeNames = [`${node.name}`]) => {
```
acmiyaguchi, line 14:
```suggestion
    const addVisibility = (node, parentNodeNames = ["__root__"]) => {
```

Starting with the name of the current node means that there are a lot of changes from the last commit. My last suggestion was to start with a literal root value (e.g. `__root__` to make it obvious that it's not an actual column name). This way, there wouldn't be any reason to check for whether `parentNodeNames` was empty, like in this block:

https://github.com/mozilla/glean-dictionary/pull/84/commits/22c7a2659d4ffe75c56eb74b222af4d244d20bb1#diff-813bcbaca6a1e7fb8e6fc6ab82618d42ae769b8beb12d76f4582c08ecfe8f1d1R22-R26

acmiyaguchi (commented):
I thought the flow and logic of the last commit was good. I see how you incorporated my feedback, but I think it might have made this more complicated. I'm sorry if any of the comments were not clear or misleading.

src/components/SchemaViewer.svelte
acmiyaguchi: And just to be clear, my suggestion is to have removed this line.

src/components/SchemaViewer.svelte
```diff
@@ -11,24 +11,38 @@
...
+            : name.includes(term)
+        );
+      modifiedNode.childrenVisible = fields
+        ? fields.some((child) => child.visible || child.childrenVisible)
         : undefined;
       return modifiedNode;
     };
 
-    nodesWithVisibility = nodes.map(addVisibility, filterTerms);
+    nodesWithVisibility = nodes.map((node) => addVisibility(node), filterTerms);
```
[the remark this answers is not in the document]
hngerebara: I did this because I get this error `number 0 is not iterable` .When addVisibility is first called, the index of the node is passed into the `parentNodeNames` which cannot be iterated

src/components/SchemaViewer.svelte
hngerebara: Removing this line flaws the entire logic as parentNames is needed

src/components/SchemaViewer.svelte
```diff
@@ -11,15 +11,29 @@
...
-    const addVisibility = (node) => {
+    const addVisibility = (node, parentNodeNames = ["__root__"]) => {
       let modifiedNode = node;
+      let parentNames = [...parentNodeNames];
 
       if (modifiedNode.fields) {
-        modifiedNode.fields.forEach(addVisibility);
+        parentNames = modifiedNode.parentNames
+          ? `${modifiedNode.parentNames}.${modifiedNode.name}`
+          : modifiedNode.name;
```
acmiyaguchi, line 21: This can be simplified to something like the following. This was part of one of your earlier revision of the PR.

```js
        parentNames = `${modifiedNode.parentNames.slice(-1)}.${modifiedNode.name}`
```

src/components/SchemaViewer.svelte
```diff
@@ -11,15 +11,29 @@
...
+
+        modifiedNode.fields.forEach((field) => {
+          let modifiedNodeField = field;
+          modifiedNodeField.parentNames = parentNames;
```
acmiyaguchi, line 25: Remove this line and adjust accordingly. Treat the node as if it cannot have fields added to it on the fly.

```suggestion

```

src/components/SchemaViewer.svelte
acmiyaguchi: My previous comment still applies.

```js
           filterTerms.every((term) => parentNames.some(...) || modifiedNode.name.includes(term));
```

acmiyaguchi (changes requested):

acmiyaguchi (approved):
Looks good to me, thanks for working through all of the request changes.

PR #85, opened 2020-10-10
ignore venv

DawoodShahat:
This will help us avoid pushing python installed dependencies to glean-dictionary repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /node_modules/
 /public/build/
+/venv
 .probe_cache
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 /node_modules/
 /public/build/
-/venv
+venv
+*.pyc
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 /node_modules/
 /public/build/
+/venv
```
acmiyaguchi, line 3: Note that putting the forward slash before the folder makes this relative to the root `venv` subdirectories won't be caught.

acmiyaguchi (commented):
Could you also add ignores for compile python bytecode (`*.pyc`) and cached files (`__pycache__/`), which will eventually crop up.

.gitignore
DawoodShahat: Thanks! I didn't take these into consideration. Creating a commit with suggested changes.

acmiyaguchi (approved):
Thanks!

PR #90, opened 2020-10-11
Refactoring some code to follow DRY principle.

DawoodShahat:


DawoodShahat:
@acmiyaguchi Hey, Take a look into this when possible.

src/pages/AppDetail.svelte
```diff
@@ -1,17 +1,14 @@
...
+
   export let params;
   const URL = `data/${params.app}/index.json`;
-  let app;
-  fetch(URL)
-    .then((r) => r.json())
-    .then((ret) => {
-      app = ret;
-    });
+  let app = fetchJSON(URL);
```
acmiyaguchi, line 6: I've dealt with async sections like this using the `onMount` lifecycle function, but the `await` keyword in the document works too. Just for reference, it would look something like this:

```javascript
import { onMount } from "svelte";

...

onMount(async () => {
  let app = await fetchJSON(URL);
});
```

src/pages/AppList.svelte
```diff
@@ -1,28 +1,31 @@
...
   function filterApps(filterText) {
     filteredApps = apps.filter((appItem) => appItem.name.includes(filterText));
   }
 </script>
 
 <h2>Applications</h2>
-<FilterInput onChangeText={filterApps} />
-{#if apps}
-  {#each filteredApps as app}
+{#await filteredApps then apps}
```
acmiyaguchi, line 23: The variable assignment flow here is confusing, especially since `filteredApps` can be both an async function (its initial assignment) and an array (assignment from `filterApps`). `filteredApps` also assigns at the same time as it returns a variable.

I'd actually recommend using the `onMount` function here, which should reduce the size of the diff.

acmiyaguchi (changes requested):

src/pages/AppList.svelte
DawoodShahat: Yes, the variable flow is confusing, and naming things is very difficult. I will look into onMount and add the suggested changes. Thanks.

src/pages/AppDetail.svelte
```diff
@@ -1,17 +1,18 @@
...
-</script>
 
-<h1>{params.app}</h1>
+  onMount(async () => {
+    app = await fetchJSON(URL);
+  });
+</script>
 
 {#if app}
+  <h1>{params.app}</h1>
```
acmiyaguchi, line 15: If the app failed to load, the page would be blank. There's no real reason to put this inside of the conditional, the `params` object doesn't depend on the state of `fetchJSON`.

src/pages/AppList.svelte
```diff
@@ -1,24 +1,26 @@
...
-  fetch(URL)
-    .then((r) => r.json())
-    .then((ret) => {
-      apps = ret.sort((a, b) => a.app_id > b.app_id);
-      filteredApps = apps;
-    });
+
+  onMount(async () => {
+    apps = await fetchJSON(URL);
+    apps = apps.sort((a, b) => a.app_id > b.app_id);
```
acmiyaguchi, line 12: Just a nit, but it's generally nicer to not have to reassign to the same variable. Maybe something along the lines of `let data = ...; apps = data.sort(...)`.

acmiyaguchi (commented):
This looks good to me, with a couple of small things in the comments.

src/pages/AppList.svelte
DawoodShahat: Thanks, It doesn't need to be re-assigned as the sort() function is in-place algorithm.

acmiyaguchi: I see, I was thinking `sorted` like python. This would be error, then?

DawoodShahat: Yes, I added it on a new line as ```apps.sort(...)``` which works fine.

acmiyaguchi (approved):
Thanks for refactoring these bits.

DawoodShahat:
I'm happy to help. I learned so much working on this project. I'm also getting more familiar with git/github workflow. Thanks for reviewing my work.

Issue #100, opened 2020-10-12
Add info about the type of metric and expiry date on Metric Detail Page


meghajain-1711:
Currently the enhancement on Metric detail page so far shows only Json data. 

We should add the information about : 
1) Type of metric 
2) Expiry date of the metric 
3) Starting point(app) of the metric [ fenix/mach/...]

Where : 
Below description of the Metric

More to do : 
1) We can also add more_info links as hyperlink to the type of metric 
2) Hyperlink to the starting point to take it back to the App page eg "mach/fenix" page in this case.

PR #79, opened 2020-10-09
Show info for Metric Details Page

meghajain-1711:
Added Metric Info to the Metric page.

The keys currently are the same as media_state_play page ( bugs, data_reviews, lifetime, disabled, .. ) 

Formatting/styling has not yet been added as mentioned in the Issue. But the code has been written so that it can be formatted similar to the above mentioned page ( Check the image in Issue) 

More info (help) has been linked for lifetime key. It can be further added for others in the similar manner.

This is for Issue : Metric pages show almost no info #66

PR #103, opened 2020-10-12
Add info about the type of metric and expiry date on Metric Detail Page

meghajain-1711:
Fixes issue #100

Added type of Metric and Expiry Information below the Metric name and description.
The Metric type links to its documentation as well.

Example of ; never expires
image

Example of : expiry date to come
image

Example of : expiry date gone past
image

@wlach your comments are welcome :)

### DIFF
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -7,24 +7,24 @@
   function getMetricDocumentationURI(type) {
     const sourceDocs = "https://mozilla.github.io/glean/book/user/metrics/";
     const links = {
-      memory_distribution: `${sourceDocs}memory_distribution.html`,
-      quantity: `${sourceDocs}quantity.html`,
-      custom_distribution: `${sourceDocs}custom_distribution.html`,
-      string_list: `${sourceDocs}string_list.html`,
-      labeled_string: `${sourceDocs}labeled_strings.html`,
-      timespan: `${sourceDocs}timespan.html`,
-      datetime: `${sourceDocs}datetime.html`,
-      string: `${sourceDocs}string.html`,
-      timing_distribution: `${sourceDocs}timing_distribution.html`,
-      boolean: `${sourceDocs}boolean.html`,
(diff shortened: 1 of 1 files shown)


DawoodShahat:
Hey @meghajain-1711 , I think you can add these changes to your previous PR [https://github.com/mozilla/glean-dictionary/pull/79](https://github.com/mozilla/glean-dictionary/pull/79) 

Because they are the same pages. 

You would want something like this that has the information like metric_type and metric_expiration date
Screenshot 2020-10-13 at 2 04 02 AM

You can put all the data in one place for a specific metric.

src/pages/MetricDetail.svelte
```diff
@@ -3,9 +3,54 @@
...
+      timing_distribution: `${sourceDocs}timing_distribution.html`,
+      boolean: `${sourceDocs}boolean.html`,
+      labeled_counter: `${sourceDocs}labeled_counters.html`,
+      uuid: `${sourceDocs}uuid.html`,
+      counter: `${sourceDocs}counter.html`,
+      event: `${sourceDocs}event.html`,
+      jwe: `${sourceDocs}jwe.html`,
+    };
+
+    return links[type] || sourceDocs;
```
wlach, line 27: This seems like a good approach to me, but there's no need to define the sourceDocs variable each time-- instead, try something like this:

```suggestion
    return `${sourceDocs}${links[type]}` || sourceDocs;
```

wlach (changes requested):
> Hey @meghajain-1711 , I think you can add these changes to your previous PR #79
> ...

Yeah, strictly I would agree -- this would be in scope for #79. However, I'm also ok with landing this separately. There will definitely be many followups to all the pages at this early stage, as we figure out what to display and how to display it.

I'm ok to land this once the one issue I mention below is addressed.

meghajain-1711:
Hi @wlach Updated as per the request. It definitely made the codebase simpler :) 
Also , I have added a link to the params.app ie Application from which the metric was found. 

Let me know if it looks good now. 
image

wlach (approved):

PR #104, opened 2020-10-12
Sort function has no effect on the list of apps

DawoodShahat:
The callback to the sort function returns a boolean. 

```
 apps = ret.sort((a, b) => a.app_id > b.app_id);
```
It has no effect on the resultant array as shown in this example

```
let arr = [9, 3, 2, 0, 3, 1000, 8, 7, 6, 5, 4, 3];

arr.sort((a, b) => a > b);
// [9, 3, 2, 0, 3, 1000, 8, 7, 6, 5, 4, 3]

```

after updating the callback
```
arr.sort((a, b) => a > b ? 1 : -1)

// [ 0, 2, 3, 3, 3, 4, 5, 6, 7, 8, 9, 1000]
```

It sorts the list of apps based on the app_id in an alphabetical order
As shown here
Screenshot 2020-10-12 at 10 56 58 PM

**before updating the callback, The list was rendered different on both Chrome and Firefox browsers**

- Chrome Browser
chrome_result

- Firefox Browser
firefox_result

**Now It will be rendered same across both browsers now**

### DIFF
--- a/src/pages/AppList.svelte
+++ b/src/pages/AppList.svelte
@@ -9,7 +9,7 @@
 
   onMount(async () => {
     apps = await fetchJSON(URL);
-    apps.sort((a, b) => (a.app_id > b.app_id ? 1 : -1));
+    apps.sort((a, b) => (a.name > b.name ? 1 : -1));
     filteredApps = apps;
   });
 
--- a/src/pages/AppList.svelte
+++ b/src/pages/AppList.svelte
@@ -9,7 +9,7 @@
 
   onMount(async () => {
     apps = await fetchJSON(URL);
-    apps.sort((a, b) => a.app_id > b.app_id);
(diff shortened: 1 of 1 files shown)


DawoodShahat:
@wlach Take a look into this when possible. Thanks.

src/pages/AppList.svelte
```diff
@@ -9,7 +9,7 @@
 
   onMount(async () => {
     apps = await fetchJSON(URL);
-    apps.sort((a, b) => a.app_id > b.app_id);
+    apps.sort((a, b) => (a.app_id > b.app_id ? 1 : -1));
```
wlach, line 12:
```suggestion
    apps.sort((a, b) => (a.name > b.name ? 1 : -1));
```

wlach (approved):
We may as well sort the apps alphabetically while we're here (by sorting on name instead of id). Other than that, looks good! I'll merge

Issue #105, opened 2020-10-13
Include more application metadata


wlach:
We don't have a ton of Glean application metadata currently, but we have more than we're currently displaying (name + description).

In particular, we should include the following:
* source code URL -- where to look up the source code
* application id -- the id of the application, corresponding to what we use on the play store with android applications
* deprecated (if true) -- whether the application is deprecated (no longer under active development)

application id and source code URL should only be displayed in the application detail screen. For now, just display them as a table, as we do for the BigQuery table view e.g. http://localhost:5000/#!/apps/fenix-nightly/tables/activation

"deprecated" should be displayed as a pill (in both the application list and application detail screen). Create a new svelte component using tailwind with a pleasing style. You can see some documentation on how to create a pill here:

https://tailwindcss.com/docs/border-radius#pills-and-circles

wlach:
@JoyLubega want to give this a try?

wlach:
Fixed by #112

PR #112, opened 2020-10-13
Add more metadata for applications

JoyLubega:
Resolve: Issue [105](https://github.com/mozilla/glean-dictionary/issues/105#event-3871348947)
### App list
Screenshot 2020-10-14 at 21 58 52

### App Details
Screenshot 2020-10-14 at 21 58 59

### DIFF
--- a/src/components/Pill.svelte
+++ b/src/components/Pill.svelte
@@ -5,6 +5,6 @@
 
 <div
   style="background-color: {bgColor}"
-  class="inline-block border  rounded-full text-xs py-1 px-3 w-15  text-white">
+  class="inline-block border rounded-full text-xs py-1 px-3 w-15 text-white">
   {message}
 </div>
--- a/src/components/Pill.svelte
+++ b/src/components/Pill.svelte
@@ -1,0 +1,10 @@
+<script>
+  export let message = "";
+  export let bgColor;
+</script>
+
+<div
+  style="background-color: {bgColor}"
+  class="inline-block border  rounded-full text-xs py-1 px-3 w-15  text-white">
+  {message}
+</div>
--- a/src/pages/AppDetail.svelte
+++ b/src/pages/AppDetail.svelte
@@ -1,5 +1,6 @@
 <script>
   import { onMount } from "svelte";
+  import Pill from "../components/Pill.svelte";
   import { fetchJSON } from "../state/api";
   import FilterInput from "../components/FilterInput.svelte";
 
@@ -20,9 +21,35 @@
   }
 </script>
 
+<style>
+  .table-header {
+    @apply table-auto;
+    @apply my-4;
+  }
+
+  .table-header td {
+    @apply border;
+    @apply p-2;
+  }
+</style>
+
 <h1>{params.app}</h1>
 {#if app}
-  <p>{app.description}</p>
+  {#if app.deprecated}
+    <Pill message="Deprecated" bgColor="#4a5568" />
+  {/if}
+  <p class="mt-2">{app.description}</p>
+  <table class="table-header">
+    <tr>
+      <td>Source code Url</td>
+      <td><a href={app.url}>{app.url}</a></td>
+    </tr>
+    <tr>
+      <td>Application id</td>
+      <td><code>{app.app_id}</code></td>
+    </tr>
+  </table>
+
   <h2>Pings</h2>
   <ul>
     {#each app.pings as ping}
--- a/src/pages/AppList.svelte
+++ b/src/pages/AppList.svelte
@@ -1,6 +1,7 @@
 <script>
   import { onMount } from "svelte";
   import FilterInput from "../components/FilterInput.svelte";
+  import Pill from "../components/Pill.svelte";
   import { fetchJSON } from "../state/api";
 
   const URL = "data/apps.json";
@@ -25,14 +26,16 @@
   <input type="checkbox" bind:checked={showDeprecated} />
   Show deprecated applications
 </label>
-
 {#if apps}
   <FilterInput onChangeText={filterApps} />
   {#each filteredApps as app}
     {#if showDeprecated || !app.deprecated}
-      <p>
+      <p class="mb-2">
         <a href="/apps/{app.name}">{app.name}</a>
         {#if app.description}<i>{app.description}</i>{/if}
+        {#if app.deprecated}
+          <Pill message="Deprecated" bgColor="#4a5568" />
+        {/if}
       </p>
     {/if}
   {/each}


JoyLubega:
hey @wlach I am not able to add you as a reviewer

src/components/MetaData.svelte
```diff
@@ -0,0 +1,11 @@
+<script>
+  export let data;
+</script>
+
+{#if data}
+  <div
+    class="inline-block border border-blue-500 rounded-full text-xs py-1 px-3 w-15 bg-blue-800 text-white">
+    Deprecated:
+    {data.deprecated}
```
wlach, line 9: This content should be set by the parent component. See the svelte documentation if you're not sure how to do this.

src/components/MetaData.svelte
```diff
@@ -0,0 +1,11 @@
+<script>
```
wlach, line 1: We should give this component a more general name, like `Pill.svelte`. The idea is that we could use it in other places as well.

src/pages/AppList.svelte
```diff
@@ -30,9 +31,10 @@
   <FilterInput onChangeText={filterApps} />
   {#each filteredApps as app}
     {#if showDeprecated || !app.deprecated}
-      <p>
+      <p class="mb-2">
         <a href="/apps/{app.name}">{app.name}</a>
         {#if app.description}<i>{app.description}</i>{/if}
+        <MetaData data={app} />
```
wlach, line 37: As above, this should only display if the deprecated property is true. Also we should just show "Deprecated", not "Deprecated: true".

src/components/MetaData.svelte
```diff
@@ -0,0 +1,11 @@
...
+  export let data;
+</script>
+
+{#if data}
+  <div
+    class="inline-block border border-blue-500 rounded-full text-xs py-1 px-3 w-15 bg-blue-800 text-white">
```
wlach, line 7: We should make the background/border colours a configurable property of the component, rather than hardcoding it here.

src/pages/AppDetail.svelte
```diff
@@ -11,9 +12,33 @@
...
+
 <h1>{params.app}</h1>
 {#if app}
-  <p>{app.description}</p>
+  <p class="mb-2">{app.description}</p>
+  <MetaData data={app} />
+  <table class="table-header">
+    <tr>
+      <td>Source code Url</td>
+      <td><code>{app.url}</code></td>
```
wlach, line 34: This should be an actual link, not just a code block.

src/pages/AppDetail.svelte
```diff
@@ -11,9 +12,33 @@
...
+    @apply border;
+    @apply p-2;
+  }
+</style>
+
 <h1>{params.app}</h1>
 {#if app}
-  <p>{app.description}</p>
+  <p class="mb-2">{app.description}</p>
+  <MetaData data={app} />
```
wlach, line 30: Two things: 

1. This should only display if the deprecated property is true. Also we should just show "Deprecated".
2. It should be above the description

wlach (changes requested):
Hi, this is a good start but it needs some work. You might find working through the [svelte tutorial](https://svelte.dev/tutorial/basics) helpful, if you haven't seen it before.

JoyLubega:
@wlach I have made the changes

src/components/Pill.svelte
```diff
@@ -0,0 +1,12 @@
+<script>
+  export let data;
```
wlach, line 2: "message" is more concrete

```suggestion
  export let message = "";
```

src/components/Pill.svelte
```diff
@@ -0,0 +1,12 @@
...
+  export let pillColors;
+</script>
+
+{#if data}
+  <div
+    style="background-color: {pillColors.bgColor}; border-color: {pillColors.borderColor}"
+    class="inline-block border  rounded-full text-xs py-1 px-3 w-15  text-white">
+    Deprecated
+  </div>
+{/if}
```
wlach, line 12:
```suggestion
<div
  style="background-color: {pillColors.bgColor}; border-color: {pillColors.borderColor}"
  class="inline-block border  rounded-full text-xs py-1 px-3 w-15  text-white">
  {message}
</div>
```

src/components/Pill.svelte
```diff
@@ -0,0 +1,12 @@
+<script>
+  export let data;
+  export let pillColors;
```
wlach, line 3: It's clearer to have each parameter (bgColor and borderColor) passed in individually, rather than as an object.

Also, they should have reasonable default values (e.g. a neutral grey color)

src/pages/AppDetail.svelte
```diff
@@ -1,19 +1,47 @@
 <script>
   import { onMount } from "svelte";
+  import Pill from "../components/Pill.svelte";
   import { fetchJSON } from "../state/api";
 
   export let params;
   const URL = `data/${params.app}/index.json`;
   let app;
+  let pillColors = { bgColor: "#2c5282", borderColor: "#ffffff" };
```
wlach, line 9: There is no need to make this a variable (since it's always the same), you can just hardcode the values when creating the component.

wlach (changes requested):
Still a few issues here, thanks for staying with it!

src/pages/AppDetail.svelte
JoyLubega: hey @wlach I need clarity on this one. Do you mean i should  set the color as I am calling the component like this
`<Pill data={app} pillBgColor ="#2c5282" />`

wlach: Yup that's correct. No need to name it `pillBgColor` though, just `bgColor` should be sufficient.

JoyLubega: Done

src/components/Pill.svelte
```diff
@@ -0,0 +1,10 @@
+<script>
+  export let message = "";
+  export let bgColor;
+</script>
+
+<div
+  style="background-color: {bgColor}"
+  class="inline-block border  rounded-full text-xs py-1 px-3 w-15  text-white">
```
wlach, line 8:
```suggestion
  class="inline-block border rounded-full text-xs py-1 px-3 w-15 text-white">
```

wlach (approved):
Looks good now! There are still improvements/changes possible (e.g. on reflection it's probably best to create a set of hardcoded "states" with colors associated with them as [bootstrap does with alerts](https://getbootstrap.com/docs/4.5/components/alerts/)) but I think this is good as a first pass.

PR #115, opened 2020-10-14
Add breadcrumbs to help user navigate (fixes #83)

### LINKED ISSUE #83: Add breadcrumbs to help users understand their place in the application
After navigating to this page I have no information on where I am on the application.

ping_page

Why not add a breadcrumb to help us navigate through the app in a proper way. 

something like this 

App / Fenix / Pings / Activation.

DawoodShahat:
This PR solves [https://github.com/mozilla/glean-dictionary/issues/83](https://github.com/mozilla/glean-dictionary/issues/83)

- Breadcrumb added
- Story for Breadcrumb component added

Here is what it looks like.

Screenshot 2020-10-14 at 8 02 23 AM

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -1,6 +1,7 @@
 <script>
   import page from "page";
   import Tailwindcss from "./Tailwindcss.svelte";
+  import Breadcrumb from "./components/Breadcrumb.svelte";
 
   // Pages
   import AppList from "./pages/AppList.svelte";
@@ -24,6 +25,7 @@
   page("/apps/:app/pings/:ping", setComponent(PingDetail));
   page("/apps/:app/metrics/:metric", setComponent(MetricDetail));
   page("/apps/:app", setComponent(AppDetail));
+  page("*", setComponent(Breadcrumb));
   page({ hashbang: true });
 </script>
 
@@ -41,6 +43,7 @@
       <i>Prototype</i>
     </a>
   </div>
+  <Breadcrumb {params} />
 </nav>
 
 <div class="container py-4 mx-auto">
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -41,7 +41,7 @@
             },
           ]
         : []),
-    ].filter(Boolean);
+    ];
   });
 
   function setComponent(c) {
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -13,12 +13,13 @@
 
   let component;
   let params;
-  let links = [{ url: "/", name: "apps" }];
+  let links = [];
 
   afterUpdate(() => {
     const { app, ping, metric } = params;
 
-    links = [];
+    links = [{ url: "/", name: "apps" }];
+
     if (app) {
       links.push({ url: `/apps/${app}/`, name: app });
     }
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -16,27 +16,32 @@
   let links = [];
 
   afterUpdate(() => {
-    const { app, ping, metric, bigquery } = params;
+    const { app, ping, metric, table } = params;
 
-    links = [{ url: "/", name: "apps" }];
-
-    let link = app && { url: `/apps/${app}/`, name: app };
-    links.push(link);
-
-    link =
-      (ping && { url: `/apps/${app}/pings/${ping}/`, name: ping }) ||
-      (metric && { url: `/apps/${app}/metrics/${metric}/`, name: metric });
-    links.push(link);
-
-    link = ping &&
-      bigquery && {
-        url: `/apps/${app}/tables/${ping}/bigquery`,
-        name: bigquery,
-      };
-    links.push(link);
-
-    // remove unwanted values
-    links = links.filter(Boolean);
+    links = [
+      ...(app
+        ? [
+            { url: "/", name: "apps" },
+            { url: `/apps/${app}/`, name: app },
+          ]
+        : []),
+      ...(ping ? [{ url: `/apps/${app}/pings/${ping}/`, name: ping }] : []),
+      ...(metric
+        ? [{ url: `/apps/${app}/metrics/${metric}/`, name: metric }]
+        : []),
+      ...(table
+        ? [
+            {
+              url: `/apps/${app}/pings/${table}/`,
+              name: table,
+            },
+            {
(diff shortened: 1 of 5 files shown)


DawoodShahat:
@wlach One question, I wasn't able to grab table param when navigated to BigQuery table for a particular ping. The table param isn't added to the params object that is passed as a prop to every component. Take a look when possible.

wlach (changes requested):
This is the right general idea, but I'd like to see a slightly different approach. See comment.

DawoodShahat:
Thanks for the insight. I will add the suggestions.

src/App.svelte
```diff
@@ -11,6 +13,29 @@
 
   let component;
   let params;
+  let links = [];
+
+  afterUpdate(() => {
+    const { app, ping, metric } = params;
+
+    links = [{ url: "/", name: "apps" }];
```
wlach, line 21: I'd bet you can simplify this function a bit (i.e. removing a bunch of the if...else) by using the ternary operator (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) and array destructuring (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).

stories/breadcrumb.stories.js
```diff
@@ -0,0 +1,37 @@
+import Breadcrumb from "../src/components/Breadcrumb.svelte";
+
+const AppsLinks = [{ url: "/", name: "apps" }];
+
+const PingLinks = [
```
wlach, line 5: Again, you could use array destructuring here.

src/App.svelte
```diff
@@ -11,6 +13,29 @@
...
+
+    if (app) {
+      links.push({ url: `/apps/${app}/`, name: app });
+    }
+
+    if (ping) {
+      links.push({
+        url: `/apps/${app}/pings/${ping}`,
+        name: `pings / ${ping}`,
```
wlach, line 30: I think we can probably skip the `pings /` part of this.

wlach (changes requested):
This is much closer to what I was expecting, thanks! See comments inline. I think you may have forgotten to add breadcrumbs for the bigquery table pages?

DawoodShahat:
Thanks, @wlach I will work on the suggestions. as for the big query page,  I wasn't able to grab table param when navigated to BigQuery table for a particular ping. The table param isn't added to the params object that is passed as a prop to every component. But I will try to come up with the solution.

DawoodShahat:
Changes made based on @wlach review:

- Updated ```afterUpdate()``` function to avoid unnecessary conditionals
- Updated route for Big Query table, so that its param is exposed through ```params.bigquery``` 
- Added story for ```TableDetail.svelte``` page

src/App.svelte
wlach: To be a little more concrete about it, here's what I'd do here:

```js
const links = [
  ...[{ url: "/", name: "apps" }], 
  ...ping ? [{ url: `/apps/${app}/pings/${ping}/`, name: ping }] : [],
  ...metric ? [{ url: `/apps/${app}/metrics/${metric}/`, name: metric }] : [],
];
```

(this example doesn't include BigQuery, leaving that as an exercise for you!)

src/App.svelte
```diff
@@ -20,10 +47,11 @@
   }
 
   page("/", setComponent(AppList));
-  page("/apps/:app/tables/:ping", setComponent(TableDetail));
+  page("/apps/:app/tables/:ping/:bigquery", setComponent(TableDetail));
```
wlach, line 50: This should actually be `:table` (using `:ping` was actually a mistake)

```suggestion
  page("/apps/:app/tables/:table", setComponent(TableDetail));
```

src/App.svelte
```diff
@@ -20,10 +47,11 @@
...
   page("/apps/:app/pings/:ping", setComponent(PingDetail));
   page("/apps/:app/metrics/:metric", setComponent(MetricDetail));
   page("/apps/:app", setComponent(AppDetail));
+  page("*", setComponent(Breadcrumb));
```
wlach, line 54: Did you mean to add this? It seems like an error and might conflict with some other work that's happening.

src/components/Breadcrumb.svelte
```diff
@@ -0,0 +1,12 @@
+<script>
+  export let links;
+</script>
+
+<style>
+</style>
+
```
wlach, line 7: You can just skip this section if you're not going to define anything.

```suggestion
```

wlach (changes requested):
This is looking pretty good, code wise! I think it's better if we put the breadcrumbs in the top-left, I actually didn't see them when I first ran your PR. I kind of like the approach I took with Mission Control, which people have generally responded postively to:

image

https://missioncontrol.telemetry.mozilla.org/#/release/windows/main_crashes

(don't worry about reproducing the styling, just the idea is important at this point)

src/App.svelte
DawoodShahat: Oh, my bad. I wanted it to be removed.

src/App.svelte
wlach: Note to do this you'll also need to make some (very minor) changes to the table component: https://github.com/mozilla/glean-dictionary/blob/9990cef3e5e04a06aaeb6d69470516325e5e8535/src/pages/TableDetail.svelte#L6

DawoodShahat:
Here is a preview

Screenshot 2020-10-21 at 1 46 40 AM

src/App.svelte
```diff
@@ -11,6 +13,36 @@
...
+              url: `/apps/${app}/pings/${table}/`,
+              name: table,
+            },
+            {
+              url: `/apps/${app}/tables/${table}/`,
+              name: `${table} table`,
+            },
+          ]
+        : []),
+    ].filter(Boolean);
```
wlach, line 44: I am pretty sure this filter should no longer be necessary?

```suggestion
    ]
```

wlach (changes requested):
This looks great! Just one nit

wlach (approved):

wlach:
Thank you!

Issue #88, opened 2020-10-11
Pagination with Next Previous buttons


meghajain-1711:
For pages like http://localhost:5000/#!/apps/fenix , navigation becomes too difficult if the data is too much. 

Apart from filtering as mentioned in another issue, we can also add Next Previous Buttons to show 1 to 100 or page limit. 

This will be useful when user does not have the exact search term to filter.

LilyLME:
Hello @wlach @meghajain-1711 can I work on this?

wlach:
> Hello @wlach @meghajain-1711 can I work on this?

Yup! We're trying to get applicants to do at least one frontend bug as part of their work and this is a good candidate. Thank you @meghajain-1711 for filing.

Pagination that looks something like this would be good:

![image](https://user-images.githubusercontent.com/20569/96122116-05931600-0ebf-11eb-81ea-17a2039c771d.png)

Some inspiration for creating pagination components can be found here:

https://tailwindcomponents.com/component/pagination-example

As part of this issue, I would like to see a story added for this component, you can see an example of that here:

https://github.com/mozilla/glean-dictionary/blob/main/stories/schemaviewer.stories.js

Iinh:
@wlach I think we can close this issue now?

wlach:
> @wlach I think we can close this issue now?

Yup! Thank you @LilyLME and @Iinh for your work here.

PR #119, opened 2020-10-16
Implemented pagination on app details page

LilyLME:
As required in #88 pagination has been added to reduce scroll on very long pages.

## Tasks

- [ ] Story book
- [x] Use TailwindCSS

### DIFF
--- a/src/components/Pagination.svelte
+++ b/src/components/Pagination.svelte
@@ -1,0 +1,85 @@
+<script>
+  export let currentPage;
+  export let lastPage;
+  export let from;
+  export let to;
+  export let total;
+  import { createEventDispatcher } from "svelte";
+
+  const dispatch = createEventDispatcher();
+  function range(size, startAt = 0) {
+    return [...Array(size).keys()].map((i) => i + startAt);
+  }
+  function changePage(page) {
+    if (page !== currentPage) {
+      dispatch("changePage", page);
+    }
+  }
+</script>
+
+<style>
+  .pagination {
+    display: flex;
+    justify-content: center;
+  }
+  .pagination ul {
+    display: flex;
+    padding-left: 0;
+    list-style: none;
+  }
+  .pagination li a {
+    position: relative;
+    display: block;
+    padding: 0.5rem 0.75rem;
+    margin-left: -1px;
+    line-height: 1.25;
+    background-color: #fff;
+    border: 1px solid #dee2e6;
+  }
+  .pagination li.active a {
+    color: #fff;
+    background-color: #007bff;
+    border-color: #007bff;
+  }
+  .pagination li.disabled a {
+    color: #6c757d;
+    pointer-events: none;
+    cursor: auto;
+    border-color: #dee2e6;
+  }
+</style>
+
+<p>
+  Page
+  <code>{currentPage}</code>
+  of
+  <code>{lastPage}</code>
+  (<code>{from}</code>
+  -
+  <code>{to}</code>
+  on
+  <code>{total}</code>
+  items)
+</p>
+
+<nav class="pagination">
+  <ul>
+    <li class={currentPage === 1 ? 'disabled' : ''}>
+      <a href={'#'} on:click|preventDefault={() => changePage(currentPage - 1)}>
+        <span aria-hidden="true">«</span>
+      </a>
+    </li>
+    {#each range(lastPage, 1) as page}
+      <li class={page === currentPage ? 'active' : ''}>
+        <a
+          href={'#'}
+          on:click|preventDefault={() => changePage(page)}>{page}</a>
+      </li>
+    {/each}
+    <li class={currentPage === lastPage ? 'disabled' : ''}>
+      <a href={'#'} on:click|preventDefault={() => changePage(currentPage + 1)}>
+        <span aria-hidden="true">»</span>
+      </a>
+    </li>
+  </ul>
+</nav>
--- a/src/components/Pagination.svelte
+++ b/src/components/Pagination.svelte
@@ -31,7 +31,7 @@
     for (let i = 0; i < page - 1; i += 1) {
       from += pages[i].length;
     }
-    let to = from + pages[page - 1].length;
+    let to = from + pages[page - 1].length - 1;
     return {
       currentPage: page,
       from,
--- a/src/components/Pagination.svelte
+++ b/src/components/Pagination.svelte
@@ -3,7 +3,9 @@
     let result = [];
     let localData = array.slice();
     for (let i = totalPages; i > 0; i -= 1) {
-      result.push(localData.splice(0, Math.max(Math.floor(localData.length / i), 100)));
+      result.push(
+        localData.splice(0, Math.max(Math.floor(localData.length / i), 100))
+      );
     }
     return result;
   };
--- a/src/components/Pagination.svelte
+++ b/src/components/Pagination.svelte
@@ -1,21 +1,26 @@
 <script context="module">
-  export const paginateData = (array, totalPages) => {
+  let DEFAULT_ITEMS_PER_PAGE = 20;
+  export const paginateData = (
+    array,
+    totalPages,
+    perPage = DEFAULT_ITEMS_PER_PAGE
+  ) => {
     let result = [];
     let localData = array.slice();
     for (let i = totalPages; i > 0; i -= 1) {
       result.push(
-        localData.splice(0, Math.max(Math.floor(localData.length / i), 100))
+        localData.splice(0, Math.max(Math.floor(localData.length / i), perPage))
       );
     }
     return result;
   };
 
-  export const makePages = (page, data, perPage = 100) => {
+  export const makePages = (page, data, perPage = DEFAULT_ITEMS_PER_PAGE) => {
     if (data.length === 0) return [];
     let total = data.length;
     let currentPage = page;
     let lastPage = Math.ceil(total / perPage);
-    let pages = paginateData(data, lastPage);
+    let pages = paginateData(data, lastPage, perPage);
     let from = page + perPage * (page - 1);
     let to = page * perPage;
     return {
--- a/src/components/Pagination.svelte
+++ b/src/components/Pagination.svelte
@@ -3,7 +3,7 @@
     let result = [];
     let localData = array.slice();
     for (let i = totalPages; i > 0; i -= 1) {
-      result.push(localData.splice(0, Math.ceil(localData.length / i)));
+      result.push(localData.splice(0, Math.max(Math.floor(localData.length / i), 100)));
     }
     return result;
   };
@@ -13,9 +13,9 @@
     let total = data.length;
     let currentPage = page;
     let lastPage = Math.ceil(total / perPage);
+    let pages = paginateData(data, lastPage);
     let from = page + perPage * (page - 1);
     let to = page * perPage;
-    let pages = paginateData(data, lastPage);
     return {
       total,
       currentPage,
@@ -27,7 +27,7 @@
   };
 
   export const goToPage = (page, pages) => {
-    let from = 0;
+    let from = 1;
     for (let i = 0; i < page - 1; i += 1) {
       from += pages[i].length;
     }
--- a/src/components/Pagination.svelte
+++ b/src/components/Pagination.svelte
@@ -16,7 +16,7 @@
   };
 
   export const makePages = (page, data, perPage = DEFAULT_ITEMS_PER_PAGE) => {
-    if (data.length === 0) return [];
+    if (data.length === 0) return {};
     let total = data.length;
     let currentPage = page;
     let lastPage = Math.ceil(total / perPage);
--- a/src/components/Pagination.svelte
+++ b/src/components/Pagination.svelte
@@ -1,53 +1,66 @@
+<script context="module">
+  export const paginateData = (array, totalPages) => {
+    let result = [];
+    let localData = array.slice(); // Don't fuckup the array!
+    for (let i = totalPages; i > 0; i -= 1) {
+      result.push(localData.splice(0, Math.ceil(localData.length / i)));
+    }
+    return result;
+  };
+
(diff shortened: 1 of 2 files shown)


src/pages/AppDetail.svelte
```diff
@@ -3,15 +3,46 @@
...
+      lastPage = Math.ceil(total / perPage);
+      from = data.page + perPage * (data.page - 1);
+      to = perPage * data.page;
+      metrics = breakUp(app.metrics, lastPage);
+    }
+    from = data.page + perPage * (data.page - 1);
+    to = perPage * data.page;
+    currentPage = data.page;
+    filteredMetrics = metrics[data.page - 1];
+  };
```
wlach, line 42: Since we're likely to want to use pagination in many places in the application, it would be better if this functionality were located in the pagination component itself. I haven't used this technique before, but it looks like it is possible to export a function from a svelte component:

https://flaviocopes.com/svelte-export-from-component/

src/pages/AppDetail.svelte
```diff
@@ -69,4 +100,15 @@
...
+  {#if total > perPage}
+    <Pagination
+      {currentPage}
+      {from}
+      {lastPage}
+      {perPage}
+      {to}
+      {total}
+      on:changePage={(ev) => loadPage({ page: ev.detail })} />
+  {/if}
```
wlach, line 113: I haven't thought about this too deeply, but this seems like a fairly complex interface. Might be worth looking at some other examples for a simpler implementation: https://github.com/fouita/svelte-tw-pagination

wlach (commented):
This looks like a good start. Just some initial thoughts.

src/pages/AppDetail.svelte
LilyLME: Makes sense, I would move the functionality then.

src/pages/AppDetail.svelte
LilyLME: @wlach the implementation is quite similar, however the interface here looks a little complex because we are working with `json` and not a paginated API so my idea is to get the JSON only once and use JavaScript to split it into the sections we need.

wlach: In general it's a little better to do most post-processing (e.g. munging a json data structure) before sending any data to a component like this, as that keeps the interface simple (following the [seperation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) design principle). Does that make sense?

LilyLME: Managed to move core pagination functions to the pagination component

LilyLME:
## Pagination view

src/pages/AppDetail.svelte
```diff
@@ -3,17 +3,45 @@
...
   const URL = `data/${params.app}/index.json`;
   let app;
+  let to;
+  let from;
+  let total;
+  let metrics = [];
+  let pages = [];
+  let paginationData;
+  let lastPage;
+  let perPage = 100;
```
wlach, line 21: Since this never varies, I think we can leave it out. Can you make 100 the default in the Pagination component?

src/components/Pagination.svelte
```diff
@@ -0,0 +1,124 @@
+<script context="module">
+  export const paginateData = (array, totalPages) => {
+    let result = [];
+    let localData = array.slice();
+    for (let i = totalPages; i > 0; i -= 1) {
+      result.push(localData.splice(0, Math.ceil(localData.length / i)));
+    }
+    return result;
+  };
```
wlach, line 9: It occurs to me that lodash's chunk would do this for us:

https://lodash.com/docs/4.17.15#chunk

(more a point of information, we don't have lodash as a dependency yet so I think we can leave using it to a followup)

src/components/Pagination.svelte
```diff
@@ -0,0 +1,124 @@
...
+  -
+  <code>{to}</code>
+  on
+  <code>{total}</code>
+  items)
+</p>
+
+<div class="flex flex-col items-center my-12">
+  <div class="flex text-gray-700">
+    <div
```
wlach, line 80: I think we should probably disable these in the case that user is already on the first/last page.

Though on second thought -- maybe it's best actually to just leave these out for now, as I don't think they add a lot of value over just showing the individual pages as links. We can consider re-adding them later.

src/components/Pagination.svelte
```diff
@@ -0,0 +1,124 @@
...
+        stroke-linejoin="round"
+        class="feather feather-chevron-left w-6 h-6">
+        <polyline points="15 18 9 12 15 6" />
+      </svg>
+    </div>
+    <div class="flex h-12 font-medium bg-gray-200">
+      {#each range(lastPage, 1) as page}
+        <div
+          on:click|preventDefault={() => changePage(page)}
+          class="w-12 md:flex justify-center items-center hidden  cursor-pointer leading-5 transition duration-150 ease-in {page === currentPage ? 'bg-teal-600 text-white' : ''}">
```
wlach, line 101: Let's skip the transition/animation parts (I'd rather do that holistically when more of the application is in place)

```suggestion
          class="w-12 md:flex justify-center items-center hidden cursor-pointer leading-5 {page === currentPage ? 'bg-teal-600 text-white' : ''}">
```

src/components/Pagination.svelte
```diff
@@ -0,0 +1,124 @@
...
+          {page}
+        </div>
+      {/each}
+    </div>
+    <div
```
wlach, line 106: Likewise let's take this out for now.

src/pages/AppDetail.svelte
wlach: This is better, however it's still a little complex. I think the pagination component really only needs two variables as input:

1. the total number of elements
2. the requested number of elements per page (optional, as mentioned above)
3. the current page that the user is on

The responsibility of rendering the current page of data can live in this component, maybe using a helper function defined in the pagination component that can take a full list of items and (2) and (3) above and return the list of items to be rendered.

src/components/Pagination.svelte
```diff
@@ -0,0 +1,124 @@
...
+      currentPage: page,
+      from,
+      to,
+      page: pages[page - 1],
+    };
+  };
+</script>
+
+<script>
+  import { createEventDispatcher } from "svelte";
```
wlach, line 45: I don't think we actually need an event dispatcher here. Try just exporting an optional callback function -- see the FilterInput component for an example of this:

https://github.com/mozilla/glean-dictionary/blob/9990cef3e5e04a06aaeb6d69470516325e5e8535/src/components/FilterInput.svelte

wlach (changes requested):
Thanks for your persistence on this! It turns out this is a rather complex issue to solve. I've put in some more feedback, hopefully it is helpful.

src/pages/AppDetail.svelte
LilyLME: Sure, noted.

src/components/Pagination.svelte
LilyLME: I would create a follow up issue for that if necessary.

src/components/Pagination.svelte
wlach: Was referring to the second chevron, which is gone now. Thanks!

src/components/Pagination.svelte
LilyLME: @wlach replacing the event dispatcher has proved a little complex given the need to pass some parameters around unlike with the example shared.

Don't know if I am missing something, I checked a few self-contained `paginators` and they use this approach as well. Please shed more light

src/pages/AppDetail.svelte
LilyLME: We have two extra params used to count items display on a page and show helpful text. "Showing x of n items" [It's not not completely necessary and be taken out]

src/pages/AppDetail.svelte
```diff
@@ -5,19 +5,46 @@
...
 
   export let params;
   const URL = `data/${params.app}/index.json`;
   let app;
+  let to;
+  let from;
+  let total;
+  let metrics = [];
+  let pages = [];
+  let paginationData;
```
wlach, line 21: I think `paginationState` might be a slightly better variable name, since "data" is quite ambiguous.

```suggestion
  let paginationState;
```

src/pages/AppDetail.svelte
```diff
@@ -5,19 +5,46 @@
...
+  let lastPage;
+  let currentPage = 1;
```
wlach, line 23: I wonder if we could replace most of this with a structure called `paginationState`, and just do the book-keeping/updating of the variables there. That would eliminate the need to pass a bunch of individual variables back to the pagination component.

src/pages/AppDetail.svelte
```diff
@@ -5,19 +5,46 @@
...
   let filteredMetrics;
   let filteredPings;
 
+  function loadPage(args) {
+    let tempPaginationData = goToPage(args.page, pages);
+    from = tempPaginationData.from;
+    to = tempPaginationData.to;
+    filteredMetrics = tempPaginationData.page;
+    currentPage = args.page;
```
wlach, line 32: If you follow my suggestion above, we can probably simplify this function to something like:

```suggestion
    paginationState = goToPage(args.page, pages);
    currentPage = args.page;
```

src/pages/AppDetail.svelte
wlach: I think we're probably good here now, if we move to the `paginationState` approach that I mentioned above.

src/components/Pagination.svelte
wlach: To be 100% honest with you, I am somewhat new to Svelte as well. 😄 Ok, let's assume this approach is good for now. We can always circle back and improve the component later.

wlach (changes requested):
Thanks for your persistence on this, and sorry for the slowness of my response. I think this is actually pretty close, after taking a second look at this, I think you may have been right about some aspects (e.g. the need to keep more state in the module calling the pagination component). To keep things more managable, I think we can keep track of the various variables inside a `paginationState` variable. Could you give that a try?

src/pages/AppDetail.svelte
LilyLME: Excellent!

src/pages/AppDetail.svelte
LilyLME: @wlach that has been done, it makes things way cleaner.

LilyLME:
Hello @wlach, I have updated this as per requirements discussed. The calling module maintains it's level of state but now very neatly embedded in a single state object.

I would have loved if the components itself could let objects with properties be exported, but looks like it's a difficult problem trying to figure out how attributes within objects change :
See : 
- https://github.com/sveltejs/svelte/issues/3381
- https://svelte.dev/tutorial/spread-props

src/components/Pagination.svelte
```diff
@@ -0,0 +1,92 @@
+<script context="module">
+  export const paginateData = (array, totalPages) => {
```
wlach, line 2: This should have a "perPage" option, like `makePages` below

src/components/Pagination.svelte
```diff
@@ -0,0 +1,92 @@
...
+    let localData = array.slice();
+    for (let i = totalPages; i > 0; i -= 1) {
+      result.push(
+        localData.splice(0, Math.max(Math.floor(localData.length / i), 100))
+      );
+    }
+    return result;
+  };
+
+  export const makePages = (page, data, perPage = 100) => {
```
wlach, line 13: Can you factor out the `perPage` option to a variable like `DEFAULT_ITEMS_PER_PAGE`. We should make this easier to tweak. Also, I think the default should be much smaller, like maybe 20 or so (so things appear on the screen at once)

wlach (changes requested):
After testing this out a bit, I'm noticing a few problems:

1. There's an unpaginated list of metrics on top of the paginated one (might just be a merging artifact), this should be easy to fix.
2. The pagination behaviour doesn't play nicely with the filtering. Basically we want to paginate through any list of items that the user has filtered (you can test with a string like "search" with the Fenix browser: http://localhost:5000/#!/apps/fenix)

Thanks again for your hard work on this.

src/components/Pagination.svelte
LilyLME: Updated.

LilyLME:
@wlach all last change requests made. The pagination now plays nicely with the `filteredMetrics` also.

wlach (changes requested):
This is getting closer for sure, but I found a bug:

* Navigate to: http://localhost:5000/#!/apps/firefox-android-release
* Skip to "page 6" on metrics
* Type "search" into the filter box
* Try to skip back to "page 1"

Neither the search or the skip will work and I see this in the console:

image

LilyLME:
Hello @wlach sorry for taking a while with this, can you review again and check for the last the bug so we move on?

wlach (approved):
Thanks, this works great now!

wlach:
Some quick followup tasks for this:

* there's the "chunk" thing we talked about (https://github.com/mozilla/glean-dictionary/pull/119#discussion_r508596663)
* we should also look into adding this to the ping page: https://deploy-preview-119--glean-dictionary-dev.netlify.app/#!/apps/firefox-android-release/pings/metrics

@LilyLME feel free to work on these (but obviously feel no obligation). I'll make a note to file some issues on these in the future (today is technically a holiday for me)

LilyLME:
@wlach that's fine, Since the simpler and faster thing to do would be adding the pagination to ping page, I would rather start with that before looking a lodash' chunk.

- You also had a storybook requirement for the pagination component is that pressing?

wlach:
> @wlach that's fine, Since the simpler and faster thing to do would be adding the pagination to ping page, I would rather start with that before looking a lodash' chunk.

For sure. For what it's worth, I think the lodash chunk task should be quite easy. Mostly just a matter of running `npm install lodash` and then using it.

>     * You also had a storybook requirement for the pagination component is that pressing?

Ah yes! We definitely want that too.

PR #120, opened 2020-10-16
Gather ping types from libraries, as well as applications (partially fixes #109)

### LINKED ISSUE #109: We should gather all metric/ping types, not just those specified by the applications

Currently we're only gathering metrics and ping data from the applications themselves, not those specified by the libraries (e.g. the `events` ping is part of glean-core: https://github.com/mozilla/glean/blob/2261845761251d91b6968f29846dc3aabbc0cc45/glean-core/pings.yaml#L80)

Mozilla Schema Generator (which does something similar to what we do) enumerates dependencies for each application and gathers data for them e.g. https://github.com/mozilla/mozilla-schema-generator/blob/1264148ff82adc3357de33a25541f1313f93449f/mozilla_schema_generator/glean_ping.py#L126

We should modify our `build-glean-metadata.py` script (https://github.com/mozilla/glean-dictionary/blob/main/scripts/build-glean-metadata.py) with similar logic.

mhmohona:
Hello @wlach ! I am super confused about this issue. :sweat_smile: 

Here I am supposed to collect all metric/ping types. At first look, it seemed I can do that just by removing the [filter](https://github.com/mozilla/glean-dictionary/blob/main/scripts/build-glean-metadata.py#L22) from line 22 in `build-glean-metadata.py` file. I could also make separate lists like applications on the front page of Glean Dictionary prototype. 
But if it was like that, you would not ask to look at Mozilla Schema Generator. 

I am confused about why there is the necessity of using the logic from Schema Generator, how to show them. Surely here you are not looking for just changing the architecture of `build-glean-metadata.py` with self function here? 

It would be really helpful if you would give me some more hints about how to approach issue #109

### DIFF
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+from requests import HTTPError
 import json
 import os
 
@@ -10,9 +11,12 @@ PROBE_INFO_BASE_URL = "https://probeinfo.telemetry.mozilla.org"
 REPO_URL = PROBE_INFO_BASE_URL + "/glean/repositories"
 PINGS_URL_TEMPLATE = PROBE_INFO_BASE_URL + "/glean/{}/pings"
 METRICS_URL_TEMPLATE = PROBE_INFO_BASE_URL + "/glean/{}/metrics"
-DEPENDENCIES_URL_TEMPLATE= PROBE_INFO_BASE_URL + "/glean/{}/dependencies"
+DEPENDENCIES_URL_TEMPLATE = PROBE_INFO_BASE_URL + "/glean/{}/dependencies"
+
 OUTPUT_DIRECTORY = os.path.join("public", "data")
 
+default_dependencies = ['glean']
+ignore_pings = {"all-pings", "all_pings", "default", "glean_ping_info", "glean_client_info"}
 
 # we get the repos ourselves instead of using GleanPing.get_repos()
 # to get all the metadata associated with the repo
@@ -76,7 +80,38 @@ for repo in list(repos):
             )
         )
 
-    ping_data = requests.get(PINGS_URL_TEMPLATE.format(app_name)).json()
+    def get_dependencies():
+        try:
+            dependencies = requests.get(DEPENDENCIES_URL_TEMPLATE.format(app_name)).json() 
+
+        except HTTPError:
+            return default_dependencies
+
+        dependency_library_names = list(dependencies.keys()) 
+        repos_by_dependency_name = {}
+
+        for repo in repos:
+            for library_name in repo.get('library_names', []):
+                repos_by_dependency_name[library_name] = repo['name'] 
+
+        dependencies = []
+
+        for name in dependency_library_names:
+            if name in repos_by_dependency_name:
+                dependencies.append(repos_by_dependency_name[name])               
+
+        if len(dependencies) == 0:
+            return default_dependencies
+        
+        return dependencies 
+
+    app_ping = requests.get(PINGS_URL_TEMPLATE.format(app_name)).json()
+    
+    for dependency in get_dependencies():
+        dependency_ping = requests.get(PINGS_URL_TEMPLATE.format(dependency)).json()        
+
+    ping_data = {**app_ping, **dependency_ping}    
+    
     for (ping_name, ping_data) in ping_data.items():
         app_data["pings"].append(
             {"name": ping_name, "description": ping_data["history"][-1]["description"]}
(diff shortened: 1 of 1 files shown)


wlach:
> Here I am supposed to collect all metric/ping types. At first look, it seemed I can do that just by removing the [filter](https://github.com/mozilla/glean-dictionary/blob/main/scripts/build-glean-metadata.py#L22) from line 22 in `build-glean-metadata.py` file. I could also make separate lists like applications on the front page of Glean Dictionary prototype.

So the filter logic is to stop us from listing dependencies (like the glean library) as *applications*. What we want to do instead is add the pings (and metrics associated with them) that Glean applications get from their dependencies ([libraries](https://en.wikipedia.org/wiki/Library_(computing))) -- this is what the schema generator does [around here](https://github.com/mozilla/mozilla-schema-generator/blob/1264148ff82adc3357de33a25541f1313f93449f/mozilla_schema_generator/glean_ping.py#L130).

Hopefully that helps a bit? If not feel free to ask in the #glean-dictionary channel. Even if I'm not around someone there might be able to help you.

scripts/build-glean-metadata.py
```diff
@@ -10,8 +11,12 @@
 REPO_URL = PROBE_INFO_BASE_URL + "/glean/repositories"
 PINGS_URL_TEMPLATE = PROBE_INFO_BASE_URL + "/glean/{}/pings"
 METRICS_URL_TEMPLATE = PROBE_INFO_BASE_URL + "/glean/{}/metrics"
+DEPENDENCIES_URL_TEMPLATE = PROBE_INFO_BASE_URL + "/glean/{}/dependencies"
+
 OUTPUT_DIRECTORY = os.path.join("public", "data")
 
+default_dependencies = ['glean']
+ignore_pings = {"all-pings", "all_pings", "default", "glean_ping_info", "glean_client_info"}
```
wlach, line 19:
```suggestion
```

scripts/build-glean-metadata.py
```diff
@@ -10,8 +11,12 @@
 REPO_URL = PROBE_INFO_BASE_URL + "/glean/repositories"
 PINGS_URL_TEMPLATE = PROBE_INFO_BASE_URL + "/glean/{}/pings"
 METRICS_URL_TEMPLATE = PROBE_INFO_BASE_URL + "/glean/{}/metrics"
+DEPENDENCIES_URL_TEMPLATE = PROBE_INFO_BASE_URL + "/glean/{}/dependencies"
+
 OUTPUT_DIRECTORY = os.path.join("public", "data")
 
+default_dependencies = ['glean']
```
wlach, line 18: Let's uppercase this to make sure it's part of the default definitions.

```suggestion
DEFAULT_DEPENDENCIES = ['glean']
```

scripts/build-glean-metadata.py
```diff
@@ -75,7 +80,38 @@
             )
         )
 
-    ping_data = requests.get(PINGS_URL_TEMPLATE.format(app_name)).json()
+    def get_dependencies():
```
wlach, line 83: Eventually we may want to move to using mozilla-schema-generator as a library, though I think it makes sense to just copy its code for now (as you are doing). However, in this context it doesn't really make sense for it to be a function, since it relies on outlying context (like app_name). Can you take out this definition and just create a `dependencies` variable that we use directly?

Can you also make a note about the original source of this function? (https://github.com/mozilla/mozilla-schema-generator/blob/master/mozilla_schema_generator/glean_ping.py)

wlach (changes requested):
This seems to be working as expected, thank you! However a few small nits about the implementation. Also, CI is failing: please run `make lint-fix`.

mhmohona:
> This seems to be working as expected, thank you! However a few small nits about the implementation. Also, CI is failing: please run `make lint-fix`.
I ran `make lint-fix` but still it is failing the CircleCI  test. What should I do? :( 
Screenshot from 2020-11-06 16-37-39

scripts/build-glean-metadata.py
mhmohona: > Can you also make a note about the original source of this function? (https://github.com/mozilla/mozilla-schema-generator/blob/master/mozilla_schema_generator/glean_ping.py)


Have I added the comment correctly? :sweat_smile:

wlach:
> > This seems to be working as expected, thank you! However a few small nits about the implementation. Also, CI is failing: please run `make lint-fix`.
> 
> I ran `make lint-fix` but still it is failing the CircleCI test. What should I do? :(
> ...

That's quite strange. It looks like you might have some invalid files in your checkout. Running `git clean -f` (before the lint-fix command) might help. If that doesn't work, what does the output of `git status` look like for you?

mhmohona:
> That's quite strange. It looks like you might have some invalid files in your checkout. Running `git clean -f` (before the lint-fix command) might help. If that doesn't work, what does the output of `git status` look like for you?

@wlach, as I got error previously, to avoid other file's error, for last commit I did not checkout. I copied the whole `build-glean-metadata.py` file's code from VSCode and pasted directly in browser.

scripts/build-glean-metadata.py
```diff
@@ -75,7 +79,37 @@
...
+
+    dependency_library_names = list(dependencies.keys())
+    repos_by_dependency_name = {}
+
+    for repo in repos:
+        for library_name in repo.get("library_names", []):
+            repos_by_dependency_name[library_name] = repo["name"]
+
+    dependencies = []
+
```
wlach, line 98: These lines aren't necessary (dependencies were already processed above)

```suggestion
```

wlach (approved):
I did some final testing, and this works as expected for the pings. However, we're still not gathering the metric data. Since this PR has been open for a while, I think I'm going to go ahead and merge it as-is-- we can handle metric data in a followup.

PR #121, opened 2020-10-16
Not Found component created to better handle cases when items do not exist (fixes #108)

### LINKED ISSUE #108: Better handle cases where application/pings/metrics/table do not exist
Currently we just silently fail if the user navigates to an entity that does not exist. e.g.: 

http://localhost:5000/#!/apps/burnham2
http://localhost:5000/#!/apps/fenix-nightly/pings/activation-doesnotexist

It would be better if we displayed some kind of friendly error page saying something like "Could not find application `burnham2`" or "Could not find ping `application-doesnotexist`". I don't expect this to happen frequently but this sort of thing can happen occasionally (e.g. if an application is added and then withdrawn)

To accomplish this task, you'll want to create a new Svelte component to cover this functionality and update each page to use/display it in the event that fetching information fails.

nonbinaryfrog:
I keep getting a ReferenceError so I assume I have not created the component correctly.


src/components/NotFoundPage.svelte
```diff
@@ -0,0 +1,30 @@
...
+ let app = new App({
+    target: document.body,
+    props: {
+        app: AppDetail,
+        list: AppList,
+        ping: PingDetail,
+        metric: MetricDetail,
+        table: TableDetail,
+    }
+})
```
acmiyaguchi, line 17: We shouldn't be creating a new app instance in a component, which is generally reserved for `main.js`. This is probably why you're getting strange issues.

In the actual pages, I would expect to something like the following in the main code. Of course, this can be implemented in a number of ways with similar behavior.

```js
{#await some_promise}
    ....
{:then ...}
   ...
{:error error}
  <NotFoundPage name={....}
{/awai}
```

The component defined here is included when fetching the resource fails. I'd recommend taking a look at the svelte tutorial on building components with props: https://svelte.dev/tutorial/declaring-props

src/components/SchemaNode.svelte
```diff
@@ -9,7 +9,9 @@
     <span
       class="text-gray-700">{parentFields.join('.')}{parentFields.length ? '.' : ''}</span><span>{node.name}</span>
   </p>
-  {#if node.description}
+ {#if node.description === 'The channel the application is being distributed on.'}
```
acmiyaguchi, line 12: Is this change related to the issue?

nonbinaryfrog: Sorry, no, that was for the initial test I needed to do before getting assigned a task.

src/components/SchemaNode.svelte
```diff
@@ -9,7 +9,9 @@
     <span
       class="text-gray-700">{parentFields.join('.')}{parentFields.length ? '.' : ''}</span><span>{node.name}</span>
   </p>
-  {#if node.description}
+  {#if node.description === 'The channel the application is being distributed on.'}
+    <p class="text-red-600 text-xs ml-2">{node.description}</p>
+  {:else}
```
acmiyaguchi, line 14: Please remove unrelated code changes.

src/components/NotFoundPage.svelte
```diff
@@ -0,0 +1,14 @@
...
+  export let pageName;
+  export let itemType;
+</script>
+
+<style>
+  h2 {
+    color: darkgray;
+    font-style: italic;
+  }
+</style>
```
acmiyaguchi, line 11: Lets leave styling out for now.

src/pages/AppDetail.svelte
```diff
@@ -9,9 +10,8 @@
     });
 </script>
 
-<h1>{params.app}</h1>
-
 {#if app}
+  <h1>{params.app}</h1>
```
acmiyaguchi, line 14: I'm pretty sure this code is out of date. Could you rebase/pull changes from the most recent version of the `main` branch?

acmiyaguchi (changes requested):
Please fix any merge conflicts by pulling in the most recent changes on `main` and any related linting errors caught by CI.

src/components/NotFoundPage.svelte
```diff
@@ -0,0 +1,14 @@
+<script>
```
acmiyaguchi, line 1: I think the name of the component should be something like `NotFound`, instead of `NotFoundPage`, since it's part of a page and not the page itself.

wlach:
@nonbinaryfrog let us know when you're ready for another review! it appears that CI is currently failing and there's a few conflicts to be resolved.

nonbinaryfrog:
I've resolved the error with Cl, I'll work on updating the code now

src/components/SchemaNode.svelte
```diff
@@ -11,6 +11,8 @@
   </p>
   {#if node.description}
     <p class="text-gray-600 text-xs ml-2">{node.description}</p>
+  {:else}
+    <p class="text-gray-600 text-xs ml-2">{node.description}</p>
```
acmiyaguchi, line 15: There should be no changes in this file.

acmiyaguchi (changes requested):
This generally looks good to me, aside from the SchemaNode changes.

acmiyaguchi:
Can you also update the PR title and description so it fits with https://github.com/mozilla/glean-dictionary/blob/main/CONTRIBUTING.md?

src/pages/AppDetail.svelte
```diff
@@ -81,4 +82,6 @@
       </li>
     {/each}
   </ul>
+{:else}
```
acmiyaguchi, line 85: One other small comment in this file -- it looks like the title in this page is outside of the `if` statement, so if it fails there will be two `<h1>` tags. Can you pull it into the conditional so it's consistent?

acmiyaguchi (commented):
FYI, if you mention the issue in a comment (e.g. #108), it makes it much easier to cross reference the PR with the original issue.

acmiyaguchi (approved):
r+

Issue #123, opened 2020-10-17
In ping page multiple emails are not shown properly


mhmohona:
On ping page, if there are multiple emails in the Notification Emails section, it is shown one after one without any gap which makes the email unreadable. 

![image](https://user-images.githubusercontent.com/14244685/96348041-c4218700-10c7-11eb-959c-f8387325d967.png)

Where to change: https://github.com/mozilla/glean-dictionary/blob/36c1fb7d5c09e2f2026b2f77f26f39dafaa9c3db/src/pages/PingDetail.svelte#L51

PR #124, opened 2020-10-17
Shown multiple emails for Notification Emails section in ping page in more readable way

mhmohona:
This is the fix for  #123. 

This is previous condition: 

image

After adding a space in between them notification emails now looks in proper formate. 

image


src/pages/PingDetail.svelte
```diff
@@ -48,7 +48,7 @@
         Notification Email{ping.notification_emails.length > 1 ? 's' : ''}
       </td>
       <td>
-        {#each ping.notification_emails as email}<span>{email}</span>{/each}
+        {#each ping.notification_emails as email}<span>{email + " "}</span>{/each}
```
wlach, line 51: A slightly more elegant way of doing this is to use `display: block`. Tailwind has a nice shorthand for this (https://tailwindcss.com/docs/display)

```suggestion
        {#each ping.notification_emails as email}<span class="block">{email}</span>{/each}
```

wlach (changes requested):
Thanks! I have a suggested improvement

src/pages/PingDetail.svelte
mhmohona: Yes! Your sugession indeed makes it better. ^_^

wlach (approved):

PR #126, opened 2020-10-18
Allow ping list to be filtered in application (fixes #125)

### LINKED ISSUE #125: Add filterable list of pings in Application Page
According to the proposal, there should be a filterable list of pings associated with the application.
proposal

Currently, in our application page there is no filter been added for pings. The page looks like this: 

Screenshot_2020-10-18 Glean Dictionary Prototype

Where to change:  https://github.com/mozilla/glean-dictionary/blob/main/src/pages/AppDetail.svelte

mhmohona:
This is a fix for #125

The updated version with filter for list of pings looks like this:

after adding filter


wlach (approved):
This looks reasonable. Note that it's generally good practice to phrase your pull requests in terms of what they do, as well as including a reference to the issue they fix (if there is one).

In this case, that would be something like this: "Allow ping list to be filtered in application (fixes #125)" (I'll do this for you this time)

PR #132, opened 2020-10-19
Bug references in ping view that are only numeric should link to Bugzilla

mhmohona:
For ping, in the Related Bugs area there is not the same formate always. Some only have numbers. 

code

These numbers are actually bugzilla bug ID. Previously bugzilla link was not added, so when we clicked on bug number it took us to an empty page.

Added link for Bugzilla:

activating bugzilla link


wlach (approved):

wlach:
We should probably add a svelte component to handle all of this. That can wait though...

PR #144, opened 2020-10-19
Add lint configuration for Python using flake8, black, isort and autoflake (fixes #54)

### LINKED ISSUE #54: Add Python linting to CI and make it easy to run
*This issue is intended as an initial onboarding task for potential outreachy applicants. Please do not work on it unless you have completed the initial qualification task and it has been assigned to you.*

We only have a minimal amount of python code so far, but we will accumulate more over time. To keep the quality level up, we should enable linting with [flake8](https://flake8.pycqa.org/en/latest/), [black](https://github.com/psf/black), and [isort](https://pypi.org/project/isort/).

This is a somewhat non-trivial issue, as it will involve:

* Setting up circleci to run python code (it only validates javascript at the moment)
* Writing up the correct linter configuration (always harder than it seems)

Some prior art which might be helpful is the mozregression repository where I recently added some linter code:

https://github.com/mozilla/mozregression/pull/575

Note however that it uses travis rather than circleci. You'll need to do some extra research to get this going with CircleCI.

Riju19:
This Pull request addresses Issue #54

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,7 @@
 version: 2.1
 orbs:
   node: circleci/node@2.0.1
+  python: circleci/python@0.2.1
 jobs:
   build-and-test:
     executor:
@@ -11,7 +12,16 @@ jobs:
       - node/install-packages
       - run: npm install
       - run: npm test
+  python-build-and-test:
+    executor: python/default
+    steps:
+      - checkout
+      - python/install-deps
+      - run:
+          name: Run lint
+          command: make lint
 workflows:
   build-and-test:
     jobs:
       - build-and-test
+      - python-build-and-test
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,14 @@
+.PHONY: help lint lint-fix
+help:
+	@echo "  lint                   Check style with flake8, black and isort"
+	@echo "  lint-fix               Format code with black and isort"
+
+lint: 
+	flake8
+	black --check .
+	isort --check-only .
+
+lint-fix:
+	autoflake -r -i --remove-all-unused-imports scripts
+	black .
+	isort .
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,13 @@
+[tool.black]
+line-length = 100
+# excludes handled in setup.cfg, since it's called via flake8
+
+[tool.isort]
+line_length = 100
+default_section = "THIRDPARTY"
+known_first_party = "glean-dictionary"
+# For compatibility with black:
+multi_line_output = 3
+include_trailing_comma = "True"
+force_grid_wrap = 0
+use_parentheses = "True"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,7 @@
+autoflake
+black
+flake8
+flake8-black
+isort
 mozilla-schema-generator
 stringcase
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,9 @@
+[flake8]
+max-line-length=100
+exclude=.git,__pycache__,venv,node_modules
+# E121,E123,E126,E226,E24,E704,W503,W504: Ignored in default pycodestyle config:
+# https://github.com/PyCQA/pycodestyle/blob/2.2.0/pycodestyle.py#L72
+# Our additions...
+# E129: visually indented line with same indent as next logical line
+# E203: pep8 is wrong, overridden by black (https://github.com/psf/black/issues/315)
+extend-ignore = E129,E203
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -5,8 +5,6 @@ import os
 
 import requests
 import stringcase
-from mozilla_schema_generator.glean_ping import GleanPing
-
 
 PROBE_INFO_BASE_URL = "https://probeinfo.telemetry.mozilla.org"
 REPO_URL = PROBE_INFO_BASE_URL + "/glean/repositories"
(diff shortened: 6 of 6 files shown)


wlach (changes requested):
Don't have time to review this just now, but note that CI is currently failing -- can you try to fix it?

setup.cfg
```diff
@@ -0,0 +1,11 @@
...
+max-line-length=120
+exclude=.git,__pycache__
+
+[isort]
+known_first_party = glean-dictionary
+multi_line_output=3
+include_trailing_comma=True
+force_grid_wrap=0
+combine_as_imports=True
+line_length=120
```
wlach, line 11:
```suggestion
line_length=120

```

.circleci/config.yml
```diff
@@ -11,7 +12,21 @@ jobs:
...
+  python-build-and-test:
+    executor: python/default
+    steps:
+      - checkout
+      - python/install-deps
+      - run: pip install flake8
+      - run: pip install black
+      - run:
+          name: Check format with Black
+          command: black .
```
wlach, line 24: I'm pretty sure this just reformats stuff, which isn't what we want (we want it to emit an error in the case that formatting failed)

You can call black automatically via flake8-black, which would make this less important (this is what mozregression does: https://github.com/mozilla/mozregression). See:

https://pypi.org/project/flake8-black/

.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
+repos:
```
wlach, line 1: I guess this is https://pre-commit.com/ ? This looks interesting, but out of scope for this PR / issue. Let's take this out.

requirements.txt
```diff
@@ -1,2 +1,3 @@
 mozilla-schema-generator
 stringcase
+flake8
```
wlach, line 3: `black` at least (along with `flake8-black`, see above) and `isort` should be added to this list.

pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[tool.black]
+line-length = 88
```
wlach, line 2: We should try to be consistent with the line length setting (I think 100 is a good number, not too long, not too short). I would also configure isort here, instead of in setup.cfg.

You can probably mostly copy the implementation from mozregression here:

https://github.com/mozilla/mozregression/blob/master/pyproject.toml

setup.cfg
```diff
@@ -0,0 +1,11 @@
+[flake8]
+max-line-length=120
+exclude=.git,__pycache__
+
+[isort]
```
wlach, line 5: The settings from mozregression are probably good to use (minus the parts that are specific to glean-dictionary, which seem correct). As mentioned above, define them in pyproject.toml:

https://github.com/mozilla/mozregression/blob/master/pyproject.toml#L12

wlach (changes requested):
Thanks! This is on the right track, see comments for some changes I'd like to see.

Also, for ergonomics, can you add a `Makefile` with targets for running the linter in both check (`lint`) and fix (`lint-fix`) modes?

You can use the Makefile in probe-scraper as an example of how to do this sort of thing (obviously our makefile will be much simpler to start):

https://github.com/mozilla/probe-scraper/blob/master/Makefile

.pre-commit-config.yaml
Riju19: Yes, I added it because I thought we might want to add a pre-commit hook with these checks as well. But yes, you're right that should be part of a separate issue/PR.

Riju19:
Currently, we have a build failure because of the lint checks we have added. Some parts of build-glean-metadata.py that have large URLs cross the max-line-length of 100 that we have set. 
I am not sure if we want to split the URL to go to the next line. The other option would be to increase the max-line-length to 120.

Makefile
```diff
@@ -0,0 +1,15 @@
...
+	@echo "  lint-fix               Format code with black and isort"
+
+lint: build
+	docker-compose run app flake8
+	docker-compose run app python -m black --check .
+	docker-compose run app python -m isort --check-only .
+
+lint-fix: build
+	docker-compose run app python -m black .
+	docker-compose run app python -m isort .
```
wlach, line 14: This clearly won't work, as we're not using `docker-compose`. 😄 

Two things:

1. Test your code by running `make lint` and `make lint-fix`
2. Have the .circleci config above call the makefile.

.circleci/config.yml
```diff
@@ -11,7 +12,18 @@ jobs:
       - node/install-packages
       - run: npm install
       - run: npm test
+  python-build-and-test:
+    executor: python/default
+    steps:
+      - checkout
+      - python/install-deps
+      - run: pip install flake8
+      - run: pip install flake8-black
```
wlach, line 21:
```suggestion
      - run: pip install -r requirements.txt
```

wlach (changes requested):
For the lines that are currently failing validation, you should be able to mark them as one-offs to be ignored by putting `# NOQA` at the end of the line.

Makefile
Riju19: Oops! 
Okay, so I have made some changes and tested them as well. 
Also, I have added another linting library called autoflake, which would remove unused imports. Before I had added this, they were just being flagged by flake8 on running `make lint`. Now, on running `make lint-fix`, that would get fixed as well. 
Let me know your thoughts on this one.

.circleci/config.yml
```diff
@@ -11,7 +12,17 @@ jobs:
       - node/install-packages
       - run: npm install
       - run: npm test
+  python-build-and-test:
+    executor: python/default
+    steps:
+      - checkout
+      - python/install-deps
+      - run: pip install -r requirements.txt
```
wlach, line 20: This is taken care of by the line above

```suggestion
```

wlach (changes requested):
Looks good. Just one last code change requested! Can you also give this PR a descriptive title as outlined in:

https://github.com/mozilla/glean-dictionary/blob/main/CONTRIBUTING.md#opening-pull-requests

Finally, please mark this as ready for review, it's no longer a draft. 😄

.circleci/config.yml
Riju19: Done!
:D I just wanted to be sure I've covered everything that was in scope for this issue

wlach (approved):

wlach:
Thank you!

PR #145, opened 2020-10-20
Avoid formatting tailwind bundle.css file

DawoodShahat:
It took around one and half minute, just to format tailwind bundle.css. 

I think we don't need formatting a minified file.

Here is screen shot.
Screenshot 2020-10-20 at 8 52 17 AM

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,5 @@
 public/build/bundle.js
 public/build/bundle.css
 public/data/*
+public/build/public/build/bundle.css
 venv


.prettierignore
```diff
@@ -1,4 +1,5 @@
 public/build/bundle.js
 public/build/bundle.css
 public/data/*
+public/build/public/build/bundle.css
```
robhudson, line 4: That path shouldn't exist. The `bundle.css` should only be written to `public/build/bundle.css`. I believe this was an error in the earlier app template that was [fixed](5e787003843d662079e8c047078c108de8b6d976).

DawoodShahat: Thanks, I think we need to update the rollup.config.js so that the tailwind minified bundle use a different name. There is already one with the same name `public/build/bundle.css`. 

By the way, the link to the fix isn't working.

robhudson: Updated the link. I guess the github linker to a commit hash thought it was a pull request. Weird.

wlach (changes requested):
@DawoodShahat do you want to take a stab at fixing the rollup configuration instead of this change?

DawoodShahat:
Yes, I think `build/public/build/bundle.css` is redundant and should be removed.

Iinh:
Hi @DawoodShahat,

This file `build/public/build/bundle.css` shouldn't exist. I think the reason why you still have the file locally is because you might have forked the project early on when the rollup config still wrote to `build/public/build/bundle.css`(see [this change](https://github.com/mozilla/glean-dictionary/commit/5e787003843d662079e8c047078c108de8b6d976)) and when you built the app it created the extra files. You might want to delete the `build` folder inside `public` and run `npm run dev` again to see the latest folder update. It should fix your format delay problem.

I don't think it's necessary to change the css file name in rollup config because the Tailwind preprocessor created `bundle.css` in the first place. If you change the name and reload, it would create the same css file, with just a different name.

DawoodShahat:
@Iinh Thanks, for clarifying this.

PR #129, opened 2020-10-19
Handle case where there is no ping or metric data (fixes #81)

DawoodShahat:
This PR solves #81 

Here is what it looks like:
Screenshot 2020-10-19 at 9 28 06 AM

PR #147, opened 2020-10-20
Show when a search doesn't match any apps/metrics/pings (fixes #146)

### LINKED ISSUE #146: Improve filter
In the filter box when we search for something and it doesnt match with any item, just a blank page stares at us. There should a text telling search doesnt match with any application or martics.

mhmohona:
This is a fix for  #146. 

Added a message so when search input doesn't meet with application or matrics name it will show a text like this: 

filter


src/pages/AppDetail.svelte
```diff
@@ -49,7 +50,11 @@
       <td><code>{app.app_id}</code></td>
     </tr>
   </table>
-
+  {#if app.prototype}
+    <AppAlert
+      message="The {params.app} application is a prototype. The metrics and pings listed below may contain inconsistencies and testing strings."
+      bgColor="#808895" />
+  {/if}
```
wlach, line 57: This seems unrelated to this change?

src/pages/AppDetail.svelte
```diff
@@ -67,6 +72,8 @@
         <a href={`/apps/${params.app}/metrics/${metric.name}`}>{metric.name}</a>
         <i>{metric.description}</i>
       </li>
+    {:else}
+      <p>Your search didn't match any metric.</p>
```
wlach, line 76: There's a bit of overlap between this and #129. We should probably account for the possibility that there are no metrics associated with an application (however rare that possibility may be in practice).

src/pages/AppList.svelte
```diff
@@ -36,7 +36,12 @@
         {#if app.deprecated}
           <Pill message="Deprecated" bgColor="#4a5568" />
         {/if}
+        {#if app.prototype}
+          <Pill message="Prototype" bgColor="#808895" />
+        {/if}
```
wlach, line 41: This also looks unrelated.

wlach (changes requested):
Can you give this pull request a more descriptive title? It would also be a good idea to specify which issue it fixes in the title (that way we can link pull requests with issues automatically)

src/pages/AppDetail.svelte
mhmohona: I did not add this. I only added the else part for this PR. 
I am confused why it has come into my edit. :sweat_smile:

mhmohona: The formate changed after I ran `npm run format`. 
Otherwise, I was getting error on CircleCI for not running Prettier.

src/pages/AppDetail.svelte
mhmohona: Yes, you are right, it is showing the error message when there is no metrics associated. 
image

Should I wait till #129 gets marged and after that work on this PR @wlach?

wlach: Yes that makes sense! I expect it will be merged soon, the code looks good just waiting on @DawoodShahat to fix a merge conflict.

wlach: #129 has been merged

wlach (approved):
Thanks!

PR #148, opened 2020-10-20
added no-result text for search filter

LilyLME:
As mentioned #146 a blank page is not appropriate when the search filter matches. Nothing. This PR checks whether there's something in the filter and displays a no result message if found blank.
### Preview
Screenshot from 2020-10-20 15-26-59


wlach:
It looks like you and @mhmohona collided on this. 😄 I'm going to close this one because their PR came before it. Idea looks correct though.

PR #152, opened 2020-10-21
Update Tailwind config to remove PurgeCSS package

Iinh:
- Tailwind [v1.4](https://tailwindcss.com/docs/release-notes#tailwind-css-v1-4) (released April 29th, 2020) and later has built-in PurgeCSS support. We no longer need to install third-party PurgeCSS and set up its configuration. There are [new features](https://tailwindcss.com/docs/release-notes) that we can utilize later by setting up tailwind config 
- What I did:
    - Move purge configuration from postcss to tailwind config
    - Remove @fullhuman/purgeCSS from postcss config and package.json

Looking forward to feedback! Thanks!

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -1,17 +1,7 @@
-const purgecss = require("@fullhuman/postcss-purgecss")({
-  content: ["./src/**/*.html", "./src/**/*.svelte"],
-
-  whitelistPatterns: [/svelte-/],
-
-  defaultExtractor: (content) => content.match(/[A-Za-z0-9-_:/]+/g) || [],
-});
-
 const postCssImport = require("postcss-import");
 
 const tailwindcss = require("tailwindcss");
 
-const production = !process.env.ROLLUP_WATCH;
-
 module.exports = {
-  plugins: [postCssImport, tailwindcss, ...(production ? [purgecss] : [])],
+  plugins: [postCssImport, tailwindcss],
 };
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -1,7 +1,4 @@
-const postCssImport = require("postcss-import");
-
-const tailwindcss = require("tailwindcss");
-
 module.exports = {
-  plugins: [postCssImport, tailwindcss],
+  /* eslint-disable global-require */
+  plugins: [require("postcss-import"), require("tailwindcss")],
 };
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -1,21 +1,5 @@
 const production = !process.env.ROLLUP_WATCH;
 
-const svelteExtractor = (content) => {
-  const regExp = new RegExp(/[A-Za-z0-9-_:/]+/g);
-  const matchedTokens = [];
-  let match = regExp.exec(content);
-
-  while (match) {
-    if (match[0].startsWith("class")) {
-      matchedTokens.push(match[0].substring(6));
-    } else {
-      matchedTokens.push(match[0]);
-    }
-    match = regExp.exec(content);
-  }
-  return matchedTokens;
-};
-
 module.exports = {
   future: {},
   purge: {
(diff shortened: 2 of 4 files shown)


postcss.config.js
```diff
@@ -1,17 +1,7 @@
...
-
 const postCssImport = require("postcss-import");
 
 const tailwindcss = require("tailwindcss");
 
-const production = !process.env.ROLLUP_WATCH;
-
 module.exports = {
-  plugins: [postCssImport, tailwindcss, ...(production ? [purgecss] : [])],
+  plugins: [postCssImport, tailwindcss],
```
robhudson, line 6: To avoid a single-use variable you could instead write:
```
module.exports = {
  plugins: [
    require('postcss-import'),
    require('tailwindcss'),
  ],
}
```

tailwind.config.js
```diff
@@ -0,0 +1,18 @@
+const production = !process.env.ROLLUP_WATCH;
+
+module.exports = {
+  future: {},
+  purge: {
+    content: ["./src/**/*.svelte", "./src/**/*.html"],
+    enabled: production, // disable purge in dev
+    options: {
+      whitelist: [/svelte-/],
```
robhudson, line 9: In [version 3](https://github.com/FullHuman/purgecss/blob/master/CHANGELOG.md) of purgeCSS, `whitelist` is now called `safelist`.

postcss.config.js
robhudson: I've heard mention that purgeCSS won't purge classes when using svelte's [class directive](https://svelte.dev/tutorial/classes). In which case, purgeCSS needs a custom extractor to also find the `class:...` directives.

I've included one in the app-template branch here:
https://github.com/graph-paper-org/app-template/blob/tailwindcss/tailwind.config.js#L1-L15

Just note that the rest of the configuration is out of date, but that extractor may be useful.

tailwind.config.js
robhudson: Extra info: PurgeCSS v3.x is now included in TailwindCSS v1.9 and above. So this will need to be changed for this to work.

robhudson (changes requested):
Single comments were made inline. Great start, and I hope to apply your final changes to the tailwindcss app-template branch. :+1:

robhudson:
Also, are there any other packages in `package.json` that we no longer need if tailwindcss installs them for us?

postcss.config.js
Iinh: > To avoid a single-use variable you could instead write:

@robhudson Eslint doesn't like inline requires, so I added an eslint disable rule in my updated code to accommodate this change.

> I've heard mention that purgeCSS won't purge classes when using svelte's class directive. In which case, purgeCSS needs a custom extractor to also find the class:... directives.

Great insight! I just tested some class styling and Tailwind did not render at all. Thank you for the custom extractor, I'll definitely need it for my future Svelte projects.

Iinh:
Hi @robhudson! Appreciate the feedback -- I've updated my code with your recommendation.
> Great start, and I hope to apply your final changes to the tailwindcss app-template branch. 

Thank you for considering this change for graph-paper template!

> Also, are there any other packages in package.json that we no longer need if tailwindcss installs them for us?

Looks like postcss-purgecss is the only major package that Tailwind added as a built-in feature recently, according to their release notes. I'll research more on this and update with another PR if there's any more we don't need.

tailwind.config.js
```diff
@@ -0,0 +1,34 @@
...
+  while (match) {
+    if (match[0].startsWith("class")) {
+      matchedTokens.push(match[0].substring(6));
+    } else {
+      matchedTokens.push(match[0]);
+    }
+    match = regExp.exec(content);
+  }
+  return matchedTokens;
+};
```
robhudson, line 17: Would you mind testing this regex on this page? If this works it will avoid another loop and may be faster to parse...

https://github.com/sarioglu/svelte-tailwindcss-template/blob/master/tailwind.config.js

Iinh: This extractor looks great, however I feel like `/(?:class:)*([\w\d-/:%.]+)/` is missing checking for underscores and would like to add `_` to the regex. What do you think @robhudson?

robhudson: The `\w` matches any word character (equal to [a-zA-Z0-9_]) so it includes the underscore.

Iinh: You're right, thank you for clarifying! Tested and everything worked. I've updated the code with this new extractor.

robhudson (approved):

PR #155, opened 2020-10-23
Render markdown in description fields (fixes #153)

### LINKED ISSUE #153: Render markdown in description fields
Currently we're now rendering markdown in Glean description fields, which doesn't look great:

image
http://localhost:5000/#!/apps/mozregression

We should render these types of fields with markdown. For the [schema dictionary](https://github.com/mozilla/schema-dictionary) I used the `marked` parser, which seems to work pretty well: https://www.npmjs.com/package/marked

It seems most likely that we'll define a svelte component for rendering this type of information, in which case we should add a story for it in: https://github.com/mozilla/glean-dictionary/tree/main/stories

alvesitalo:
This creates a markdown render component and adds a story.

### Pull Request checklist

      fixes, if applicable)


wlach (approved):
This looks great, I actually don't have any feedback. :) Merging as-is.

src/components/PingDescription.svelte
```diff
@@ -0,0 +1,10 @@
+<script>
+  import marked from "marked";
+
+  export let app;
+  export let name = "";
+  export let description;
+</script>
+
+<a href={`/apps/${app}/pings/${name}`}>{name}</a>
```
wlach, line 9: Should probably not include this by default, since in some cases (on the ping page) we may not show this.

wlach (changes requested):
Actually, sorry, there are some changes I'd request (it's been a long day).  With regard to ping descriptions, we should also use this component on this page: https://github.com/mozilla/glean-dictionary/blob/main/src/pages/PingDetail.svelte

Also, I think we need to apply this to more than just pings. For example, metrics might have markdown in their descriptions as well. See, for example:

http://localhost:5000/#!/apps/fenix/metrics/metrics.search_count

alvesitalo:
Pings and metrics are having markdown in their descriptions now. In addition, the markdown is being rendered inline so pings will no longer have a <p> tag :tada:.

wlach (changes requested):
Code looks good now, thanks!

Thinking about this a bit more, I think we should generalize this to a `Markdown` component and use [slots]((https://svelte.dev/tutorial/slots)), instead of hardcoding a description field . That way you could do something like this just about anywhere:

```svelte
<Markdown>
This my *markdown* text with a [link](https://mozilla.org)
</Markdown>
```

Do you mind making the modifications? After that, I'd be very happy to land.

alvesitalo:
I don't. I think It's a great idea.

alvesitalo:
Done! Markdown component is working, I just had to add a text var to work as a fallback for slot on StoryBook. Let me know if everything is OK.

src/components/Markdown.svelte
```diff
@@ -0,0 +1,13 @@
...
+  let markdown;
+  export let text = "";
+</script>
+
+{#if markdown}
+  {@html marked.parseInline(markdown.textContent)}
+{/if}
+<div class="raw-markdown hidden" bind:this={markdown}>
+  <slot>{text}</slot>
+</div>
```
wlach, line 13: Hmm, it's a bit unfortunate that this adds a hidden div to the document:

image

I offhand don't know of a better way of doing this though, and your solution does work well, so am inclined to merge.

wlach (approved):

wlach:
Awesome work @alvesitalo! Congratulations on getting this over line.

Issue #142, opened 2020-10-19
External link should be opened in new tab


mhmohona:
Opening an `external link` in a new tab allows one to explore the other site as much as they want without having to hit the back button again. It helps to keep focus in one place without losing other websites' information. 

Currently, **metrics page** opens external links in new tab. **Pings page** and **Tables page** need to be updated. 

Here another thing to be noted, we should not open `internal link` in the new tab because it might confuse users. Also, Glean dictionary has the flexibility to navigate back or elsewhere according to users' needs. So for this, keeping users in the same tab helps them understand the navigation flow better.

wlach:
Not sure about this one, it's fairly easy to open links in a new tab by right clicking on things and people might prefer it that way. We should get more feedback from the intended audience of the tool (data practitioners at Mozilla) before going ahead with this. I'm going to leave this open for now, but close the pull request.

mhmohona:
So when I was checking the ping page and matrics page for the first time, for matrics page external links were opening in new tab but for ping page it was not. I got confused at first why same thing is happening in two different ways for two pages. I thought maybe it would be better to make thingswork in same way for all pages. Thats why rised this issue and made PR.

wlach:
Incidentally we can set navigation links to open in a new tab by default via the `<base>` element. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base

mhmohona:
> Incidentally we can set navigation links to open in a new tab by default via the `<base>` element. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base

@wlach, shall I refactor #143?

wlach:
> > Incidentally we can set navigation links to open in a new tab by default via the `<base>` element. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base
> 
> @wlach, shall I refactor #143?

Sorry for the late response on this. I think still no, for now: I'd like to get more feedback from the potential users of this thing first.

mhmohona:
Alright!

PR #169, opened 2020-10-26
Display notification emails on application pages (fixes #151)

### LINKED ISSUE #151: Display notification emails on application pages
We currently display email addresses on the ping page:

Screen Shot 2020-10-21 at 3 57 49 PM
http://localhost:5000/#!/apps/mozregression/pings/usage

The repository metadata also has this information, however, and it would be good to display it on the application page (http://localhost:5000/#!/apps/mozregression).

As part of this implementation, let's create a component for rendering this type of information (maybe `EmailAddresses`?) nicely and create a story for it. One nicety we could add would be creating a `mailto:` URL for the email address, to make it a little easier to send a mail to the relevant party.

Iinh:
This PR fixes #151.
### What I did:
- Create a separate EmailAddress component to be used in Application and Ping pages. 
- Put the loop logic in the component since emails are always returned in an array
- Add `mailto:` link 
- Create stories for EmailAddress

 #### Pull Request checklist

      fixes, if applicable)


src/components/EmailAddress.svelte
```diff
@@ -0,0 +1,20 @@
...
+<style>
+  td {
+    @apply border;
+    @apply p-2;
+  }
+</style>
+
+<tr>
+  <td>Notification Email{emails.length > 1 ? 's' : ''}</td>
+  <td>
```
wlach, line 14: I think we should not include the table elements here-- we should focus the component purely on displaying email addresses (we may want to use different elements to organize stuff in different places, but still use this component).

src/components/EmailAddress.svelte
```diff
@@ -0,0 +1,20 @@
...
+    {#each emails as email}
+      <a href="mailto:{email}" target="_blank">{email}</a>
+      <br />
```
wlach, line 17: Instead of using `<br/>` here (which is a bit hacky), I'd put them in a span where class="block" (see original code)

wlach (changes requested):
Great stuff, a few comments. Sorry for the delay on this review.

wlach (changes requested):
Good stuff! In addition to the item pointed in the review, I think we should call this component `EmailAddresses`

src/components/EmailAddresses.svelte
```diff
@@ -0,0 +1,8 @@
+<script>
+  export let emails;
+</script>
+
+{#each emails as email}
+  <span class="block">
+    <a href="mailto:{email}" target="_blank">{email}</a></span>
```
wlach, line 7: Any reason this was set to target=blank_? I don't think it should be necessary (suspecting it got added by mistake see also discussion in #142)

```suggestion
    <a href="mailto:{email}">{email}</a></span>
```

wlach (changes requested):
Just one last question

src/components/EmailAddresses.svelte
Iinh: I thought it would be nice for the users to open their browser-based email app in a new tab without having to leave the current Glean page. However I read discussion #142, I understand the reasoning for keeping default behavior for links for now. Suggested change committed.

wlach (approved):
Great stuff, thank you!

PR #170, opened 2020-10-27
Filter text gets populated with the search query in the URL (fixes #106)

### LINKED ISSUE #106: Persist search state in BigQuery table view
In the BigQuery table view we currently allow the user to search through the column names to find the ones of interest:

image
http://localhost:5000/#!/apps/mozphab/tables/usage

It would be very handy if we could *persist* that search in the URL (and restore it when they visit it), so that people could link to specific views and have the search prepopulated. In the above example, that would be:

http://localhost:5000/#!/apps/mozphab/tables/usage?search=build

To perform this task, have a look at the documentation for page.js, which is what we currently use for routing: https://visionmedia.github.io/page.js/

You will probably need to modify the main router (`App.svelte`) in addition to the component for the table view.

Riju19:
### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -14,6 +14,7 @@
   let component;
   let params;
   let links = [];
+  let queryString;
 
   afterUpdate(() => {
     const { app, ping, metric, table } = params;
@@ -51,6 +52,18 @@
     };
   }
 
+  function parseQuery(ctx, next) {
+    queryString = ctx.querystring;
+    queryString = queryString !== "" ? queryString.split("=")[1] : "";
+    next();
+  }
+
+  function updateURL({ detail: searchQuery }) {
+    let hash = window.location.hash.split("?")[0];
+    page(`${hash}?search=${searchQuery}`);
+  }
+
+  page("*", parseQuery);
   page("/", setComponent(AppList));
   page("/apps/:app/tables/:table", setComponent(TableDetail));
   page("/apps/:app/pings/:ping", setComponent(PingDetail));
@@ -77,5 +90,9 @@
 <Breadcrumb {links} />
 
 <div class="container py-4 mx-auto">
-  <svelte:component this={component} bind:params />
+  <svelte:component
+    this={component}
+    bind:params
+    bind:queryString
+    on:updateURL={updateURL} />
 </div>
--- a/src/components/SchemaViewer.svelte
+++ b/src/components/SchemaViewer.svelte
@@ -1,11 +1,22 @@
 <script>
+  import { createEventDispatcher } from "svelte";
   import SchemaNode from "./SchemaNode.svelte";
   import FilterInput from "./FilterInput.svelte";
 
   export let app;
   export let nodes = [];
+  export let searchText;
   let nodesWithVisibility;
-  const filterTextChanged = (filterText = "") => {
+
+  const dispatch = createEventDispatcher();
+
+  function updateSearchText(filterText) {
+    searchText = filterText;
+    dispatch("updateURL");
+  }
+
+  const filterTextChanged = (filterText = searchText || "") => {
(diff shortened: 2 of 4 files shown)


Riju19:
These changes do not work yet, but I just wanted to get some feedback on if I'm heading in the right direction with this. 
Currently, I am just trying to do this one-way, that is, take in the search query passed in the URL and update the filter text component with that.

wlach (commented):
> These changes do not work yet, but I just wanted to get some feedback on if I'm heading in the right direction with this.
> Currently, I am just trying to do this one-way, that is, take in the search query passed in the URL and update the filter text component with that.

@Riju19 yup this looks like it's on roughly the right track!

src/components/SchemaViewer.svelte
```diff
@@ -1,11 +1,20 @@
...
 
   export let app;
   export let nodes = [];
+  export let queryString;
   let nodesWithVisibility;
-  const filterTextChanged = (filterText = "") => {
+
+  function updateQueryString(filterText) {
+    queryString = filterText;
+    page(`${window.location.hash.split("?")[0]}?search=${queryString}`);
```
Riju19, line 13: This works, but doesn't seem quite right to me. Any suggestions on how this could be done in a better way?

wlach: Hmm, I think we should probably define a callback function inside the components using this, and delegate the responsibility of updating the URL to them. Calling into page.js from a sub-component seems like a pretty big violation of seperation-of-concerns.

After that, we should be able to pass this page-level data to the component via page.js's callback facility, which is a little cleaner than what you're doing here. However, that's a little more involved, I think for a first pass what you're doing is fine-- we can do that fairly easily in a followup later.

src/components/SchemaViewer.svelte
```diff
@@ -1,11 +1,20 @@
 <script>
+  import page from "page";
   import SchemaNode from "./SchemaNode.svelte";
   import FilterInput from "./FilterInput.svelte";
 
   export let app;
   export let nodes = [];
+  export let queryString;
```
wlach, line 8: If you implement my suggestion below, I think we can remove any concept of "query strings" from this component, and just focus on an interface to setting/handling changes to a searchText parameter.

```suggestion
  export let searchText;
```

wlach (changes requested):
Hi @Riju19, this is solid work. I have some suggestions on the approach, but this is definitely on the right track. Thanks for digging into it!

src/App.svelte
```diff
@@ -51,8 +52,19 @@
     };
   }
 
+  function setQueryString(ctx, next) {
```
wlach, line 55: I'm pretty sure we should be able to avoid managing a global queryString object by just setting it on the context via a handler, then handling it in the `setComponent` method. This has the added bonus of making it possible for us to use the query string in other pages in the future (something we will almost certainly want to do).

Search for `query-string` in https://visionmedia.github.io/page.js/ for some code examples which should help you out here.

wlach (changes requested):
Awesome! I tested it locally and it seems to work great. Just have one last piece of feedback to address, then I think we can land.

src/App.svelte
```diff
@@ -51,6 +52,18 @@
     };
   }
 
+  function parseQuery(ctx, next) {
+    queryString = ctx.querystring;
+    queryString = queryString !== "" ? queryString.split("=")[1] : "";
```
wlach, line 57: We eventually should make this more flexible (i.e. break it out into an object of keys and values), but this is good enough for now.

wlach (approved):
Thanks @Riju19 -- I just did some final testing and this is working great. Thanks so much for your persistence on this.

Issue #190, opened 2020-11-11
Descriptions of metrics on app page seem incorrect


wlach:
For example on https://glean-dictionary-dev.netlify.app/#!/apps/fenix we see this on the second page:

<img width="907" alt="image" src="https://user-images.githubusercontent.com/20569/98827437-e081bc80-2404-11eb-8af3-39831b9f4aaa.png">

This is clearly incorrect, which you can see if you navigate to the metric detail:

https://glean-dictionary-dev.netlify.app/#!/apps/fenix/metrics/browser.search.ad_clicks

It's unclear to me right now whether this is a bug in the ETL or the display, but it's hopefully not too hard to fix.

Iinh:
@wlach Could I work on this issue?

wlach:
> @wlach Could I work on this issue?

For sure! Please go ahead.

mhmohona:
In app page, this is happening for some other metric disruptions as well, not only for browser.search.ad_clicks.

PR #192, opened 2020-11-11
Fix incorrect metric description caused by markdown rendering (#190)

Iinh:
**Issue #190** 

The bug is caused by the Markdown component currently producing a hidden div that does not update the text data properly after pagination. 

Screen Shot 2020-11-11 at 2 10 35 PM

Specifically, [this code](https://github.com/mozilla/glean-dictionary/blob/4580911e58a23896bb832f9f98ec2740f6954bf0/src/components/Markdown.svelte#L8-L10) will cause the initial rendering of page 1 metric data to persist on the page, even after the page number is changed. To fix this, I removed the hidden div logic and just rendered the text directly. 

@wlach 

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -1,13 +1,9 @@
 <script>
   import marked from "marked";
 
-  let markdown;
-  export let text = "";
+  export let text;
 </script>
 
-{#if markdown}
-  {@html marked.parseInline(markdown.textContent)}
-{/if}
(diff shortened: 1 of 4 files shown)


wlach (approved):
This looks like the right change, thanks @Iinh -- I think I incorrectly insisted on the previous approach when originally reviewing @alvesitalo's PR. I think passing the markdown text in as a property is the right thing here.

PR #194, opened 2020-11-14
Add Snapshot testing 

### LINKED ISSUE #117: Run storybook tests in ci
Storybook snapshot tests in CI (where we verify that stories still render after a pull request) can often catch problems, e.g. it was helpful in the iodide project (see iodide-project/iodide#2506). I'm not sure how to set this up in Svelte but I'm guessing it should be possible. You can see the aforementioned PR for iodide for some ideas.

DawoodShahat:
This PR fixes #117 

### Pull Request checklist

      fixes, if applicable)


src/components/FilterInput.svelte
```diff
@@ -1,5 +1,5 @@
 <script>
-  import throttle from "just-throttle";
+  const throttle = require("just-throttle");
```
wlach, line 2: I think we should probably not make these changes, as the `import` syntax is more intuitive to people. I gather you're seeing some kind of error here? Can you post a trace and/or Google the results? I'm guessing that we probably just need to add a pre/post processing step somewhere in the build configuration.

wlach (changes requested):
This looks great @DawoodShahat! I have some concerns about modifying the svelte files, but other than that this is looking very close.

Checking the CI logs, I noticed this warning:


```
    console.warn
      <FilterInput> was created without expected prop 'filterText'



      at new FilterInput (src/components/FilterInput.svelte:132:12)
      at create_fragment (stories/FilterableList.svelte:56:16)
      at init (node_modules/svelte/internal/index.js:1465:37)
      at new FilterableList (stories/FilterableList.svelte:193:3)
      at getRenderedTree (node_modules/@storybook/addon-storyshots/dist/frameworks/svelte/renderTree.js:20:5)
      at node_modules/@storybook/addon-storyshots/dist/test-bodies.js:11:22
      at Object.<anonymous> (node_modules/@storybook/addon-storyshots/dist/api/snapshotsTestsTemplate.js:42:20)
```

It looks like maybe the story is lacking something? It would be good if we could make issues like that fail the CI, as they often indicate bugs.

src/components/FilterInput.svelte
DawoodShahat: You are right. Currently I'm getting an error similar to this issue [https://github.com/facebook/jest/issues/6229](https://github.com/facebook/jest/issues/6229). 

- This Issue arises when an external library is used e.g. `marked` or `just-throttle` that is not transpiled to make ES6 imports work with nodejs. 

- `babel-jest` ignores node_modules altogether. I tried different plugins to make it work e.g.  `@babel/plugin-transform-modules-commonjs`. I also tried different solutions provided in the aforementioned issue, but still facing the same issue.

- Here is what I get when I try to use `marked`
Screenshot 2020-11-18 at 10 27 21 PM

wlach: Ah ok. Went down a bit of a rabbit hole with this. After looking at things a bit more, I think this has to do with the particulars of how these two libraries work as you describe. In the case of `just-throttle`, the solution is probably just to use [lodash's throttle function](https://lodash.com/docs/4.17.15#throttle) instead (this will require depending on lodash, which we want to do anyway IMO). In the case of marked, it's actually just a minor change we need to make in how it's imported. I'll suggest it.

wlach: Oops! I spoke too soon. We actually *can't* use require, as it's not recognized by client-side JavaScript:

image

Do you want to just go ahead and add lodash to the repo as part of this PR? Just run `npm install lodash` and you should be good to go.

wlach, line 2: With lodash we can just do:

```suggestion
  import { throttle } from "lodash";
```

... and it'll work

src/components/Markdown.svelte
```diff
@@ -1,5 +1,5 @@
 <script>
-  import marked from "marked";
+  const marked = require("marked");
```
wlach, line 2:
```suggestion
  import { parseInline } from "marked";
```

src/components/FilterInput.svelte
DawoodShahat: Yes, I will make the suggested changes in the next commit. Thanks.

DawoodShahat:
Here is what I updated upon @wlach review:

- Removed `just-throttle` dependency and instead used `_throttle` from `lodash`
- Fixed the warning that was generated by `FilterInput.svelte` component
- Used ES6 imports instead of CJS require

The CI will now fail when there is an unintended change in the markup. I think we should update the README and add some instructions on how to update the generated snapshots with your intended changes e.g. `npm run test:jest -- -u`.

Also one more thing, What do you think about deploying Stories to netlify? 

By the way thanks for the review, I spent a good amount of time on the error.

wlach:
> The CI will now fail when there is an unintended change in the markup. I think we should update the README and add some instructions on how to update the generated snapshots with your intended changes e.g. `npm run test:jest -- -u`.

Great idea, could you add something about this to the README? I think we can add them to the storybook: https://github.com/mozilla/glean-dictionary#storybook

> Also one more thing, What do you think about deploying Stories to netlify?

This is also a great idea! I think we could probably publish a static copy to `/storybook/`. I would obviously do this in a followup.

wlach (changes requested):
Code looks great, I think we just need an update to the README and we'll be good to land.

README.md
```diff
@@ -49,6 +49,21 @@ run:
...
+### Storybook Snapshot Testing
+
+To run snapshot tests, run:
+
+```bash
+npm run test:jest
+```
+
+If intended change is made to the markup, re-generate snapshots using the
+following command:
```
wlach, line 61:
```suggestion
If you intentionally made a change to a component that results in a change 
to the output of the storybook snapshots, you can re-generate them using 
the following command:
```

README.md
```diff
@@ -49,6 +49,21 @@ run:
 npm run storybook
 ```
 
+### Storybook Snapshot Testing
+
+To run snapshot tests, run:
```
wlach, line 54:
```suggestion
To give us more confidence that changes don't unintentionally break the UI, we run [storybook snapshot tests](https://storybook.js.org/docs/react/workflows/snapshot-testing).

You can run them manually as follows:
```

wlach (approved):
Great stuff, just have some minor suggested changes to the README which I'll apply before landing.

PR #202, opened 2020-11-17
Separate Pings and Metrics in a tab layout

Riju19:
### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/Tabs.svelte
+++ b/src/components/Tabs.svelte
@@ -1,0 +1,47 @@
+<script>
+  import { createEventDispatcher } from "svelte";
+
+  export let tabItems;
+  export let activeTab;
+
+  const dispatch = createEventDispatcher();
+</script>
+
+<style>
+  ul {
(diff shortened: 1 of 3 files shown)


wlach:
Thanks for this @Riju19! I've run out of time to look at this today but will try to provide some feedback ASAP (first thing tomorrow my time)

Riju19:
@wlach No problem! I actually came up with another alternative to this implementation, which can be found here - https://github.com/mozilla/glean-dictionary/pull/203. That one is more like that repl example you had shared, in terms of grouping tabs and their content into a more component-driven design, and I think I prefer that over the one in this pull request.  
But, let me know which one of these would you prefer, or we could also go back and do it exactly the way it was in that repl example, I'm good with that too! :-)

wlach:
I agree, I think #203 is looking a bit better. Let's continue work there. 😄

PR #238, opened 2020-12-02
Parse out all application metadata (fixes #109)

### LINKED ISSUE #109: We should gather all metric/ping types, not just those specified by the applications

wlach:
Previously, pings and metrics which were defined as application
dependencies (e.g. the `glean.error.invalid_label` metric)

Part of this work has been a factoring out of the appropriate
logic from mozilla-schema-generator, in a way that can be hopefully
be reused across projects like this (perhaps living alongside
the code in probe-scraper, see mozilla/mozilla-schema-generator#167)

@acmiyaguchi can you take a look at this? I'd also like your thoughts on whether we could consider moving the code
in glean.py to probe-scraper (and perhaps publishing it as a library that can be used across several packages, including mozilla-schema-generator)

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/scripts/glean.py
+++ b/scripts/glean.py
@@ -1,0 +1,221 @@
+from __future__ import annotations
+
+import logging
+from datetime import datetime
+from typing import List
+
+import cachecontrol
+import requests
+
+logger = logging.getLogger(__name__)
+
+
+class _Cache:
+    """
+    Simple cache manager so we can avoid refetching the same dependency data
+    over and over
+    """
+
+    def __init__(self):
+        self.sess = requests.Session()
+        self.sess.mount("http://", cachecontrol.CacheControlAdapter())
+        self.sess.mount("https://", cachecontrol.CacheControlAdapter())
+
+    def get(self, url: str):
+        return self.sess.get(url)
+
+    def get_json(self, url: str):
+        return self.get(url).json()
+
+
+_cache = _Cache()
+
+
+class GleanObject(object):
+    NAME_KEY = "name"
+    HISTORY_KEY = "history"
+
+
+class GleanMetric(GleanObject):
+    """
+    Represents an individual Glean metric, as defined by probe scraper
+    """
+
+    ALL_PINGS_KEYWORDS = ("all-pings", "all_pings")
+
+    def __init__(self, identifier: str, definition: dict, *, ping_names: List[str] = None):
+        self.identifier = identifier
+        self._set_dates(definition)
+        self._set_definition(definition)
+        self._set_description(self.definition)
+
+        defn_pings = set(
+            [p for d in definition[self.HISTORY_KEY] for p in d.get("send_in_pings", ["metrics"])]
+        )
+        self.definition["send_in_pings"] = defn_pings
+
+        if ping_names is not None:
+            self._update_all_pings(ping_names)
+
+    def _update_all_pings(self, pings: List[str]):
+        if any([kw in self.definition["send_in_pings"] for kw in self.ALL_PINGS_KEYWORDS]):
+            self.definition["send_in_pings"] = set(pings)
+
+    def _set_definition(self, full_defn: dict):
+        self.definition_history = list(
+            sorted(
+                full_defn[self.HISTORY_KEY],
+                key=lambda x: datetime.fromisoformat(x["dates"]["last"]),
+                reverse=True,
+            )
+        )
+
+        # The canonical definition for up-to-date schemas
+        self.definition = self.definition_history[0]
+        self.definition["name"] = full_defn[self.NAME_KEY]
+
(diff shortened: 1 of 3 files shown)


scripts/glean.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+class GleanMetric(GleanObject):
+    """
+    Represents an individual Glean metric, as defined by probe scraper
+    """
+
+    ALL_PINGS_KEYWORDS = ("all-pings", "all_pings")
+
+    def __init__(self, identifier: str, definition: dict, *, ping_names: List[str] = None):
```
acmiyaguchi, line 46: What does the standalone `*` do here?

scripts/glean.py
```diff
@@ -0,0 +1,221 @@
...
+        for dependency in self.get_dependencies():
+            dependency_metrics = _cache.get_json(GleanApp.METRICS_URL_TEMPLATE.format(dependency))
+            metrics += list(dependency_metrics.items())
+
+        ping_names = set(self._get_ping_data().keys())
+
+        processed = []
+        for _id, defn in metrics:
+            metric = GleanMetric(_id, defn, ping_names=ping_names)
+            processed.append(metric)
```
acmiyaguchi, line 206: I want to note https://github.com/mozilla/mozilla-schema-generator/issues/118 where some custom logic was added for schema compatibility where a field was accidentally added with the wrong type. This is specific to schema generation in mozilla-pipeline-schemas where removing a column is not done, it shouldn't matter outside of the module.

acmiyaguchi (changes requested):
I'd suggest factoring out the tests alongside the main module code if the intent is to factor this out from mozilla-schema-generator. You can probably plop in a `scripts/test_glean.py` file with a pytest dependency without too much issue. I think it's fine to bring this closer to the source of truth, as long as we can verify we are creating the same schemas in msg without incompatibilities (which we can verify with our current CI).

Do you also have a diff of the data files? It's be a good sanity check to make sure that the outputs are the same, else document the differences in the PR/commits somewhere. One way to do this for each branch (`main` vs `get-all-the-data`), dump the generated assets into an untracked folder, then run a recursive diff between the two folders.

scripts/glean.py
wlach: Nothing useful. :) I think this is a holdover from what I copied over from mozilla-schema-generator, I'll take it out.

scripts/glean.py
wlach: Right, I deleted that part of the code but the structure here still reflects it. I'll update the code. If/when we use this code in mozilla-schema-generator, we can apply the transform on the output of this function.

wlach:
> I'd suggest factoring out the tests alongside the main module code if the intent is to factor this out from mozilla-schema-generator. You can probably plop in a `scripts/test_glean.py` file with a pytest dependency without too much issue. I think it's fine to bring this closer to the source of truth, as long as we can verify we are creating the same schemas in msg without incompatibilities (which we can verify with our current CI).

I'd rather hold off on adding / doing more here test-wise, as I'm in a bit of a rush to get this done before the end of the half, and I think any test infrastructure I'd add would just need to be ported over to probe-scraper. Also, the current tests in msg seem pretty minimal and not really so useful here:

https://github.com/mozilla/mozilla-schema-generator/blob/master/tests/test_glean.py

> Do you also have a diff of the data files? It's be a good sanity check to make sure that the outputs are the same, else document the differences in the PR/commits somewhere. One way to do this for each branch (`main` vs `get-all-the-data`), dump the generated assets into an untracked folder, then run a recursive diff between the two folders.

Went ahead and did this:

https://gist.github.com/wlach/0fb52a8b3048ebdbea95fcea3fa3b61e

Aside from adding a bunch of stuff that wasn't there previously, it seems like some information which was incorrect/out of date is now correct.

acmiyaguchi (approved):
Thanks for the diff, I skimmed over it and it seems to look as expected. I'm guessing it's a bit large since it also includes updates to the underlying data in addition to the expected changes here.

PR #276, opened 2020-12-15
Consolidate data display between application and ping pages (fixes #225)

### LINKED ISSUE #225: Consolidate metrics display between application and ping pages
Application and ping pages show metrics in more or less the same way, but the application page has some nice features (e.g. pagination by @LilyLME, filtering by @DawoodShahat) that the ping page lacks. Compare:

https://glean-dictionary-dev.netlify.app/#!/apps/fenix

vs. 

https://glean-dictionary-dev.netlify.app/#!/apps/fenix/pings/metrics

We should extract the logic that displays the pings from the application view to a separate svelte component that can also be used inside the ping view. As always when adding a new svelte component, we should create a story for it.

Iinh:
Fixes #225.

## Pull Request checklist

      fixes, if applicable)


stories/displaydata.stories.js
```diff
@@ -0,0 +1,102 @@
+import DisplayData from "../src/components/DisplayData.svelte";
+
+const testData = {
+  name: "Test data",
+  metrics: [
+    {
```
Iinh, line 6: The reason why I had 21 metrics in this test data is the pagination bar only appears when there's more than 20 metrics.

wlach (changes requested):
Great stuff, thank you @Iinh -- this will be very helpful as we start styling and making other functional improvements to the metadata views. I have some small bits of feedback which will hopefully be easy to address.

src/components/DisplayData.svelte
```diff
@@ -0,0 +1,62 @@
...
+  import FilterInput from "./FilterInput.svelte";
+  import Markdown from "./Markdown.svelte";
+
+  export let paginationState = {
+    pages: [],
+    currentPage: 1,
+  };
+
+  export let appName;
+  export let data;
```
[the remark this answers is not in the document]
Iinh: If I passed `data[itemType]` directly as `items` to the component, then at the child component level we would not have access the app name, which is something that we need in the item link

src/components/ItemList.svelte
```diff
@@ -0,0 +1,59 @@
+<script>
+  import { _ } from "lodash";
```
Iinh, line 2: @wlach Using named import here to bypass the snapshot test, but the story would not render. Only default import seems to work in storybook, but if I use default import the test fails. Not sure what to do here

wlach: Changing this to `import { capitalize } from "lodash"` seems to fix it!

src/components/DisplayData.svelte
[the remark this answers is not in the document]
wlach: Is the appName not defined in the line above?

Iinh:
Thanks @wlach for your review! Looks like I have addressed all requested changes.

wlach:
@Iinh I just gave you write access to this repository, do you want to try merging? Please squash this PR into a single commit with a single title referencing the issue. 😄

wlach (approved):

Iinh:
Squashed and merged 🚀

PR #282, opened 2020-12-17
Fix pagination bar (fixes #246 and closes #278)

### LINKED ISSUE #246: Pagination doesn't work well with a large number of metrics
Apps like fenix have a very large number of metrics:

image

We probably shouldn't show all the pages as links in this case. Instead, let's use the approach [defined here](https://ux.stackexchange.com/a/4129) and show something like this:

image

### LINKED ISSUE #278: Use lodash's chunk feature to simplify pagination logic
As suggested in: https://github.com/mozilla/glean-dictionary/pull/119#discussion_r508596663

Note that we now have lodash included in the repository since #194, so you don't even need to add the library. Just use it and go. :)

Iinh:
Fixes pagination bar issue #246, and also closes #278 by using lodash to simplify pagination logic.

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/Pagination.svelte
+++ b/src/components/Pagination.svelte
@@ -93,14 +93,48 @@
 
 <div class="flex flex-col items-center my-12">
   <div class="flex text-gray-700">
-    <div class="flex h-12 font-medium bg-gray-200">
+    <div
+      on:click|preventDefault={() => changePage(currentPage !== 1 ? currentPage - 1 : 1)}
+      class="h-12 w-12 mr-1 flex justify-center items-center rounded-full bg-gray-200 cursor-pointer {currentPage === 1 ? 'bg-teal-600 text-white' : ''}">
+      <svg
+        xmlns="http://www.w3.org/2000/svg"
+        width="100%"
+        height="100%"
+        fill="none"
+        viewBox="0 0 24 24"
+        stroke="currentColor"
+        stroke-width="2"
+        stroke-linecap="round"
+        stroke-linejoin="round"
+        class="feather feather-chevron-left w-6 h-6">
+        <polyline points="15 18 9 12 15 6" />
+      </svg>
+    </div>
+    <div class="flex h-12 font-medium rounded-full bg-gray-200">
       {#each truncatedPagination(currentPage, lastPage) as page}
         <div
           on:click|preventDefault={() => changePage(page)}
-          class="m-4 md:flex justify-center items-center cursor-pointer leading-5 {page === currentPage ? 'bg-teal-600 text-white' : ''}">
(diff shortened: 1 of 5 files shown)


src/components/Pagination.svelte
```diff
@@ -84,14 +93,48 @@
 
 <div class="flex flex-col items-center my-12">
   <div class="flex text-gray-700">
-    <div class="flex h-12 font-medium bg-gray-200">
-      {#each range(lastPage, 1) as page}
+    <div
```
Iinh, line 96: From [this comment](https://github.com/mozilla/glean-dictionary/pull/119#discussion_r508611891) I understand you didn't want the transition buttons in the original pagination implementation. I think having the buttons would be really helpful, especially for users with smaller screens, considering right now we don't have any pagination available at a small screen size.
transition buttons

wlach: Yup, sounds good-- I suggested taking them out of the initial implementation because it was already quite complex, but I think they make a nice addition here.

src/components/Pagination.svelte
```diff
@@ -84,14 +93,48 @@
...
+        height="100%"
+        fill="none"
+        viewBox="0 0 24 24"
+        stroke="currentColor"
+        stroke-width="2"
+        stroke-linecap="round"
+        stroke-linejoin="round"
+        class="feather feather-chevron-left w-6 h-6">
+        <polyline points="15 18 9 12 15 6" />
+      </svg>
```
wlach, line 111: Could you put this in a seperate svelte file under an `./icons/` directory and import it? This will make it easier to reuse later (and also make this page easier to understand).

src/components/Pagination.svelte
```diff
@@ -84,14 +93,48 @@
...
+        height="100%"
+        fill="none"
+        viewBox="0 0 24 24"
+        stroke="currentColor"
+        stroke-width="2"
+        stroke-linecap="round"
+        stroke-linejoin="round"
+        class="feather feather-chevron-right w-6 h-6">
+        <polyline points="9 18 15 12 9 6" />
+      </svg>
```
wlach, line 137: Same as above

wlach (changes requested):
This looks great! Thank you @Iinh.

The only remaining problem is that the content jumps around a little bit as you scroll through. I'll file a new issue about that.

Iinh:
@wlach Code's updated with the requested change.

> The only remaining problem is that the content jumps around a little bit as you scroll through. I'll file a new issue about that.

Yeah I noticed that too. I'm guessing we could fix it with absolute positioning the tab components, or using css grid? I'll be happy to work on that issue.

wlach (approved):
Thanks @Iinh!

PR #283, opened 2020-12-17
Add BigQuery URL to Metric page (fixes #280)

### LINKED ISSUE #280: Link to BigQuery tables from metrics
The only way to get to the bigquery table for a metric is via the ping view:

image

https://glean-dictionary-dev.netlify.app/apps/fenix/pings/metrics/

We should *also* link out to this information from the page for the metric. For example:

https://glean-dictionary-dev.netlify.app/apps/fenix/metrics/addons.has_enabled_addons

Should link to:

https://glean-dictionary-dev.netlify.app/apps/fenix/tables/metrics?search=addons_has_enabled_addons

I would do the linking via a `BigQuery` column. Note that in some cases metrics may appear in multiple pings, which means they'll appear in multiple tables.

Iinh:
This PR fixes #280.

@wlach Could you review if this approach is the one you're thinking of? Thanks!

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -45,7 +45,7 @@ for app in apps:
 
     app_id = app.app_id
     app_id_snakecase = stringcase.snakecase(app_id)
-    
+
     # metrics data
     metrics = app.get_metrics()
     metric_pings = dict(data=[])
@@ -68,13 +68,15 @@ for app in apps:
 
         stable_ping_table_names = []
         for ping in metric.definition["send_in_pings"]:
-            stable_ping_table_names.append([ping, f"{app_id_snakecase}.{stringcase.snakecase(ping)}"])
-        
+            stable_ping_table_names.append(
+                [ping, f"{app_id_snakecase}.{stringcase.snakecase(ping)}"]
+            )
+
         metric_type = metric.definition["type"]
         metric_name_snakecase = stringcase.snakecase(metric.identifier)
         bigquery_names = dict(
-            stable_ping_table_names = stable_ping_table_names,
-            metric_table_name = f"metrics.{metric_type}.{metric_name_snakecase}"
+            stable_ping_table_names=stable_ping_table_names,
+            metric_table_name=f"metrics.{metric_type}.{metric_name_snakecase}",
         )
 
         open(os.path.join(app_metrics_dir, f"{metric.identifier}.json"), "w").write(
@@ -83,7 +85,7 @@ for app in apps:
                     metric.definition,
                     name=metric.identifier,
                     history=metric.definition_history,
-                    bigquery_names=bigquery_names
+                    bigquery_names=bigquery_names,
                 ),
                 default=_serialize_sets,
             )
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -43,6 +43,9 @@ for app in apps:
 
     app_data = dict(app.repo, pings=[], metrics=[])
 
+    app_id = app.app_id
+    app_id_snakecase = stringcase.snakecase(app_id)
+    
     # metrics data
     metrics = app.get_metrics()
     metric_pings = dict(data=[])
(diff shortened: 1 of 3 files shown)


wlach:
Hi @Iinh, great start! A few things:

1. You can't directly lookup the columns in events tables, so we should skip the search string for those cases (e.g. https://deploy-preview-283--glean-dictionary-dev.netlify.app/apps/fenix/metrics/about_page-libraries_tapped)
2. I would give the fully-qualified name in BigQuery (e.g. for https://deploy-preview-283--glean-dictionary-dev.netlify.app/apps/fenix/metrics/geckoview-page_load_time should be `metrics.timing_distribution.geckoview_page_load_time`). 
3. I would also provide the stable name of the table, not just the shortened version. E.g. for fenix metrics it is `org_mozilla_fenix.metrics`

(2) and (3) may require augmenting the python etl a bit, but hopefully it won't be too bad! There may be a few other things I'm missing, but I think that's most of it.

For your reference, I got some of these requirements from looking at the probe dictionary:

image

https://probes.telemetry.mozilla.org/?view=detail&probeId=histogram%2FA11Y_CONSUMERS

wlach (changes requested):
See comments above.

src/pages/MetricDetail.svelte
```diff
@@ -176,6 +177,21 @@
       <td>Version</td>
       <td>{metric.version || 0}</td>
     </tr>
+    <tr>
+      <td>BigQuery</td>
+      <td>
+        {#each metric.bigquery_names.stable_ping_table_names as name}
+          {#if name[0] !== 'events'}
+            <a
```
wlach, line 185: I think it would be nice to have a URL like:

```html
In <a href="/apps/fenix/tables/activation">org_mozilla_fenix.activation</a> as <a href="/apps/fenix/tables/activation?search=metrics.string.activation_identifier">metrics.string.activation_identifier</a>
```

That way people can choose either to browse the full schema, or start off with a specific search. They can also just copy the BigQuery column name directly from the ping page, if that's what they want to do.

wlach (changes requested):
@Iinh This is looking great. The commit history looks a little wonky here, I think you probably want to rebase on main. Try something like:

```bash
git checkout main
git pull origin main
git checkout add-bigquery-link-to-metrics
git rebase main
```

Iinh:
Thanks @wlach for your help with the commit history! I actually run into this problem quite often (rebasing off the wrong branch) so the help was greatly appreciated.

Iinh:
@wlach Could you review this when you're available? Thanks!

src/pages/MetricDetail.svelte
```diff
@@ -176,6 +177,24 @@
...
       <td>{metric.version || 0}</td>
     </tr>
+    <tr>
+      <td>BigQuery</td>
+      <td>
+        {#each metric.bigquery_names.stable_ping_table_names as name}
+          <div>
+            In
+            <a href={getMetricBigQueryURL(params.app, name[0])}>{name[1]}</a>
+            {#if name[0] !== 'events'}
```
wlach, line 187: Just realized this breaks down in this case, where an event metric was (at one point) sent in the metrics ping:

https://deploy-preview-283--glean-dictionary-dev.netlify.app/apps/fenix/metrics/downloads_misc-download_added

We probably need to look at the `type` of the event when making this decision, rather than the name.

scripts/build-glean-metadata.py
```diff
@@ -62,12 +65,27 @@ def _serialize_sets(obj):
...
+        bigquery_names = dict(
+            stable_ping_table_names=stable_ping_table_names,
+            metric_table_name=f"metrics.{metric_type}.{metric_name_snakecase}",
+        )
+
         open(os.path.join(app_metrics_dir, f"{metric.identifier}.json"), "w").write(
             json.dumps(
                 dict(
                     metric.definition,
                     name=metric.identifier,
```
wlach, line 86: Should probably serialize the `metric_type` here, so we can easily access it later (see below)

wlach (changes requested):
Looking great! Just one last edge case that I noticed

wlach (approved):
Awesome, thank you!

PR #296, opened 2020-12-24
Keep metrics list in a fixed height container (fixes #285)

### LINKED ISSUE #285: Metrics list should be fixed in height
Currently the metrics list jumps around a bunch as you paginate through results:

image

This is because the height is not fixed. We should probably make it a fixed-height container, like we do with the schema browser:

image
https://glean-dictionary-dev.netlify.app/apps/fenix/tables/activation

Same argument goes for the list of pings. @Iinh you mentioned you'd be interested in working on this?

Iinh:
Fixes #285.

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -33,20 +33,29 @@
   }
 </script>
 
+<style>
+  .item-browser {
+    max-height: 400px;
+    overflow: scroll;
+  }
+</style>
+
 {#if !items.length}
   <p>Currently, there are no {itemType} available for {items.name}</p>
 {:else}
   <FilterInput onChangeText={filterItems} />
-  <ul>
-    {#each filteredItems as item}
-      <li>
-        <a href={getItemURL(appName, itemType, item.name)}>{item.name}</a>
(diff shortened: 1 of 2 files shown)


src/components/ItemList.svelte
```diff
@@ -33,20 +33,29 @@
...
-    {#each filteredItems as item}
-      <li>
-        <a href={getItemURL(appName, itemType, item.name)}>{item.name}</a>
-        <i><Markdown text={item.description} /></i>
-      </li>
-    {:else}
-      <p>Your search didn't match any {itemType}.</p>
-    {/each}
-  </ul>
+  <div class="item-browser mx-auto my-4 p-2">
```
wlach, line 47: I wonder if it's better to define these properties here or in the class itself? It makes some intuitive sense that we'd do padding inline (since they aren't really a property of the "component"). Probably not worth worrying about too much at this juncture in any case.

wlach (approved):
Looks great! Thanks @Iinh. I'll go ahead and merge

PR #294, opened 2020-12-22
Initial attempt at using the protocol design system

wlach:
This is an initial attempt at styling the site using Mozilla's [protocol design system](https://protocol.mozilla.org/). It's a bit rough right now, but I think it shows potential. I like how using protocol implies a strong association with Mozilla's values, including the [lean](https://www.mozilla.org/en-US/about/policy/lean-data/) and [transparent](https://wiki.mozilla.org/Firefox/Data_Collection) data practices which are behind products like this.

image

I think for initial phases, using it alongside Tailwind makes sense-- longer term, we may want to move towards extending/using its scss framework (@craigcook recommended looking at this [gpg site](https://github.com/craigcook/gpg-static) for some inspiration)

### Work items (must be addressed before initial merge)

- [ ] Add a proper glean dictionary logo to the top-left of the site header
- [ ] Figure out appropriate heights for the various headers we're using
- [ ] Use a proper footer
- [ ] Fix the BigQuery schema viewer (some conflicts between our own styling and what's provided, it seems)

@craigcook I'd love to hear your thoughts on this, if you have any (no need to rush, I'll be on break after today).
@Iinh (and anyone else reading this) lmk if you have any thoughts too!

### Pull Request checklist

- [ ] The pull request has a descriptive title (and a reference to an issue it
      fixes, if applicable)
- [ ] All tests and linter checks are passing
- [ ] The pull request is free of merge conflicts

PR #297, opened 2020-12-26
Add git version information in footer (fixes #295)

### LINKED ISSUE #295: Add Git version information to footer
When working on [iodide](https://alpha.iodide.io) I added a handy feature that would tell you the version that was deployed in the footer. We should add something similar for the glean dictionary. This will become more important as we move to production deploys and it becomes less obvious what version is in use.

image

This is pretty simple in outline:

* Run `git rev-list HEAD --max-count=1` from rollup and capture the result into a variable
* Use the replace plugin (here's an existing example: https://github.com/mozilla/glean-dictionary/blob/b7ae2ced6684bacbaed49bd99ef21f0b75f1c91e/rollup.config.js#L53) to inject it into the footer.

The current footer will probably change a bunch in the near future, but it's probably fine to just put the data next to the existing items on the right for now.

Fenn-CS:
Solves #295, uses `rollupjs bundler` `replace` plugin and `child_process`, `execSync` to get an attach git rev information in footer.

### View

Screenshot_2020-12-26 Glean Dictionary Prototype

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -4,7 +4,7 @@ import resolve from "@rollup/plugin-node-resolve";
 import commonjs from "@rollup/plugin-commonjs";
 import livereload from "rollup-plugin-livereload";
 import { terser } from "rollup-plugin-terser";
-import { spawn } from "child_process";
+import { spawn, execSync } from "child_process";
 import sveltePreprocess from "svelte-preprocess";
 
 const production = !process.env.ROLLUP_WATCH;
@@ -53,6 +53,9 @@ export default {
     replace({
       __GOOGLE_ANALYTICS_ID__:
         process.env.CONTEXT === "production" && process.env.GOOGLE_ANALYTICS_ID,
+      __GLEAN_VERSION__: execSync("git rev-list HEAD --max-count=1")
+        .toString()
+        .trim(),
     }),
     // If you have external dependencies installed from
     // npm, you'll most likely need these plugins. In
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -53,7 +53,7 @@ export default {
     replace({
       __GOOGLE_ANALYTICS_ID__:
         process.env.CONTEXT === "production" && process.env.GOOGLE_ANALYTICS_ID,
-      __GLEAN_VERSION__: execSync("git rev-list HEAD --max-count=1")
+      __VERSION__: execSync("git rev-list HEAD --max-count=1")
         .toString()
         .trim(),
     }),
--- a/src/components/Footer.svelte
+++ b/src/components/Footer.svelte
@@ -1,8 +1,5 @@
 <script>
   import MozillaLogo from "./icons/MozillaLogo.svelte";
-
-  let rev = "__GLEAN_VERSION__";
-  let revLink = `https://github.com/mozilla/glean-dictionary/tree/${rev}`;
 </script>
 
 <style>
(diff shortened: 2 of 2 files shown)


rollup.config.js
```diff
@@ -53,6 +53,9 @@ export default {
     replace({
       __GOOGLE_ANALYTICS_ID__:
         process.env.CONTEXT === "production" && process.env.GOOGLE_ANALYTICS_ID,
+      __GLEAN_VERSION__: execSync("git rev-list HEAD --max-count=1")
```
wlach, line 56: I think just `__VERSION__` is better here, `__GLEAN_VERSION__` might imply that the version has something to do with the [Glean SDK](https://github.com/mozilla/glean) (which it does not)

```suggestion
      __VERSION__: execSync("git rev-list HEAD --max-count=1")
```

src/components/Footer.svelte
```diff
@@ -37,5 +38,8 @@
       <a href="https://github.com/mozilla/glean-dictionary/issues">Issues</a>
     </li>
     <li><a href="https://github.com/mozilla/glean-dictionary">Source</a></li>
+    <li>
+      <span>Glean - revision </span><a href={revLink}>{rev.substring(0, 10)}</a>
```
wlach, line 42: We probably ought to rethink the footer in #294, for now I think just a parenthetical next to "Source" ought to do the trick. Saying "Glean version" is problematic for the same reason as I suggest above. 

```suggestion
    <li><a href="https://github.com/mozilla/glean-dictionary">Source</a>
      <span>(</span><a href={`https://github.com/mozilla/glean-dictionary/tree/${rev}`}>{rev.substring(0, 10)})</a>
```

src/components/Footer.svelte
```diff
@@ -1,5 +1,8 @@
 <script>
   import MozillaLogo from "./icons/MozillaLogo.svelte";
+
+  let rev = "__GLEAN_VERSION__";
```
wlach, line 4:
```suggestion
  const rev = "__REVISION__";
```

src/components/Footer.svelte
```diff
@@ -1,5 +1,8 @@
...
+  let revLink = `https://github.com/mozilla/glean-dictionary/tree/${rev}`;
```
wlach, line 5: Since we only use this once, I wouldn't define a variable for it.

```suggestion
```

src/components/Footer.svelte
```diff
@@ -11,12 +14,10 @@
     @apply py-4;
   }
   .project-links {
-    @apply grid;
-    @apply grid-cols-4;
-    @apply justify-items-center;
-    @apply p-0;
-    @apply m-0;
-    @apply gap-2;
+    @apply flex;
```
wlach, line 17: Seems like a reasonable simplification! 😄

wlach (changes requested):
This looks good! I have some minor requests for changes, note that you might need to update the formatting after applying my suggestions.

src/components/Footer.svelte
Fenn-CS: @wlach I did eliminate this rev variable constantly. What do you think?

Fenn-CS:
Requested changes done, variables for revision within component removed, as you said there not being used multiple times.

src/components/Footer.svelte
wlach: Yeah, I was on the fence about it too-- in the end I think it is better to keep it, as it makes the logic below a little more understandable. Can you make that change? After that, I'd be happy to merge this. Pre-emptively approving.

wlach (approved):
r+ after recreating the temporary variable

Fenn-CS:
Done!

PR #300, opened 2021-01-03
Added tooltips for context help in ping and metric fields. Fixes  #293

### LINKED ISSUE #293: Add tooltips with contextual help for application, ping and metric fields
One trick we've been using with internal dashboards is to have help icons which, on hover, pop up contextual help to give people an idea of what the field means. See for example, the [desktop funnel dashboard](https://github.com/mozilla/funnel):

image

We should do something similar in the Glean dictionary. For example on the metrics page, we could put a help icon beside each of these fields:

image
https://glean-dictionary-dev.netlify.app/apps/burnham/metrics/mission-status

Many of the metric parameters are defined in the Glean documentation. For example, metric parameters are defined here:

https://mozilla.github.io/glean/book/user/metric-parameters.html

I think we can probably just lift the definitions and put them in here verbatim. It may be good idea to also *link* to their definition in the glean documentation, which people could visit for further reference (the old glean dictionary [attempts to do this](https://glean-dictionary.netlify.app/?metric=perf_awesomebar_bookmark_suggestions), although it is incorrect in some places). You can look at the code here:

https://github.com/mozilla/probe-dictionary/blob/glean-only/src/components/GleanProbeDetails.svelte#L105

For rendering the tooltip, we can probably just use a library like [sveltejs-tippy](https://github.com/mdauner/sveltejs-tippy) -- maybe best to define our own component based on it and create a story, as usual.

Fenn-CS:
This PR addresses #293 

## Method
- Introduce an info icon svg
- Created a ToolTip component, that accepts a child element on which the tooltip would linger over.

## Notes 
 - Tailwind was not used immediately for convenience (styles can easily be swapped if approach is approved) 
 - Thinking of an effective way to keep the tooltips hanging for a while after hover, so users can click on link in tooltips.
 - Check metrics detail page to see how it works

## Preview

tooltip

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6365,8 +6365,7 @@
     "@popperjs/core": {
       "version": "2.5.4",
       "resolved": "https://registry.npmjs.org/@popperjs/core/-/core-2.5.4.tgz",
-      "integrity": "sha512-ZpKr+WTb8zsajqgDkvCEWgp6d5eJT6Q63Ng2neTbzBO76Lbe91vX/iVIW9dikq+Fs3yEo+ls4cxeXABD2LtcbQ==",
-      "dev": true
+      "integrity": "sha512-ZpKr+WTb8zsajqgDkvCEWgp6d5eJT6Q63Ng2neTbzBO76Lbe91vX/iVIW9dikq+Fs3yEo+ls4cxeXABD2LtcbQ=="
     },
     "@reach/router": {
       "version": "1.3.3",
@@ -30267,6 +30266,14 @@
         }
       }
     },
+    "sveltejs-tippy": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/sveltejs-tippy/-/sveltejs-tippy-3.0.0.tgz",
+      "integrity": "sha512-LAfQikm61AeqZW1hCFWUsmGEMmhMHDvrPV0JhbT38Ell9SqBjmFaY//+YKUxaJvx6nLMxGQdlcHA1rDg2Ml8tQ==",
+      "requires": {
+        "tippy.js": "~6.0.1"
+      }
+    },
     "svgo": {
       "version": "1.3.2",
       "resolved": "https://registry.npmjs.org/svgo/-/svgo-1.3.2.tgz",
@@ -31006,6 +31013,14 @@
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/tinydate/-/tinydate-1.3.0.tgz",
       "integrity": "sha512-7cR8rLy2QhYHpsBDBVYnnWXm8uRTr38RoZakFSW7Bs7PzfMPNZthuMLkwqZv7MTu8lhQ91cOFYS5a7iFj2oR3w=="
+    },
+    "tippy.js": {
+      "version": "6.0.3",
+      "resolved": "https://registry.npmjs.org/tippy.js/-/tippy.js-6.0.3.tgz",
+      "integrity": "sha512-buQF6HugTA4YeSA/s9xBhu0ferEAx7sRSn45G+Juh3p+Dz3vEKfxtqdg6JUnHIWDCV/r0u8Lrobs9AB4dtYcUQ==",
+      "requires": {
+        "@popperjs/core": "^2.1.0"
+      }
     },
     "tmp": {
       "version": "0.0.33",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,5 +1,6 @@
 import replace from "@rollup/plugin-replace";
 import svelte from "rollup-plugin-svelte";
+import postcss from "rollup-plugin-postcss";
 import resolve from "@rollup/plugin-node-resolve";
 import commonjs from "@rollup/plugin-commonjs";
 import livereload from "rollup-plugin-livereload";
@@ -39,6 +40,9 @@ export default {
     file: "public/build/bundle.js",
   },
   plugins: [
+    postcss({
+      plugins: [],
+    }),
     svelte({
       // enable run-time checks when not in production
       dev: !production,
@@ -56,6 +60,10 @@ export default {
       __VERSION__: execSync("git rev-list HEAD --max-count=1")
         .toString()
         .trim(),
+      // workaround the way sveltejs-tippy imports tippy: https://github.com/mdauner/sveltejs-tippy/issues/117
+      "process.env.NODE_ENV": JSON.stringify(
+        production ? "production" : "development"
+      ),
     }),
     // If you have external dependencies installed from
     // npm, you'll most likely need these plugins. In
--- a/src/components/HelpHoverable.svelte
+++ b/src/components/HelpHoverable.svelte
@@ -4,7 +4,7 @@
 
   export let content;
   export let link;
-  export let props = {
+  let props = {
     content,
     allowHTML: true,
     placement: "top",
--- a/src/components/HelpHoverable.svelte
+++ b/src/components/HelpHoverable.svelte
@@ -1,0 +1,25 @@
+<script>
+  import tippy from "sveltejs-tippy";
+  import InfoIcon from "./icons/InfoIcon.svelte";
+
+  export let content;
+  export let link;
+  export let props = {
+    content,
+    allowHTML: true,
+    placement: "top",
+  };
+</script>
+
+<style>
+  .main {
+    @apply inline-block;
+    @apply text-gray-500;
+  }
+</style>
+
+<div class="main">
+  <a href={link} use:tippy={props}>
+    <InfoIcon />
+  </a>
+</div>
--- a/src/data/help.js
+++ b/src/data/help.js
@@ -1,0 +1,17 @@
+export default {
+  bugs: {
+    text:
+      "Required. A list of bugs (e.g. Bugzilla or GitHub) that are relevant to this metric. For example, bugs that track its original implementation or later changes to it.",
+    link: "https://mozilla.github.io/glean/book/user/metric-parameters.html",
+  },
+  lifetime: {
+    text:
+      "Defines the lifetime of the metric. Different lifetimes affect when the metrics value is reset.",
+    link: "https://mozilla.github.io/glean/book/user/metric-parameters.html",
+  },
+  send_in_pings: {
+    text:
+      "Defines which pings the metric should be sent on. If not specified, the metric is sent on the 'default ping', which is the events ping for events and the metrics ping for everything else. Most metrics don't need to specify this unless they are sent on custom pings.",
+    link: "https://mozilla.github.io/glean/book/user/metric-parameters.html",
+  },
+};
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -1,9 +1,10 @@
 <script>
   import { getMetricData } from "../state/api";
   import { getMetricBigQueryURL } from "../state/urls";
-
   import Markdown from "../components/Markdown.svelte";
   import NotFound from "../components/NotFound.svelte";
+  import HelpHoverable from "../components/HelpHoverable.svelte";
+  import helpText from "../data/help";
 
   export let params;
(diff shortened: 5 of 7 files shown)


src/pages/MetricDetail.svelte
```diff
@@ -75,7 +77,12 @@
   </p>
   <table class="metrics-table">
     <tr>
-      <td>Relevant Bugs</td>
+      <td>
+        Relevant Bugs
+        <ToolTip top={true} tip={getHelpMessage('bugs')}>
```
wlach, line 82: It probably makes sense to create a small component for this (in another project, I called a similar thing `HelpHoverable`)

src/helplist.tooltip.js
```diff
@@ -0,0 +1,20 @@
+export default {
```
wlach, line 1: This is the right idea, but I'd probably put this file in `src/data/help.js` (I don't think there's any need to specify that they have to do with tooltips specifically)

src/helplist.tooltip.js
```diff
@@ -0,0 +1,20 @@
...
+  bugs: {
+    text:
+      "Required. A list of bugs (e.g. Bugzilla or GitHub) that are relevant to this metric. For example, bugs that track its original implementation or later changes to it.",
+    more_info:
```
wlach, line 5: I think we can just call this `link`, for simplicity

```suggestion
    link:
```

wlach (changes requested):
This looks like a great start! A few things in addition to the comments:

1. The tooltips sometimes go off the screen-- I think using [sveltejs-tippy](https://github.com/mdauner/sveltejs-tippy) (which builds off the tippy library) instead of rolling our own should fix it.

image

2. I think instead of persisting the tooltip so the user can click "Learn More", we can just make the help icons link to the documentation. This is what the [old glean dictionary does](https://glean-dictionary.netlify.app/?metric=logins_store_write_query_error_count) and I think it works ok.
3. The pull request title should specify the issue fixed (#293). See the [contributing docs](https://github.com/mozilla/glean-dictionary/blob/main/CONTRIBUTING.md#opening-pull-requests).

wlach:
It looks like sveltejs-tippy including the tippy library directly causes some minor issues. In particular, we need to import a plugin to postprocess the css, as well. This is the same as mdauner/sveltejs-tippy#117

After fixing this, I noticed the additional problem of the tippy library referencing something we don't have in a browser context (`process`). I fixed this using rollup-plugin-replace, but rollup-plugin-babel *might* do the same thing (it seems to work for sveltejs-tippy's demo) and would be a cleaner solution: 

https://github.com/mdauner/sveltejs-tippy/blob/master/package.json#L35
https://github.com/mdauner/sveltejs-tippy/blob/master/rollup.config.js#L50

The following patch, if applies, makes things work again (though the styling still needs some work):

https://gist.github.com/wlach/bdaab17805e0a24fdabff718d15e9a12

Do you want to experiment with applying that, and maybe the babel thing I mention?

Fenn-CS:
Ok, I see! I could fix that now that it's clear, however, if we style need to do some styling, Why don't we eliminate this dependency all together, the overflow issue can easily be stripped down with a few tailwind classes.

wlach:
> Ok, I see! I could fix that now that it's clear, however, if we style need to do some styling, Why don't we eliminate this dependency all together, the overflow issue can easily be stripped down with a few tailwind classes.

I think the underlying [tippy library](https://atomiks.github.io/tippyjs/) does a lot of stuff, I'm a bit skeptical that the behaviour can be reproduced just with CSS. You're definitely welcome to try though!

Fenn-CS:
I checked the tipsy library again, it offers alot (way more than we need) but the flexibility is enormous so I went with it.

package-lock.json
```diff
@@ -6365,8 +6365,7 @@
     "@popperjs/core": {
```
wlach, line 6365: It looks like there's some unwanted changes in here. Can you try reverting the changes to `package-lock.json` and `package.json` and *just* installing sveltejs-tippy?

src/components/ToolTip.svelte
```diff
@@ -0,0 +1,43 @@
+<script context="module">
+  import HelpStrings from "../data/help";
+
+  export const getHelpMessage = (key) => {
```
wlach, line 4: I don't think we need the indirection of this function (and the one below it) -- I'd just grab the message / links inside the component and pass them as properties here.

src/components/ToolTip.svelte
```diff
@@ -0,0 +1,43 @@
...
+    if (HelpStrings[key]) {
+      return HelpStrings[key].link === undefined
+        ? ""
+        : `${HelpStrings[key].link}`;
+    }
+    return "No link found.";
+  };
+</script>
+
+<script>
```
wlach, line 21: Typically svelte components only have one script block

```suggestion
```

src/components/ToolTip.svelte
```diff
@@ -0,0 +1,43 @@
...
+  import tippy from "sveltejs-tippy";
+
+  export let props = {
```
wlach, line 24: It's a bit confusing to pass this props stucture into a component. I would construct this dynamically below with `use:tippy`.

src/components/ToolTip.svelte
```diff
@@ -0,0 +1,43 @@
...
+</script>
+
+<style>
+  .main {
+    @apply inline-block;
+  }
+</style>
+
+<div class="main">
+  <section use:tippy={props}>
```
wlach, line 38: I'm pretty sure this section element is unnecessary.

src/components/ToolTip.svelte
```diff
@@ -0,0 +1,43 @@
+<script context="module">
```
wlach, line 1: I think it makes more sense to define this as a class specifically oriented around help, like `HelpHoverable`. We can consider generalizing this into a tooltip class that wraps sveltejs-tippy later if needed.

This would imply putting the help icon directly inside this component (only the text and links would be parametrized).

src/components/ToolTip.svelte
```diff
@@ -0,0 +1,43 @@
...
+    return "No link found.";
+  };
+</script>
+
+<script>
+  import tippy from "sveltejs-tippy";
+
+  export let props = {
+    content: "<span class='tooltip'>Tooltip Info</span>",
+    allowHTML: true,
```
wlach, line 26: This can always be `true` for a HelpHoverable.

src/components/ToolTip.svelte
```diff
@@ -0,0 +1,43 @@
...
+    placement: "top",
```
wlach, line 27: Likewise, let's always make the placement "top" for HelpHoverable.

src/components/ToolTip.svelte
```diff
@@ -0,0 +1,43 @@
...
+  };
+</script>
+
+<style>
+  .main {
+    @apply inline-block;
```
wlach, line 33: Might make sense to use a more subtle shade than blue for the help icon.

```suggestion
    @apply inline-block;
    @apply text-gray-500;
```

wlach (changes requested):
This looks really good! I have some concerns about the implementation, hopefully they won't be too hard to address. Once we have an initial draft of this working, we can start using this in more places. :)

package-lock.json
```diff
@@ -2660,53 +2660,6 @@
       "integrity": "sha512-6U71C2Wp7r5XtFtQzYrW5iKFT67OixrSxjI4MptCHzdSVlgabczzqLe0ZSgnub/5Kp4hSbpDB1tMytZY9pwxxA==",
```
wlach, line 2660: I think some of these changes are incorrect (e.g. it looks like eslint got removed from the dependency list)-- probably the right thing to do here is just to revert the changes to both `package-lock.json` and `package.json` and re-add the dependency using `npm add`

wlach (changes requested):
Hey, so a few things:

I think maybe some of my feedback got a bit garbled / misunderstood. Just to reframe again (maybe a bit more concretely):

1. I would rename ToolTip.svelte to HelpHoverable.svelte and define the various types of behaviour we want (the info icon, etc.) inside it.
2. The file you renamed to HelpHoverable.js can just be removed (since it's just an abstraction layer over the actual data which we don't need, at least not right now). 
 
After the above suggestions are applied, `MetricDetail.svelte` should look something like this:

```svelte
<script>
  ...
  import HelpHoverable from "../components/HelpHoverable.svelte"
  import HelpStrings from "../data/help";
  ...
</script>

...
<td>Send in Pings <HelpHoverable content={help.send_in_pings.content} link={help.send_in_pings.link} /></td>
...
```

rollup.config.js
```diff
@@ -56,6 +60,9 @@ export default {
       __VERSION__: execSync("git rev-list HEAD --max-count=1")
         .toString()
         .trim(),
+      "process.env.NODE_ENV": JSON.stringify(
```
wlach, line 63: Let's add a quick note about why we added this

```suggestion
      // workaround the way sveltejs-tippy imports tippy: https://github.com/mdauner/sveltejs-tippy/issues/117
      "process.env.NODE_ENV": JSON.stringify(
```

src/data/help.js
```diff
@@ -0,0 +1,18 @@
...
+  lifetime: {
+    text:
+      "Defines the lifetime of the metric. Different lifetimes affect when the metrics value is reset.",
+    link: "https://mozilla.github.io/glean/book/user/metric-parameters.html",
+  },
+  send_in_pings: {
+    text:
+      "Defines which pings the metric should be sent on. If not specified, the metric is sent on the 'default ping', which is the events ping for events and the metrics ping for everything else. Most metrics don't need to specify this unless they are sent on custom pings.",
+    // link:
+    //   "https://mozilla.github.io/glean/book/user/metric-parameters.html",
```
wlach, line 16: I imagine you had this commented out for testing purposes.

```suggestion
    link: "https://mozilla.github.io/glean/book/user/metric-parameters.html",
```

src/pages/MetricDetail.svelte
```diff
@@ -1,9 +1,11 @@
 <script>
   import { getMetricData } from "../state/api";
   import { getMetricBigQueryURL } from "../state/urls";
-
   import Markdown from "../components/Markdown.svelte";
   import NotFound from "../components/NotFound.svelte";
+  import InfoIcon from "../components/icons/InfoIcon.svelte";
+  import HelpHoverable from "../components/HelpHoverable.svelte";
+  import help from "../data/help";
```
wlach, line 8: Let's give this variable a slightly more descriptive name (I know it technically includes links as well, but I think `helpText` is the main outline of what this data is about.

```suggestion
  import helpText from "../data/helpText";
```

wlach (changes requested):
This looks great, just have a few last requests for changes before we land.

src/components/HelpHoverable.svelte
```diff
@@ -0,0 +1,23 @@
+<script>
+  import tippy from "sveltejs-tippy";
+
+  export let props = {
+    content: "<span class='tooltip'>Tooltip Info</span>",
+    allowHTML: true,
+    placement: "top",
+  };
+  export let link;
```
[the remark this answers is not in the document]
Fenn-CS: Makes sense

src/components/HelpHoverable.svelte
```diff
@@ -0,0 +1,25 @@
+<script>
+  import tippy from "sveltejs-tippy";
+  import InfoIcon from "./icons/InfoIcon.svelte";
+
+  export let content;
+  export let link;
+  export let props = {
```
wlach, line 7: We don't need to export `props` anymore.

```suggestion
  let props = {
```

wlach (approved):
Looks great, just one change (which I'll make), then we can land. @Iinh -- we'll need to rebase the protocol work on top of this at some point-- the amount of tailwind styling in here is pretty minimal so hopefully merging won't be so bad.

wlach:
Thanks for your persistence on this @Fenn-CS!

PR #305, opened 2021-01-05
Simplify Pagination state with Context API

Iinh:
Further simplify pagination logic using Svelte stores to manage the state of paginated items. 
### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/__snapshots__/storyshots.test.js.snap
+++ b/src/__snapshots__/storyshots.test.js.snap
@@ -645,6 +645,7 @@ exports[`Storyshots ItemList Default 1`] = `
     </div>
   </div>
   
+  
 </section>
 `;
 
@@ -681,7 +682,7 @@ exports[`Storyshots Pagination Default 1`] = `
   of
   
     <code>
-      21
+      11
     </code>
     
   (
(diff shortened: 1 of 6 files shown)


wlach (changes requested):
This definitely looks like an improvement-- however, I feel like maybe we should use Svelte's context API instead of stores here? I believe stores are meant for application global state, which this isn't quite. Context (which allows a child component to access a parent's state) seems like exactly what we need here:

https://svelte.dev/tutorial/context-api

Let me know what you think.

Iinh:
@wlach My understanding of context api is while it allows the descendants to access top component state, it is only read access. Context data is not reactive: setContext() and getContext() can only used at component initialization: https://svelte.dev/docs#setContext. This is restrictive in our case because to let the user change the current page and have the page position data updated accordingly in Pagination, the data needs to be reactive and we need write access. I agree that using context api provides more context(no pun intended!) to the state relationship between ItemList and Pagination. To use context api here, I will still have to use a store and set it as value in the context, which would add another layer to the state. If you prefer that approach I'll refactor and add context api  👍

wlach:
> To use context api here, I will still have to use a store and set it as value in the context, which would add another layer to the state. If you prefer that approach I'll refactor and add context api 👍

I think that makes the most sense based on my read of the documentation! My problem with the existing PR isn't that it uses a store so much that it uses a global one.

wlach (changes requested):
Looks good, just a few more small cleanup nits.

wlach (approved):
Looks good now, thanks!

PR #306, opened 2021-01-08
Add a link to GLAM for the selected metric

robhudson:


src/pages/MetricDetail.svelte
```diff
@@ -10,6 +10,34 @@
...
+    };
+    if (Object.keys(map).includes(app)) {
+      const p = map[app];
+      return function makeGlamUrl(metric) {
+        return `https://glam.telemetry.mozilla.org/${p.product}/probe/${metric}/explore?app_id=${p.app_id}`;
+      };
+    }
+
+    // The app isn't one GLAM supports so return nothing.
+    return null;
```
wlach, line 36: `undefined` is a little more idiomatic here

```suggestion
    return undefined;
```

src/pages/MetricDetail.svelte
```diff
@@ -197,6 +225,15 @@
         {/each}
       </td>
     </tr>
+    {#if glamUrl}
+      <tr>
+        <td>GLAM</td>
+        <td>
+          <a
+            href={glamUrl(metric.bigquery_names.glam_etl_name)}>{glamUrl(metric.bigquery_names.glam_etl_name)}</a>
```
wlach, line 233: Showing the full URL in the UI seems a little verbose (you can probably skip the `const` declaration above if you take it out, as well).

```suggestion
            href={glamUrl(metric.bigquery_names.glam_etl_name)}>{metric.bigquery_names.glam_etl_name}</a>
```

wlach (changes requested):
This looks and works great! Just a few minor changes and we'll be good to land.

src/pages/MetricDetail.svelte
```diff
@@ -197,6 +225,15 @@
         {/each}
       </td>
     </tr>
+    {#if glamUrl}
+      <tr>
+        <td>GLAM</td>
```
robhudson, line 230: Would you prefer this titled "GLAM URL" instead?

wlach: Nope! I think just `GLAM` is fine. We can add a help icon when @Fenn-CS finishes #293

src/pages/MetricDetail.svelte
robhudson: I updated this but still need the `const` for the `{#if}` check to show the table row.

wlach (approved):
Looks great, thanks!

src/pages/MetricDetail.svelte
```diff
@@ -197,6 +225,15 @@
         {/each}
       </td>
     </tr>
+    {#if glamUrl}
```
wlach, line 228: Argh, sorry, did some last minute testing and it looks like this doesn't do the right thing for events (it incorrectly suggests we can view them in GLAM). Maybe the suggestion below will fix that?

```suggestion
    {#if glamUrl && metric.type === "event"}
```

wlach (changes requested):

src/pages/MetricDetail.svelte
robhudson: Cool, thanks. I added a condition similar to one just above it. Let me know if that's correct. It worked locally for me.

wlach (approved):
Awesome. I'm going to hold off on landing this until #312 is merged so as not to cause unnecessary merge conflicts.

PR #307, opened 2021-01-08
Improve pill aesthetics

wlach:
Use better font, border choices -- more similar to GitHub labels:

https://primer.style/css/components/labels

Just something I worked on while looking at something else.

Before:

image

After:

image

### Pull Request checklist

      fixes, if applicable)


Iinh (approved):
Thanks @wlach! Note that the Tailwind syntax used here is going to be converted to SCSS with Protocol.

wlach:
> Thanks @wlach! Note that the Tailwind syntax used here is going to be converted to SCSS with Protocol.

👍 Yeah, it should be a pretty straightforward transformation I hope!

PR #312, opened 2021-01-11
Replace TailwindCSS with Protocol and SCSS for styling

Iinh:
Continuation from the work started in [PR#294](https://github.com/mozilla/glean-dictionary/pull/294) to transition to Protocol from Tailwindcss.
- Add SCSS support for Svelte
- Add SCSS support for Storybook
- Use Protocol: right now only what's needed from the package is copied over. I had to copy the files instead of importing for customization purposes, such as replacing the type scale setting from regular to condensed.

Since the goal was to remove Tailwind, my focus was less on making sure the transformation looks pixel perfect. There are sections where styling still looks awkward, I'd appreciate the feedback on this front. 

Work to be done in follow up PRs:
- Improve header from its current minimal state
- Abstract the logic used for App/Metric/Ping/TableDetails description table to a separate component. Right now styling is repeated in these 4 components.  

### Pull Request checklist

      fixes, if applicable)


.storybook/preview.js
```diff
@@ -1 +1 @@
-import "./story.css";
+import "!style-loader!css-loader!sass-loader!../node_modules/@mozilla-protocol/core/protocol/css/protocol.scss";
```
Iinh, line 1: My current blocker: I had to use sass-loader so storybook can read scss, but the snapshot test doesn't like this. Trying to figure out a way to work around this

.storybook/main.js
```diff
@@ -10,9 +11,26 @@ module.exports = {
...
     svelteLoader.options = {
       ...svelteLoader.options,
-      preprocess: sveltePreprocess({ postcss: true }),
+      preprocess: sveltePreprocess({
+        postcss: true,
+        defaults: {
+          style: "scss",
+        },
+        scss: {
+          prependData: `@import 'src/scss/protocol/includes/_lib.scss';`,
```
wlach, line 20:
```suggestion
          prependData: `@import 'node_modules/@mozilla-protocol/core/protocol/css/includes/_lib.scss/scss/protocol/includes/_lib.scss';`,
```

src/__snapshots__/storyshots.test.js.snap
```diff
@@ -5,7 +5,7 @@ exports[`Storyshots Breadcrumb App Details Page 1`] = `
   class="storybook-snapshot-container"
 >
   <ol
-    class="flex border-b-2 py-2 divide-x "
+    class="flex py-2 divide-x "
```
wlach, line 8: I think the storyshot snaps might need to be regenerated?

wlach (changes requested):
This looks great! 

My main feedback at this point is that I think some files shouldn't be copied over into the repo. In particular, I believe we should have everything we need (after running `npm install`) in `node_modules/@mozilla-protocol`. For the fonts, I believe we should be able to get rollup to copy them over from there into `public/`.

Let me know if I'm missing something.

.storybook/main.js
Iinh: The type scale is defined in [/includes/lib](https://github.com/mozilla/glean-dictionary/blob/5e8a4c0f4267b7d195ed43d2eedd6b10ac6c2654/src/scss/protocol/includes/_lib.scss#L43), which then gets the value for sizing from [/includes/themes](https://github.com/mozilla/glean-dictionary/blob/5e8a4c0f4267b7d195ed43d2eedd6b10ac6c2654/src/scss/protocol/includes/_themes.scss#L167). I could just directly update this in node_modules/@mozilla-protocol, but since `node_modules` is gitignored, should I remove all the copied scss files, make the edit directly in the package, and un-gitignore `node_modules/@mozilla-protocol`?

wlach: Oh I see, no definitely don't un-gitignore anything in `node_modules/*`! That's meant to be installed dynamically at build time.

I think it would be better if we copied as little as possible from protocol into this repository, and reuse as much as possible as-is from what's installed in `node_modules/`. That way we can just upgrade the protocol dependency as we go to gain new features and fixes as they come in, with hopefully a minimal number of changes inside this repository.

However, I admit that I've never done anything like this before (integrating a sass library into a repository) so there might be something I'm missing. I wonder if @craigcook might be able to give us some guidance on how best to handle this (whether that be copying parts of protocol into glean dictionary, or something else...)

Iinh: Yeah right now only what's needed is copied over, which is the /base folder and a couple files from /components. The reason why the number of migrated files is big is because I can't just pick and choose the component files, I have to also copy the dependencies which are the scss variables and mixins in `/includes`. Not sure if it's necessary, but I did add [a README](https://github.com/mozilla/glean-dictionary/blob/use-protocol-and-scss/src/scss/README.md) that explains the process to minimize confusion for other contributors.

wlach:
Was playing around with the site a bit this afternoon. I think the styling mostly looks really good. There's just two things that stick out:

1. The breadcrumb widget could use some more padding.

image

2. This is a little subjective, but I find the font sizes a little small and hard to read on my screen. I think the defaults for standard text that were in the initial PR were better.

.storybook/main.js
craigcook: I'm not sure I see a reason you couldn't copy the whole Protocol framework into your project rather than referencing `node_modules` directly, though I'm not sure how you'd actually do the copying in your particular setup (I've only done it with Gulp). Since you seem to be copying over some of it already you might as well get whole thing.

And because you're now compiling Sass yourself you can definitely pick-and-choose specific components at the compile step. The `protocol-components.css` file is only provided for use if you're **not** compiling your own Sass, but it ends up being a huge file with EVERY component so it will always be more than you need. You don't need it at all if you're compiling yourself, just `@import` the components you're actually using.

What we've done on other sites is copy Protocol from `node_modules` into some local build folder and all our own SCSS is in a local `scss` folder (or whatever you want to name it). The local SCSS can reference the local copy of Protocol files, and any customization or extensions we need to make to it are only done in our own SCSS (relying on specificity or the cascade to override the original Protocol styling). Definitely don't change any of the Protocol source files since any changes would just be overwritten next time you copy it.

You can also configure the brand theme and type scale in your local SCSS.

wlach: > What we've done on other sites is copy Protocol from `node_modules` into some local build folder and all our own SCSS is in a local `scss` folder (or whatever you want to name it). The local SCSS can reference the local copy of Protocol files, and any customization or extensions we need to make to it are only done in our own SCSS (relying on specificity or the cascade to override the original Protocol styling). Definitely don't change any of the Protocol source files since any changes would just be overwritten next time you copy it.

Is there any reason you'd recommend copying protocol into the project and checking it in, vs. just referencing it via `node_modules`? The latter feels a little more elegant to me.

craigcook: > Is there any reason you'd recommend copying protocol into the project and checking it in, vs. just referencing it via `node_modules`? The latter feels a little more elegant to me.

I suppose not, other than shortening the paths on your imports. I just meant to say you *can* copy it without hurting anything as long as you're not modifying anything.

That's just what we've done on other sites like bedrock or kitsune, which admittedly have a lot more stuff happening during the build process so there might be performance advantages to a local copy in those cases that might not apply to you. And we only copy it into a local build folder; the copy isn't committed to git. 

Some things you _will_ want to copy and commit, such as the fonts, icons, and any scripts you're using from Protocol. Anything that needs to be served to the client, but you could do that manually. All the Protocol SCSS could stay in `node_modules` if that's better for you.

Iinh:
> This is a little subjective, but I find the font sizes a little small and hard to read on my screen. I think the defaults for standard text that were in the initial PR were better.

The reduction in font size was due to the change in type scale to "condensed". If your preference is go back to the standard scale, and the only place that we want condensed typography is the h1 tags, would it be better that we go with your original idea of creating a separate component for it? This will also eliminate the need for theme customization (for now) and let us directly import from `node_modules/`.

src/components/Footer.svelte
```diff
@@ -1,45 +1,43 @@
...
+            <a href="https://github.com/mozilla/glean-dictionary">Source</a>
+          </li>
+        </ul>
+        <p>
+          Alpha Software - revision
+          <a
+            href="https://github.com/mozilla/glean-dictionary/tree/{rev}">{rev.substring(0, 10)}</a>
+        </p>
+        <p class="mzp-c-footer-license">
+          © 2020 Mozilla and other contributors.
```
wlach, line 38:
```suggestion
          © 2020 - 2021 Mozilla and other contributors.
```

src/components/SchemaViewer.svelte
```diff
@@ -56,21 +57,28 @@
...
 <style>
-  .schema-browser {
-    @apply p-2;
-    @apply border-4;
-    @apply border-gray-400;
-    max-height: 400px;
-    overflow: scroll;
+  .schema-viewer {
+    margin-top: 2rem;
+    .schema-browser {
```
wlach, line 62: Interesting, I didn't know it was possible to nest css declarations like this!

.eslintrc.js
```diff
@@ -53,12 +53,14 @@ module.exports = {
...
         "import/no-mutable-exports": "off",
         "import/no-unresolved": "off",
-
         // Temporarily work around a bug in eslint-plugin-svelte3.
         //
         // https://github.com/sveltejs/eslint-plugin-svelte3/issues/41#issuecomment-572503966
         "no-multiple-empty-lines": ["error", { max: 2, maxBOF: 2, maxEOF: 0 }],
       },
+      settings: {
+        "svelte3/ignore-styles": () => true,
```
wlach, line 62: Can you add a comment on why we need to ignore styles? (I assume it's because we're now using sass tokens)

.prettierignore
```diff
@@ -4,3 +4,4 @@ public/data/*
 storybook-static/*
 venv
 src/ga.js
+src/scss/
```
wlach, line 7:
```suggestion
src/scss/

```

package.json
```diff
@@ -56,6 +58,7 @@
     "lunr": "^2.3.9",
     "marked": "^1.2.7",
     "node-fetch": "^2.6.1",
+    "node-sass": "^5.0.0",
```
wlach, line 61: This should technically be a dev dependency (I know that other dependencies aren't specified properly, but we can fix that another time). Maybe we should just fix that in bulk in a followup at this point though, you can leave this for now.

src/App.svelte
```diff
@@ -84,28 +84,70 @@
...
+<GlobalStyles />
+
+<div class="mzp-c-navigation mzp-is-sticky">
+  <div class="mzp-c-navigation-l-content">
+    <div class="mzp-c-navigation-container">
+      <div class="mzp-c-navigation-logo-glean">
+        <a class="glean-logo" href="/">
+          <img
+            src="/img/logos/glean_logo_bw.png"
+            alt="Glean Dictionary logo" /></a>
```
wlach, line 126:
```suggestion
            alt="Glean Dictionary Logo" /></a>
```

package-lock.json
```diff
@@ -6108,6 +6108,12 @@
         }
```
wlach, line 6108: I feel like some dependencies might have gotten a bit mixed up in this file, can you try reverting this change and regenerating it?

src/components/AppAlert.svelte
```diff
@@ -3,10 +3,23 @@
   export let bgColor;
```
wlach, line 3: I think we should rethink this component and where/how it's displayed. That can wait until later though.

src/components/Pill.svelte
```diff
@@ -5,15 +5,13 @@
...
-    @apply py-0;
-    @apply text-sm;
-    @apply leading-5;
-    @apply text-white;
-    @apply font-medium;
+    display: inline-block;
+    border-radius: 9999px;
+    border-color: transparent;
+    padding: 0 $spacing-sm 0 $spacing-sm;
+    @include text-body-sm;
```
wlach, line 12: I wonder if we should move these `@include` directives to the top of the declaration. Might make it a little more clear.

src/components/Title.svelte
```diff
@@ -0,0 +1,15 @@
+<script>
```
wlach, line 1: I wonder if maybe we should call this component `PageTitle`, since that's how it's used?

src/components/icons/InfoIcon.svelte
```diff
@@ -1,3 +1,11 @@
+<style>
+  svg {
+    display: inline-block;
+    width: $spacing-md;
+    color: $color-dark-gray-50;
```
wlach, line 5: I might use a lighter shade of grey, to make the info icon a little less prominent. We want it to be visible, but not draw too much attention when people are skimming the content. I think the old glean dictionary got the contrast levels about right:

image

src/pages/AppDetail.svelte
```diff
@@ -7,36 +7,41 @@
...
   {#if app.deprecated}
     <Pill message="Deprecated" bgColor="#4a5568" />
   {/if}
-  <p class="mt-2">{app.description}</p>
-  <table class="table-header">
+  <p>{app.description}</p>
+  <table class="mzp-u-data-table">
     <tr>
-      <td>Source code Url</td>
+      <td>Source Code URL</td>
```
wlach, line 40: I had been meaning to remove the URL part of this, since it's redundant. May as well change it here, if we're gonna change it.

```suggestion
      <td>Source Code</td>
```

svelte.config.js
```diff
@@ -0,0 +1,16 @@
+const sveltePreprocess = require("svelte-preprocess");
+const autoPrefixer = require("autoprefixer");
+
+module.exports = {
+  preprocess: sveltePreprocess({
+    defaults: {
+      style: "scss",
+    },
+    scss: {
+      prependData: `@import 'src/scss/protocol/includes/_lib.scss';`,
```
wlach, line 10: It looks like we're importing some stuff that isn't in the repository?

wlach (changes requested):
This is looking quite excellent, thank you @Iinh! Two items:

1. I think we can get around needing to import the fonts into `public/` by adding the following symbolic links:

node_modules/@mozilla-protocol/core/protocol/img -> public/img
node_modules/@mozilla-protocol/core/protocol/fonts -> public/fonts

Can you give that a try? You should be able to make the links like so from the root directory:

`ln -s node_modules/@mozilla-protocol/core/protocol/img public/img`

(then just add them to the PR)

2. It seems like storybook building is timing out on netlify, do you have any idea why? I turned it off for now, since the last build took 20 minutes (!) Here's a snapshot of the log, which was 250k+ lines long: https://gist.github.com/wlach/bb799b808500e8ddd9c390907a143ddb

I am personally ok with fixing the second item later (since I only recently added the storybook deploy to the netlify deploy preview, and wasn't really depending on it) but if it's easy to figure out...

src/components/SchemaViewer.svelte
Iinh: You can't do it with css, this is actually sass syntax: https://sass-lang.com/documentation/style-rules ;)

Iinh:
> It seems like storybook building is timing out on netlify, do you have any idea why? I turned it off for now, since the last build took 20 minutes (!) Here's a snapshot of the log, which was 250k+ lines long: https://gist.github.com/wlach/bb799b808500e8ddd9c390907a143ddb

Because we imported the whole Protocol package but are only using some of it, there's a lot of unused CSS selectors, and `svelte-loader` doesn't like this. From reading the build log, I think the reason why it took so long to deploy was because Netlify logged the warnings for every single unused CSS selector, which understandably took a long time since there are hundreds of them. 

My initial approach to solve this ([this commit](https://github.com/mozilla/glean-dictionary/pull/312/commits/ac21dad37be5f71254b83e3d34f6d23e0735147a)) was to find a way to remove the unused CSS selectors using `postcss-purgecss` -- the library that we used with Tailwind. It did remove the majority of the unused CSS, but not all of them. It also threw off the styling, probably because it might have removed some of the global styles. 

Since what's causing the gridlock is the 250k+ lines of warning in Storybook, my next approach was to turn off these warnings completely in the `svelte-loader` setting inside Storybook. I think doing this is safe because this error is Storybook specific, I believe Svelte does not compile unused rules: https://github.com/sveltejs/svelte/issues/697. So this was done in [this commit](https://github.com/mozilla/glean-dictionary/pull/312/commits/9934c1f97de57af9f1d9560419ab48061c4733c0). I think this will solve the issue, although I'm not 100% sure. Is it possible to add me to the Netlify team so I can check?

Other than that, I think I've addressed all of your requested changes. Let me know what you think @wlach.

jest.config.js
```diff
@@ -1,10 +1,10 @@
 module.exports = {
   transform: {
-    "^.+\\.svelte$": "svelte-jester",
+    "^.+\\.svelte$": ["svelte-jester", { preprocess: true }],
     "^.+\\.js$": "babel-jest",
   },
   moduleFileExtensions: ["js", "svelte", "json"],
-  testPathIgnorePatterns: ["/node_modules/", "/public/", "/storybook-static/"],
+  testPathIgnorePatterns: ["/node_modules/", "/public/", "/.storybook"],
```
wlach, line 7: I think `storybook-static` is what's generated by `npm run build-storybook`. I think it should still be in the ignore list.

wlach:
> > It seems like storybook building is timing out on netlify, do you have any idea why? I turned it off for now, since the last build took 20 minutes (!) Here's a snapshot of the log, which was 250k+ lines long: https://gist.github.com/wlach/bb799b808500e8ddd9c390907a143ddb
> 
> ... 
> Since what's causing the gridlock is the 250k+ lines of warning in Storybook, my next approach was to turn off these warnings completely in the `svelte-loader` setting inside Storybook. I think doing this is safe because this error is Storybook specific, I believe Svelte does not compile unused rules: [sveltejs/svelte#697](https://github.com/sveltejs/svelte/issues/697). So this was done in [this commit](https://github.com/mozilla/glean-dictionary/pull/312/commits/9934c1f97de57af9f1d9560419ab48061c4733c0). I think this will solve the issue, although I'm not 100% sure. Is it possible to add me to the Netlify team so I can check?

Great stuff @Iinh, thanks for such a thorough investigation. I just tried re-enabling the storybook build in netlify, and it now completes in a very reasonable 2 minutes and 18 seconds. 🚀 This seems like a very reasonable approach, I have some minor extra comments but this is looking 99% of the way done.

public/fonts/fonts
```diff
@@ -0,0 +1 @@
+node_modules/@mozilla-protocol/core/protocol/fonts
```
wlach, line 1: It looks like you accidentally created the symbolic link *inside* the fonts directory. I forgot to mention that you would want to *delete* the contents of the directory before copying.

Now that I think about it though, I suspect this approach isn't going to work -- since there are also some assets we want to keep in this directory locally (e.g. the glean logo). I think a better approach is to copy the files from protocol *into* this directory at build time, using something like [rollup-plugin-copy](https://www.npmjs.com/package/rollup-plugin-copy). Can you give that a try?

src/components/SchemaViewer.svelte
wlach: Nice!

.storybook/main.js
```diff
@@ -10,9 +12,35 @@ module.exports = {
...
+        },
+        scss: {
+          prependData: `@import '@mozilla-protocol/core/protocol/css/protocol.scss';`,
+        },
+        postcss: {
+          plugins: [autoPrefixer],
+        },
+      }),
+
+      // turn off warning about unused CSS selectors
```
wlach, line 28: Can you give a little bit more context on why we're doing this, for the benefit of someone (maybe even you) looking at it in the future? 😄 

```suggestion
      // turn off warning about unused CSS selectors
```

wlach (changes requested):
A few small things before we land (in addition to the review comments). Thanks, as always, for your patience and persistence:

1. `npm run dev` doesn't seem to work for me anymore, I'm getting a similar error to the one you were getting a few days ago: https://gist.github.com/wlach/8c7fc5abd046537f1d7348bfd8d55203
2. On reflection, I think we probably ought to use the colored Glean logo, if it looks at all reasonable-- I think it's probably more familiar to people, and more consistent with what's in the Glean documentation, etc. I was going to test this out myself, but got stuck on (1).

src/components/ItemList.svelte
```diff
@@ -40,14 +40,18 @@
   .item-browser {
     max-height: 400px;
     overflow: scroll;
+    margin: $spacing-md auto $spacing-md auto;
```
robhudson, line 43: The above CSS looks to be repeated twice.

Iinh:
> npm run dev doesn't seem to work for me anymore, I'm getting a similar error to the one you were getting a few days ago: https://gist.github.com/wlach/8c7fc5abd046537f1d7348bfd8d55203

I only get this error if I compile SCSS with `sass`. Do you still have `sass` in your `package.json`? If yes you need to remove it and install `npm i node-sass` instead (so `svelte-preprocess` doesn't use `sass` as default over `node-sass`)

> On reflection, I think we probably ought to use the colored Glean logo, if it looks at all reasonable

Done! I actually like this header better. Thank you for suggesting this improvement. 

One last thing, I was trying to remove the symbolic links created, but I'm not very familiar with this so could you help check if the links have been removed? Thank you!

wlach:
> > npm run dev doesn't seem to work for me anymore, I'm getting a similar error to the one you were getting a few days ago: https://gist.github.com/wlach/8c7fc5abd046537f1d7348bfd8d55203
> 
> I only get this error if I compile SCSS with `sass`. Do you still have `sass` in your `package.json`? If yes you need to remove it and install `npm i node-sass` instead (so `svelte-preprocess` doesn't use `sass` as default over `node-sass`)

Just to be sure, I tried this with a fresh checkout on your branch and am still getting the same error. Are you sure you checked everything in?

package.json
```diff
@@ -37,15 +38,18 @@
     "eslint-plugin-svelte3": "3.0.0",
     "jest": "^26.6.3",
     "lerna": "3.22.1",
+    "node-sass": "^5.0.0",
```
Iinh, line 41: @wlach Yup everything was checked in, sass was removed and node-sass installed. Your error message https://gist.github.com/wlach/8c7fc5abd046537f1d7348bfd8d55203#file-gistfile1-txt-L31 indicates that you still have `sass` in `node_modules` though, you probably need to do a refresh `npm install`

wlach: As mentioned, I did a fresh checkout and am still seeing the error. I definitely don't have `sass` in `node_modules`:

```
wlach@antwerp glean-dictionary % ls node_modules/sass-*
node_modules/sass-graph:
CHANGELOG.md		package.json		sass-graph.js
bin			parse-imports.js
node_modules		readme.md

node_modules/sass-loader:
CHANGELOG.md	README.md	node_modules
LICENSE		dist		package.json
```

Does it really work for you if you do a fresh checkout? If so I can't imagine what the problem might be...

wlach: Nevermind, just figured it out. I had a copy of `sass` installed globally. I wonder if there's a way to tell glean-dictionary to prefer `node-sass` in that case?

wlach: > Nevermind, just figured it out. I had a copy of `sass` installed globally. I wonder if there's a way to tell glean-dictionary to prefer `node-sass` in that case?

Just googled and I'm not seeing any easy solutions :/ Let's table this for now, I don't think it should block merging (unless you have easy ideas).

public/_redirects
```diff
@@ -1 +0,0 @@
-/*    /index.html   200
```
wlach, line 1: Did you mean to remove this?

wlach (changes requested):
> One last thing, I was trying to remove the symbolic links created, but I'm not very familiar with this so could you help check if the links have been removed? Thank you! @wlach

It looks like the links have been removed, but the files are still in your commit. You probably want to `git rm -r public/fonts public/img` and then make another commit.

Aside from that, just one more question. 😄

wlach (approved):

wlach:
This PR was 💯 -- congrats @Iinh! 🎉

Iinh:
@wlach Thank you 😭 Filing a couple follow up issues from our discussion in this PR.

PR #319, opened 2021-01-14
Add "type" to metric summary (fixes #258)

### LINKED ISSUE #258: Add "type" to metric summary
From @fbertsch:

The first prototype included "types" of each metric on the summary page, we should do the same in this version:

image

Iinh:
Include "type" in metric summary on App and Ping pages to fix #258.

### Pull Request checklist

- [x] The pull request has a descriptive title (and a reference to an issue it fixes, if applicable)

### DIFF
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -46,11 +46,14 @@
   }
 
   table {
+    table-layout: fixed;
+    width: 100%;
     background: $color-light-gray-05;
     border-collapse: collapse;
     margin: auto;
     tr {
       td {
+        word-wrap: break-word;
         border: 1px dotted $color-light-gray-30;
       }
       &:nth-child(odd) td {
@@ -60,7 +63,6 @@
         background: $color-dark-gray-30;
         color: $color-light-gray-05;
         a {
-          font-weight: bold;
           color: $color-light-gray-05;
           &:hover {
             color: $color-blue-40;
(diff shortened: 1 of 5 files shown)


src/components/ItemList.svelte
```diff
@@ -37,31 +37,83 @@
...
+    width: 100%;
+    .table-header td {
+      font-weight: bold;
+    }
+    td {
+      padding: $spacing-xs;
+      border-right: 1px dotted $color-light-gray-30;
+      &:first-child {
+        width: 76%;
+        @media (max-width: $tablet-size) {
```
wlach, line 64: Curious what the reason for these media queries is? FWIW I am not too concerned with what this site looks like on mobile since I expect 99% of its traffic to be on desktop

wlach (changes requested):
Code mostly looks good, but I think we should the layout should be a table with the following format:

| Name | Type | Description |
| --- | --- | --- |
|about_page.libraries_tapped | event | A user tapped on "Libraries that we use" item from About page  |
| ... | ... | ... |

I also wonder if it might be worth experimenting with the existing protocol styles for a [data table](https://protocol.mozilla.org/patterns/atoms/typographic.html#data-tables) -- want to give it a try and see how it looks?

src/components/ItemList.svelte
Iinh: When window size is reduced to about ~800px which is the case if you have 2 browser windows open side by side, long metric type names such as "memory_distribution" or "labeled_counter" would overflow:
Screen Shot 2021-01-14 at 12 18 51 PM
I find that quite annoying so I set up the media queries to prevent it. I can remove it though if you don't find it necessary, it was more of a subjective decision as I tend to have 2 windows in split view quite often.

wlach: Makes sense. Maybe just add a comment to let people know your rationale?

Iinh:
Refactored with Protocol table. I also removed the media queries in favor of simpler styling. Protocol table is responsive, so everything scales down when you reduce the screen size which is nice, but it does make the columns jump around a bit. Let me know what you think @wlach

wlach:
> Refactored with Protocol table. I also removed the media queries in favor of simpler styling. Protocol table is responsive, so everything scales down when you reduce the screen size which is nice, but it does make the columns jump around a bit. Let me know what you think @wlach

Looks good, the jumping around is a bit disconcerting though. Would it be possible to make the first two columns fixed width (I think we know roughly how much space we should need for most cases), then make the last column (description) fit any remaining space?

wlach (changes requested):
Looking really good. One last request: can you make the table header sticky? It would be nice to keep it present as the user scrolls up/down through the list. [This approach](https://css-tricks.com/position-sticky-and-table-headers/) seems to work. Specifically adding this css to the table header elements:

```css
position: sticky;
top: 0;
background-color: white;
```

wlach (approved):

wlach:
Thank you, this is great!

PR #332, opened 2021-01-18
Add human-readable bug and issue links in UI (fixes #191)

### LINKED ISSUE #191: Bugs and issues should be shortened in UI
Right now we display the full URLs to bugzilla bugs and github issues. For example:

image

and:

image

It would be nice to use an easily readable shorthand for these. For bugzilla, a nice convention @robhudson found for glam was:

`bugzil.la/{bug id}` (e.g. `bugzil.la/1234`)

For GitHub issues and pull requests, the convention is:

`mozilla/glean-dictionary#101`

Sometimes there are links to a specific bug comment in the form of a bookmark (e.g. `#c95`). In that case, we should append something to the end of the title (to make it clear that it's a specific comment being referenced). I think for Bugzillla we can just use the full bookmark (since it's relatively short). For GitHub, I'd suggest just appending `-comment` (as `#issuecomment-592187878` is pretty verbose and not overly descriptive).

[bugzil.la/1599877#c25](https://bugzil.la/1599877#c25)
[mozilla-mobile/fenix#6174-comment](https://github.com/mozilla-mobile/fenix/issues/6174#issuecomment-592187878)

In terms of implementation, I would write a small svelte component that takes a string as input and creates a small UI element with a properly formatted title and URL. Please write a story for it, in line with other components we've added recently (see e.g. #155)

wlach:
Instead of using full URLs or context-free numbers, use
idiomatic bugzilla and github shortened forms for bugs and
issues.

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/stories/buglink.stories.js
+++ b/stories/buglink.stories.js
@@ -1,0 +1,19 @@
+import BugLink from "../src/components/BugLink.svelte";
+
+export default {
+  title: "Bug Links",
+};
+
+export const Bugzilla = () => ({
+  Component: BugLink,
+  props: {
+    ref: "https://bugzilla.mozilla.org/1234",
+  },
+});
+
+export const GitHub = () => ({
+  Component: BugLink,
+  props: {
+    ref: "https://github.com/mozilla-mobile/fenix/issues/1234",
+  },
+});
--- a/src/__snapshots__/storyshots.test.js.snap
+++ b/src/__snapshots__/storyshots.test.js.snap
@@ -281,6 +281,32 @@ exports[`Storyshots Breadcrumb Ping Page 1`] = `
 </section>
 `;
 
+exports[`Storyshots Bug Links Bugzilla 1`] = `
+<section
+  class="storybook-snapshot-container"
+>
+  <a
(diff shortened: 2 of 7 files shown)


stories/buglink.stories.js
```diff
@@ -0,0 +1,19 @@
...
+};
+
+export const Bugzilla = () => ({
+  Component: BugLink,
+  props: {
+    ref: "https://bugzilla.mozilla.org/1234",
+  },
+});
+
+export const GitHub = () => ({
```
Iinh, line 14: I noticed that this component name was rendered as "Git Hub" instead of "Github" in Storybook. I think it's just a Storybook bug.

Iinh (approved):
Looks good!

stories/buglink.stories.js
wlach: Right, this is probably just how storybook interprets camel case. I tried to see if there was a way to override the title for a subcomponent, but didn't find one offhand. I think let's just leave this for now, since it doesn't impact the UX of the actual app. :)

PR #333, opened 2021-01-18
'Cardify' the application list (fixes #318)

### LINKED ISSUE #318: "Cardify" the application list
Currently the applications are rendered as one long list in the main page: https://glean-dictionary-dev.netlify.app/

@wlach mentioned in [#glean-dictionary](https://chat.mozilla.org/#/room/#glean-dictionary:mozilla.org) that we should move to using [cards](https://protocol.mozilla.org/patterns/molecules/card.html) to render this data. 

Screen Shot 2021-01-13 at 6 30 01 PM

Iinh:
Render the application list with Protocol's [cards](https://protocol.mozilla.org/patterns/molecules/card.html) to fix #318. While I was working on this I also restyled the metadata table a little bit.

### Pull Request checklist

      fixes, if applicable)


src/pages/AppList.svelte
```diff
@@ -19,44 +23,159 @@
...
+
+  .app-list {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: center;
+
+    .mzp-c-card {
+      flex-basis: 100%;
+      margin: 3px;
+      padding: 3px;
```
wlach, line 108: Some of this might not be necessary if you use CSS grid, but regardless I would use protocol tokens for margin/spacing.

src/pages/AppList.svelte
```diff
@@ -19,44 +23,159 @@
...
+    margin: $spacing-md $spacing-xl;
+    .deprecation-checkbox {
+      text-align: right;
     }
-    i {
-      @include text-body-md;
+  }
+
+  .app-list {
+    display: flex;
```
wlach, line 101: I think we should use [CSS grid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Basic_Concepts_of_Grid_Layout) for this, it's kind of exactly what it's designed for (flex box is a little more low-level).

A few extra points on what this should look like:

* I think we should try to have more padding between the cards (at least 24px, I think -- don't know what the equivalent would be in protocol tokens)
* I think the model of having a variable number of columns depending on display size is the right one, but I think extra elements (if there is a trailing row) should be *left* aligned, rather than centred (I think you should be able to do this with [grid-auto-flow](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-auto-flow).

src/pages/AppList.svelte
```diff
@@ -19,44 +23,159 @@
...
+    if (includes(app, "iOS")) {
+      return "https://i.imgur.com/swlzpbM.jpg";
+    }
+    if (includes(app, "Android")) {
+      return "https://i.imgur.com/ji8NP0b.png";
+    }
+    if (includes(app, "Amazon")) {
+      return "https://i.imgur.com/EwaktM3.png";
+    }
+    return null;
```
wlach, line 88: `undefined` is a bit more idiomatic in JS-land.

```suggestion
    return undefined;
```

wlach (changes requested):
This looks good! Could you move the table changes to another PR? I'd like to review/land them seperately. In general it's best to just have one PR per issue / idea (I sometimes make  exceptions for this when making very small cleanups/improvements)

Iinh:
Keeping the FilterInput styling in this PR because I only restyled it so that it would fit better with the AppList page, hope that's okay.

src/components/FilterInput.svelte
```diff
@@ -16,14 +18,17 @@
...
+    cursor: pointer;
+    box-shadow: $box-shadow-sm;
   }
 </style>
 
 <div>
   <input
-    type="text"
+    placeHolder={text}
+    style="border-radius: 15px; border: 1px solid #000;"
```
wlach, line 29: This should be in the css class above, no?

src/components/FilterInput.svelte
```diff
@@ -4,6 +4,8 @@
   export let onChangeText;
   export let filterText = "";
 
+  export let text;
```
wlach, line 7: I think placeHolder might be a better name here?

```suggestion
  export let placeHolder;
```

src/pages/AppDetail.svelte
```diff
@@ -19,49 +21,53 @@
...
 </style>
 
-{#await appDataPromise then app}
-  {#if app.prototype}
-    <AppAlert
-      status="warning"
-      message="This application is a prototype. The metrics and pings listed below may contain inconsistencies and testing strings." />
-  {/if}
-  <PageTitle text={params.app} />
+<div transition:fade>
```
wlach, line 24: I'd suggest figuring out transitions in a seperate PR. I totally agree we should do them, but I'd like to approach them deliberately.

src/pages/AppList.svelte
```diff
@@ -19,44 +23,150 @@
...
-        <a href="/apps/{app.name}">{app.name}</a>
-        {#if app.description}<i>{app.description}</i>{/if}
-        {#if app.deprecated}
-          <Pill message="Deprecated" bgColor="#4a5568" />
+    <div class="app-list">
+      {#each filteredApps as app}
+        {#if showDeprecated || !app.deprecated}
+          <div
+            class="mzp-c-card mzp-c-card-extra-small has-aspect-3-2"
+            style="max-width: 230px;">
```
wlach, line 139: This property should probably be in a class.

src/pages/AppList.svelte
```diff
@@ -19,44 +23,150 @@
...
+            <a class="mzp-c-card-block-link" href="/apps/{app.name}">
+              <div class="mzp-c-card-media-wrapper" style="border-radius: 4px;">
+                <img
+                  class="mzp-c-card-imgage"
+                  src={getAppLogo(app.name)}
+                  alt="${app.name} Logo"
+                  style="width: 230px;" />
```
wlach, line 146: Likewise, should be in a class.

src/pages/AppList.svelte
```diff
@@ -19,44 +23,150 @@
...
+    position: absolute;
+    top: 7px;
+    left: 7px;
+    width: 30px;
+    background-color: inherit;
   }
 </style>
 
-<PageTitle text="Applications" />
+<div transition:fade>
```
wlach, line 122: Likewise, let's do transitions in a seperate PR.

wlach (changes requested):
This looks good. Makes sense to me to change the input styling as part of this PR, but let's skip the transition stuff for now-- I'd like to look at animations/transitions holistically.

src/components/FilterInput.svelte
Iinh: The reason for inline styling was because I wanted to override Protocol rules in class selectors: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity, but I realized just now that I can also achieve this by giving them an id

wlach: I think in this case we should use `!important` to override the default behaviour. I know this is normally discouraged but I think we're following the rules of thumb outlined here:

https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception

wlach (changes requested):
Looks good except for the id thing, let me know what you think:

https://github.com/mozilla/glean-dictionary/pull/333#discussion_r560416771

src/components/FilterInput.svelte
Iinh: I'm having reservations about this, mostly because we need to override Protocol quite often (due to it being compiled last in the process hence always has higher priority over Svelte component css), and using `!important` across the the stylesheet like that in the long term would make it harder to debug.

> I know this is normally discouraged but I think we're following the rules of thumb outlined here:
> 
> [https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception)

Rule of thumb number 1 is "Always look for a way to use specificity before even considering !important", and since id selector is included in the docs as one of selector types recommended when we need to use specificity: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#selector_types and https://developer.mozilla.org/en-US/docs/Web/CSS/ID_selectors, I don't think we're straying from convention by using ids. I don't have a strong attachment to this though, let me know what you think.

src/components/FilterInput.svelte
```diff
@@ -14,16 +16,20 @@
     margin-top: $spacing-xs;
     margin-bottom: $spacing-xs;
   }
-  input {
+  #filter-input {
     width: 100%;
+    cursor: pointer;
```
wlach, line 21: Just noticed this. I think we should actually not do this, it's more common for search fields to just use the default pointer type.

```suggestion
```

src/components/FilterInput.svelte
wlach: Here's a resource which suggests that id's should be unique (which more or less corresponds to what I've seen in the wild): https://css-tricks.com/the-difference-between-id-and-class/#ids-are-unique

*However*, I agree that using ids is probably preferably to `!important` for the reasons you mention. I assume that option (2) under the [important exception](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception) ("use more specific rules") wouldn't work for some reason?

wlach (approved):
Going to r+ this, minus one last nit about pointer type that I noticed (see comments). Also, if you can find a way to not use id's that would be best, but I can live with it otherwise.

This looks so good BTW! Can't wait to show it to others.

src/components/FilterInput.svelte
Iinh: > However, I agree that using ids is probably preferably to !important for the reasons you mention. I assume that option (2) under the important exception ("use more specific rules") wouldn't work for some reason?

Yes I tried it, but even with more specific targeting, it would still end up being overridden by Protocol (if limited to type and class selectors) because Protocol rules are compiled last so they show up last on the style sheet. The only way to counter this is either using ids or `!important`. What I'm worried about `!important` is that since the need to customize styling is quite frequent, and if we decide to use `!important` pattern whenever we need to customize, it can get messy really fast. If everything is important, what's actually important?

> Here's a resource which suggests that id's should be unique (which more or less corresponds to what I've seen in the wild): https://css-tricks.com/the-difference-between-id-and-class/#ids-are-unique

I will make sure that the ids are a) unique (which is usually the case because when I want to override Protocol I aim at a very specific element) and b) used sparingly, only when absolutely needed.

wlach:
@Iinh Could you fix the merge conflicts then land when you have a chance? Thanks again for doing this!

PR #338, opened 2021-01-19
Fix footer jumping around (fixes #322)

### LINKED ISSUE #322: Fix footer to the bottom of the page
Right now the footer is not sticky, it jumps up when the content length changes, and leaves a white gap at the bottom.

Screen Shot 2021-01-14 at 10 34 40 AM

Example:
https://glean-dictionary.protosaur.dev/apps/fenix/metrics/about_page-libraries_tapped

Iinh:
(sort of) Fixes #322.

This seemingly simple problem turned out to be trickier than I expected. The proposed solution in this PR doesn't _completely_ fix the problem: it worked fine on my laptop screen, but when I tested on a bigger screen (32") and extended the browser window height the problem still persisted.

**On laptop screen**

Screen Shot 2021-01-19 at 2 28 47 PM

**On bigger screen**

gd

I tried various solutions, and the ones that sort of worked were all a variation of [what @wlach suggested in the original issue](https://stackoverflow.com/a/21207313). I don't have much experience with this, but it seems like even GitHub has this "problem" too, so maybe we're not looking at a problem, but more like an expected footer behavior when the content height is shorter? It's just more prominent on our site because the background color of our footer is black? 

github

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -94,12 +94,11 @@
 </script>
 
 <style>
-  html,
-  body {
-    /* keep the  below value the same as footer height 
-       to prevent footer from jumping around */
-    /* stackoverflow.com/questions/21206058/place-footer-at-bottom-only-if-page-is-short/21207313#21207313 */
-    padding-bottom: 350px;
+  .app {
+    min-height: 100vh;
+    display: grid;
+    grid-template-rows: auto 1fr auto;
+    grid-template-columns: 100%;
   }
 
   .mzp-c-navigation {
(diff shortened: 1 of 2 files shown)


wlach (changes requested):
I think maybe I gave bad advice (sorry). I think this is actually a good problem to solve with CSS grid (or Flexbox). Could you give this a try? If it works (95% certain it will), we can avoid hardcoding pixel values in the CSS.

https://css-tricks.com/couple-takes-sticky-footer/#there-is-grid

Iinh:
It worked! This solution is simple and elegant too, I love CSS Grid!

wlach (approved):
💯

PR #339, opened 2021-01-22
Add Searchfox link for metrics (fixes #65)

### LINKED ISSUE #65: Add searchfox links for metrics
It would be super handy to provide searchfox search URLs that search the code base for uses of a given metric.  It should be enough to search for `category_name.metric_name` as a string.

Depends on https://bugzilla.mozilla.org/show_bug.cgi?id=1668547

Iinh:
Fixes #65.

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/formatters/searchfox.js
+++ b/src/formatters/searchfox.js
@@ -1,3 +1,5 @@
+import { camelCase, upperFirst, snakeCase } from "lodash";
+
 const searchfoxMap = {
   fenix: "fenix",
 
@@ -26,14 +28,38 @@ const searchfoxMap = {
   "reference-browser": "reference-browser",
 };
 
-const MozillaCentralURL = "https://searchfox.org/mozilla-central/";
-const MozillaMobileURL = "https://searchfox.org/mozilla-mobile/";
-
 export const getSearchfoxLink = (app, metric) => {
+  const [category, name] = metric.split(/\.(?=[^.]+$)/);
+
+  // Generate all possible casings for a Searchfox query:
+  // https://github.com/mozilla/glean-dictionary/pull/339#issuecomment-766904256
+
+  // Javascript: camelCase.camelCase
+  const camelCased = name
+    ? `${camelCase(category)}.${camelCase(name)}`
+    : camelCase(category);
+
+  // Kotlin, Swift: CamelCase.camelCase
+  const capitalizedCamelCased = name
+    ? upperFirst(`${camelCase(category)}.${camelCase(name)}`)
+    : camelCase(category);
+
+  // Python: snake_case.snake_case
+  const snakedCased = name
+    ? `${snakeCase(category)}.${snakeCase(name)}`
+    : snakeCase(category);
+
+  // Rust, C++: snake_case::snake_case
+  const dblColonSnakeCased = name
+    ? `${snakeCase(category)}::${snakeCase(name)}`
+    : snakeCase(category);
+
+  const allLanguagePatterns = `${camelCased}|${capitalizedCamelCased}|${snakedCased}|${dblColonSnakeCased}`;
+
   if (searchfoxMap[app]) {
     return app === "firefox-desktop"
-      ? `${MozillaCentralURL}search?q=${metric}`
-      : `${MozillaMobileURL}search?q=${metric}&path=${searchfoxMap[app]}`;
+      ? `https://searchfox.org/mozilla-central/search?q=${allLanguagePatterns}&regexp=true`
+      : `https://searchfox.org/mozilla-mobile/search?q=${allLanguagePatterns}&path=${searchfoxMap[app]}&regexp=true`;
   }
-  return false;
+  return undefined;
 };
--- a/src/formatters/searchfox.js
+++ b/src/formatters/searchfox.js
@@ -1,0 +1,39 @@
+const searchfoxMap = {
+  fenix: "fenix",
+
+  "firefox-android-beta": "fenix",
+
+  "firefox-android-release": "fenix",
+
+  "firefox-desktop": "https://searchfox.org/mozilla-central/source/",
+
+  "firefox-focus-ios": "focus-ios",
+
+  "firefox-for-echo-show": "firefox-echo-show",
+
+  "firefox-for-fire-tv": "firefox-tv",
+
+  "firefox-ios-beta": "firefox-ios",
+
+  "firefox-ios-dev": "firefox-ios",
+
+  "firefox-ios-release": "firefox-ios",
+
+  "firefox-reality": "FirefoxReality",
+
+  "lockwise-android": "lockbox-android",
+
+  "reference-browser": "reference-browser",
+};
+
+const MozillaCentralURL = "https://searchfox.org/mozilla-central/";
+const MozillaMobileURL = "https://searchfox.org/mozilla-mobile/";
+
+export const getSearchfoxLink = (app, metric) => {
+  if (searchfoxMap[app]) {
+    return app === "firefox-desktop"
+      ? `${MozillaCentralURL}search?q=${metric}`
+      : `${MozillaMobileURL}search?q=${metric}&path=${searchfoxMap[app]}`;
+  }
+  return false;
+};
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -7,6 +7,8 @@
   import HelpHoverable from "../components/HelpHoverable.svelte";
   import helpText from "../data/help";
   import PageTitle from "../components/PageTitle.svelte";
+
+  import { getSearchfoxLink } from "../formatters/searchfox";
   import { pageTitle } from "../state/stores";
 
   export let params;
@@ -204,6 +206,14 @@
       <td>Version</td>
       <td>{metric.version || 0}</td>
     </tr>
+    {#if getSearchfoxLink(params.app, metric.name)}
+      <tr>
+        <td>Searchfox</td>
+        <td>
+          <a href={getSearchfoxLink(params.app, metric.name)}>{metric.name}</a>
+        </td>
+      </tr>
+    {/if}
     <tr>
       <td>BigQuery</td>
       <td>
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -208,9 +208,15 @@
     </tr>
     {#if getSearchfoxLink(params.app, metric.name)}
       <tr>
-        <td>Searchfox</td>
         <td>
-          <a href={getSearchfoxLink(params.app, metric.name)}>{metric.name}</a>
+          Searchfox
+          <HelpHoverable
+            content={helpText.searchFox.text}
+            link={helpText.searchFox.link} />
+        </td>
+        <td>
+          <a href={getSearchfoxLink(params.app, metric.name)}>
+            <code>{metric.name}</code></a>
         </td>
       </tr>
     {/if}
--- a/src/data/help.js
+++ b/src/data/help.js
@@ -18,4 +18,9 @@ export default {
     text: "Defines which pings the metric is sent in.",
     link: OPTIONAL_METRIC_PARAMS_DOCS,
   },
+  searchFox: {
+    text:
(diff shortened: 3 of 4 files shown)


src/components/SearchfoxLink.svelte
```diff
@@ -0,0 +1,69 @@
...
+
+    "lockwise-android": "lockbox-android", // (?)
+
+    "lockwise-ios": "", // (?)
+
+    mach: "", // (?)
+
+    mozphab: "", // (?)
+
+    mozregression: "", // (?)
```
Iinh, line 51: I was not able to find some applications in Searchfox. Could you help point me to where I might be able to find these repos @wlach? Thanks!

wlach: Right. Someone needs to explicitly add a repo to be indexed to searchfox. I think we can just skip these repos for now, where they're not indexed -- we can always add them at a later time, and update this list. The major ones (e.g. Firefox for Android) are.

src/data/searchfox.js
```diff
@@ -0,0 +1,47 @@
+export const searchfoxMap = {
+  burnham: "",
```
wlach, line 2: I would just omit entries from the map that aren't in searchfox.

src/pages/MetricDetail.svelte
```diff
@@ -204,6 +207,14 @@
       <td>Version</td>
       <td>{metric.version || 0}</td>
     </tr>
+    {#if searchfoxMap[params.app]}
+      <tr>
+        <td>Find in</td>
```
wlach, line 212: Let's just use "Searchfox" here. It's on me to add some help text on what this means. 😄 

```suggestion
        <td>Searchfox</td>
```

src/pages/MetricDetail.svelte
```diff
@@ -204,6 +207,14 @@
...
+        <td>
+          <SearchfoxLink metricName={metric.name} appName={params.app} />
```
wlach, line 214: Creating a component here seems like overkill to me. How about just extracting out the getSearchFoxLink method in the component and put into `src/formatters/searchfox.js`? Bonus: in that case you can skip importing `searchFoxMap` in this file, and just check if `getSearchFoxLink` returns a value.

wlach (changes requested):
The ideas are right, but I think we can simplify this a bit.

src/pages/MetricDetail.svelte
```diff
@@ -204,6 +206,14 @@
       <td>Version</td>
       <td>{metric.version || 0}</td>
     </tr>
+    {#if getSearchfoxLink(params.app, metric.name)}
+      <tr>
+        <td>Searchfox</td>
+        <td>
+          <a href={getSearchfoxLink(params.app, metric.name)}>{metric.name}</a>
+        </td>
```
wlach, line 214: I feel like maybe we should add some kind of warning if the metric is expired here, but we can do that in a followup.

wlach (changes requested):
Some minor nits, but LGTM otherwise. The only other thing that concerns me a bit is that I'm actually not seeing many references to the metrics in the actual repositories. e.g. for for this fenix metric:

https://deploy-preview-339--glean-dictionary-dev.netlify.app/apps/fenix/metrics/activation-activation_id

We only get one result from the URL, a link to the generated markdown documentation:

https://searchfox.org/mozilla-mobile/search?q=activation.activation_id&path=fenix

In #65, @mdboom linked to https://bugzilla.mozilla.org/show_bug.cgi?id=1668547, but that landed quite a while ago so am a bit surprised not to see anything more come up for fenix. Is there more we need to do here?

@badboy @mdboom can you give more insight on this? Also I'd appreciate any thoughts you might have on the UI parts of this, I spend a lot less time in searchfox than either of you. 😄

mdboom:
> @badboy @mdboom can you give more insight on this? Also I'd appreciate any thoughts you might have on the UI parts of this, I spend a lot less time in searchfox than either of you. smile

I think the crux of the problem is that the metric names are always in `snake_case` in the `metrics.yaml` file, but they may have different casing depending on the conventions of the various language.  For example, in Kotlin, members are in `camelCase`, so you'd have to search for [activation.activationId](https://searchfox.org/mozilla-mobile/search?q=activation.activationId&path=&case=false&regexp=false).  I think for the 80% solution, we'd need to generate a searchfox query with all of the possible casings, which are (where the first part is the metric category, and the second part the name):

- Kotlin & Swift: `CamelCase.camelCase`
- Python: `snake_case.snake_case`
- Rust: `snake_case::snake_case`

It's my bad thinking that https://bugzilla.mozilla.org/show_bug.cgi?id=1668547 would be helpful here -- that only affects generated code, which isn't available in searchfox.

src/formatters/searchfox.js
```diff
@@ -0,0 +1,39 @@
...
+  "firefox-ios-dev": "firefox-ios",
+
+  "firefox-ios-release": "firefox-ios",
+
+  "firefox-reality": "FirefoxReality",
+
+  "lockwise-android": "lockbox-android",
+
+  "reference-browser": "reference-browser",
+};
```
mdboom, line 27: I wonder if (longer term -- maybe as a follow-on to this work) this should be in probe-scraper's repositories.yaml instead.

src/pages/MetricDetail.svelte
mdboom: :+1:

wlach:
> I think the crux of the problem is that the metric names are always in `snake_case` in the `metrics.yaml` file, but they may have different casing depending on the conventions of the various language. For example, in Kotlin, members are in `camelCase`, so you'd have to search for [activation.activationId](https://searchfox.org/mozilla-mobile/search?q=activation.activationId&path=&case=false&regexp=false). I think for the 80% solution, we'd need to generate a searchfox query with all of the possible casings, which are (where the first part is the metric category, and the second part the name):
> 
>     * Kotlin & Swift: `CamelCase.camelCase`
>     * Python: `snake_case.snake_case`
>     * Rust: `snake_case::snake_case`

Got it, I think this should not be too bad. @Iinh do you want to take a crack at this? Lodash's [capitalize](https://lodash.com/docs/4.17.15#capitalize) and [snakeCase](https://lodash.com/docs/4.17.15#snakeCase) functions may be useful here.

src/formatters/searchfox.js
Iinh: 👍

src/pages/MetricDetail.svelte
Iinh: I'll add the warning in a followup PR after #341 has landed

Iinh:
@wlach I added all the possible casings as @mdboom suggested, however I'm not sure if my approach to the Searchfox UI is correct. Guidance appreciated!

https://deploy-preview-341--glean-dictionary-dev.netlify.app/apps/fenix/metrics/activation_activation_id

Screen Shot 2021-01-25 at 11 49 16 AM

mdboom:
> @wlach I added all the possible casings as @mdboom suggested, however I'm not sure if my approach to the Searchfox UI is correct. Guidance appreciated!
> 
> https://deploy-preview-341--glean-dictionary-dev.netlify.app/apps/fenix/metrics/activation_activation_id
> 
> Screen Shot 2021-01-25 at 11 49 16 AM

Personally, I'd prefer a single link that shows results for all of the languages (if that's possible with a searchfox query).  I don't think most users of this will know or care which language the telemetry is implemented in.  But that's just a hunch -- @chutten thoughts?

Also, it occurs to me that we'll need to support C++ and Javascript for Firefox desktop as well.  @chutten: would you mind providing a description of what those would look like [as in my comment above?](https://github.com/mozilla/glean-dictionary/pull/339#issuecomment-766904256).

Lastly, it would be great to have one of those informational tooltips on this field that says:

```
Finds uses of this metric using simple matching.
Metrics used dynamically may not appear in the results.
```

wlach:
> Personally, I'd prefer a single link that shows results for all of the languages (if that's possible with a searchfox query). I don't think most users of this will know or care which language the telemetry is implemented in. But that's just a hunch -- @chutten thoughts?

searchfox supports regex queries, so we should be able to do this as follows:

https://searchfox.org/mozilla-mobile/search?q=(Activation.activationId|activation.activation_id)&path=fenix&regexp=true

> Lastly, it would be great to have one of those informational tooltips on this field that says:
> 
> ```
> Finds uses of this metric using simple matching.
> Metrics used dynamically may not appear in the results.
> ```

👍

chutten:
C++ uses `snake_case::snake_case`
JS uses `camelCase.camelCase`

As for whether users would care what language it would be written in... definitely not (for Firefox Desktop). Firefox is written in JS and C++ and Rust and which language a probe is written in might be surprising if you don't understand the underlying patterns of which languages are commonly used for which subsystems. And then for different probes its language would only be surprising if you _did_ understand the patterns : |

Iinh:
I added all the possible casings to Searchfox search query (Javascript, C++, Rust for `firefox-desktop`, and Kotlin, Swift, Python, Rust for mobile applications), thanks @mdboom, @wlach  and @chutten! However I notice that the search query for Glean metrics in [firefox-desktop](https://deploy-preview-339--glean-dictionary-dev.netlify.app/apps/firefox-desktop/) doesn't yield any result, e.g: 

https://deploy-preview-339--glean-dictionary-dev.netlify.app/apps/firefox-desktop/metrics/glean-error-io, or
https://deploy-preview-339--glean-dictionary-dev.netlify.app/apps/firefox-desktop/metrics/glean-error-preinit_tasks_overflow

Am I missing something?

mdboom:
> I added all the possible casings to Searchfox search query (Javascript, C++, Rust for `firefox-desktop`, and Kotlin, Swift, Python, Rust for mobile applications), thanks @mdboom, @wlach and @chutten! However I notice that the search query for Glean metrics in [firefox-desktop](https://deploy-preview-339--glean-dictionary-dev.netlify.app/apps/firefox-desktop/) doesn't yield any result, e.g:
> 
> https://deploy-preview-339--glean-dictionary-dev.netlify.app/apps/firefox-desktop/metrics/glean-error-io, or
> https://deploy-preview-339--glean-dictionary-dev.netlify.app/apps/firefox-desktop/metrics/glean-error-preinit_tasks_overflow
> 
> Am I missing something?

In Glean, the app itself can define metrics, but also any libraries that the app uses (including Glean itself). In this case, these metrics are defined in Glean itself, so we'd need to search in Glean's github repo in searchfox.

As a first pass (to merge this PR), I'd say it's fine to only search the app for metrics (that's the common case).  As an additional hint, we could add to the tooltip "Only metrics defined in the application will be found."

To improve this, though, we'd need to search in the library the metric is defined in, not the app.  Does Glean dictionary have that information?  It's certainly known as part of the inputs to probe-scraper, but I'd have to check whether that's preserved in the output of probe-scraper...

wlach:
> To improve this, though, we'd need to search in the library the metric is defined in, not the app. Does Glean dictionary have that information? It's certainly known as part of the inputs to probe-scraper, but I'd have to check whether that's preserved in the output of probe-scraper...

I believe it is not currently, but it's on my agenda to add this to a next-generation `/v2/` metrics API (proposal forthcoming).

mdboom:
So, it seems like:

- For now, let's ignore the fact it can't find library metrics, and document that shortcoming in the tooltip.  We can file a bug and revisit this once the v2 probe-scraper format ships.
- Also, because of this multi-library thing, let's just search for all language patterns all the time (no reason to have different behavior for desktop). For example, even for Fenix (which is written in Kotlin), it uses libraries that are written in Rust.  I don't think there's a good reason to not just show all language patterns all the time.

mdboom (approved):
Looks good!  Let's not forget to file a follow-on issue to correctly support libraries when the other requirements for that exist.

wlach (approved):
💯

src/pages/MetricDetail.svelte
Iinh: Should I still add a warning here @wlach, or do you think the expired Pill is sufficient enough?

wlach: I think it would be worth adding a quick warning. Doesn't hurt.

PR #341, opened 2021-01-22
Hide expired metrics by default (fixes #23)

### LINKED ISSUE #23: Don't show expired metrics by default
Many of the probes in the existing probe dictionary are expired or deprecated (and they don't always have build end dates).  This causes confusion, since it might not be clear that no data will be available in those probes for currently-released products.  While the ability to access historical probes should always exist, we should optimize for what is presumably the common case of looking at new data flowing in.

This might include:

- A visually obvious identifier for deprecated or expired probes
- Not returning search results for expired probes (unless checking a box)
- Ranking search results by date of last update to the probe definition

Iinh:
Fixes #23.

- Provide a "Show expired metrics" checkbox under "Metrics" tab that lets users see expired metrics.
- Add an "Expired" pill to metrics that are expired in Metric Page.

### Pull Request checklist

      fixes, if applicable)


src/components/ItemList.svelte
```diff
@@ -18,22 +20,41 @@
...
-  $: filteredItems = paginatedItems[$currentPage - 1];
+  function handleExpiredItems(checked) {
+    if (checked) {
+      shownItems = items;
+      filterItems(filterTerm);
+    } else {
+      shownItems = nonExpiredItems;
+      filterItems(filterTerm);
+    }
+  }
```
wlach, line 53: You could simplify this to:

```suggestion
   shownItems = showExpired ? items : nonExpiredItems;
   filterItems(filterTerm);
```

src/components/ItemList.svelte
```diff
@@ -18,22 +20,41 @@
...
+    filteredItems = shownItems.filter((item) => item.name.includes(filterText));
     if (filteredItems.length > 0) {
       currentPage.set(1);
       paginatedItems = chunk([...filteredItems], DEFAULT_ITEMS_PER_PAGE);
       totalItems = filteredItems.length;
     }
   }
 
-  $: filteredItems = paginatedItems[$currentPage - 1];
+  function handleExpiredItems(checked) {
```
wlach, line 45: Should use a more descriptive variable name here:

```suggestion
  function handleExpiredItems(showExpired) {
```

I feel like we should somehow combine this function with filterItems above, to generate the list of shownItems. Is that possible? It would be nice if we could simplify this component a bit.

src/utils/index.js
```diff
@@ -0,0 +1,11 @@
+export const isExpired = (date) => {
```
wlach, line 1: I try to avoid creating utils functions, see [this blog post](https://lostechies.com/chrismissal/2009/06/01/anti-patterns-and-worst-practices-utils-class/) for a set of good arguments on why.

Could you define this function in something like `src/state/metrics.js`? It seems to perform a similar task to `urls.js` in that folder.

src/utils/index.js
```diff
@@ -0,0 +1,11 @@
...
+  if (date === "never" || date === undefined) {
+    return false;
+  }
+  const GivenDate = new Date(date);
+  const CurrentDate = new Date();
+  if (GivenDate > CurrentDate) {
+    return false;
+  }
+  return true;
```
wlach, line 10: I think you could probably turn this into a one-liner.

```suggestion
  return new Date() > new Date(expiryDate)
```

src/utils/index.js
```diff
@@ -0,0 +1,11 @@
+export const isExpired = (date) => {
+  if (date === "never" || date === undefined) {
+    return false;
+  }
```
wlach, line 4: It's a little clearer if you name the variable "expiryDate".

```suggestion
export const isExpired = (expiryDate) => {
  if (expiryDate === "never" || expiryDate === undefined) {
    return false;
  }
```

wlach (changes requested):
A few nits, but this is on the right track!

src/components/ItemList.svelte
```diff
@@ -18,14 +20,22 @@
...
-  paginatedItems = chunk([...items], DEFAULT_ITEMS_PER_PAGE);
+  totalItems = shownItems.length;
+  paginatedItems = chunk([...shownItems], DEFAULT_ITEMS_PER_PAGE);
 
-  function filterItems(filterText) {
-    filteredItems = items.filter((item) => item.name.includes(filterText));
+  function filterItems(filterText, expiryCheck) {
+    shownItems = expiryCheck ? items : nonExpiredItems;
+    filterTerm = filterText;
+    filteredItems = shownItems.filter((item) => item.name.includes(filterText));
```
wlach, line 38: I'm not sure about this function-- it's modifying global state, yet also takes parameters. I think in general its best if functions [don't use/modify global state](https://thevaluable.dev/global-variable-explained/) -- but I case see in this case that might be a bit complicated to refactor. If we're going to leave it as-is, I'd take out the function parameters, which are basically shadowing global state already.

```suggestion
  function filterItems() {
    shownItems = expiryCheck ? items : nonExpiredItems;
    filterTerm = filterText;
    filteredItems = shownItems.filter((item) => item.name.includes(filterText));
```

src/components/ItemList.svelte
```diff
@@ -34,6 +44,7 @@
   }
 
   $: filteredItems = paginatedItems[$currentPage - 1];
+  $: filterItems(filterTerm, showExpired);
```
wlach, line 47:
```suggestion
  $: filteredItems = paginatedItems[$currentPage - 1];
  $: filterItems(filterTerm, showExpired);
```

Do we need both of these? Could we just move `filteredItems = paginatedItems[$currentPage - 1];` into the filterItems function above?

wlach (changes requested):
Getting close, I think there's still a few more things that I'd like to simplify/change.

src/utils/index.js
Iinh: I did not know util functions were considered an anti pattern. Thank you for this great tip!

src/components/ItemList.svelte
```diff
@@ -9,31 +9,45 @@
...
+  let showExpired = false;
+  let filteredItems = items.filter((item) => !isExpired(item.expires));
+  let pagedItems;
 
   let currentPage = writable(1);
   setContext("currentPage", currentPage);
 
-  totalItems = items.length;
-  paginatedItems = chunk([...items], DEFAULT_ITEMS_PER_PAGE);
+  const goToPage = (page) => {
```
Iinh, line 29: I took out the pagination logic from the old filter function and put in a separate function to make the code a bit easier to read, hopefully…

src/utils/index.js
wlach: Opinions on this topic do vary a bit (you will see `utils` functions in lots of packages and places), but I think the argument I presented above is pretty solid. 😄

src/components/ItemList.svelte
```diff
@@ -9,31 +9,45 @@
...
-    }
-  }
+  const handleFilter = () => {
+    let shownItems = showExpired
+      ? items
+      : items.filter((item) => !isExpired(item.expires));
+    filteredItems = shownItems.filter((item) => item.name.includes(filterText));
+    // show the first page of result
+    currentPage.set(1);
+    goToPage(1);
```
wlach, line 43: Is this necessary? I feel like if we set the store we should get the callback automatically below.

src/components/ItemList.svelte
```diff
@@ -96,10 +128,11 @@
         </tr>
       </thead>
       <tbody>
-        {#each filteredItems as item}
+        {#each pagedItems as item}
           <tr>
             <td>
               <a href={getItemURL(appName, itemType, item.name)}>{item.name}</a>
+              {#if isExpired(item.expires)}<i>(expired)</i>{/if}
```
wlach, line 135: Just curious, what was the rationale for this style as opposed to the pills? Do you think it looks better?

wlach (changes requested):
Just a few very small items.

src/components/ItemList.svelte
Iinh: if `currentPage` is already 1, setting it to 1 would not call `goToPage()`

wlach: Makes sense, maybe we could just add a quick comment about that?

src/components/ItemList.svelte
Iinh: I wanted to keep the table items looking minimal, but now that I think about it more I don't think using Pills would clutter it too much. Do you think using Pills might be better? 

Screen Shot 2021-01-26 at 7 59 02 AM
Screen Shot 2021-01-26 at 7 59 45 AM

wlach: I'd probably vote in favour of pills, just to be consistent with the rest of the application.

Iinh: 👍

src/components/ItemList.svelte
```diff
@@ -8,32 +8,48 @@
...
-    }
-  }
+  const handleFilter = () => {
+    const shownItems = showExpired
+      ? items
+      : items.filter((item) => !isExpired(item.expires));
+    filteredItems = shownItems.filter((item) => item.name.includes(filterText));
+    // show the first page of result
+    currentPage.set(1);
+    // if currentPage is already 1, we need to manually call goToPage() to get the first page
```
wlach, line 44:
```suggestion
    // even if currentPage is already 1, we need to manually call goToPage() to get the first page
```

wlach:
Thanks so much for this @Iinh, this will make the dictionary so much more useful.

PR #346, opened 2021-01-24
Use `_` instead of `-` to identify metric names (fixes #313)

### LINKED ISSUE #313: Backlinks from BigQuery view -> Metric don't work
In the BigQuery view, we currently have a backlink to the metric that generated it.

image

But the link is often wrong. For example for:

https://glean-dictionary-dev.netlify.app/apps/firefox-android-release/tables/metrics?search=metrics.string_list.addons_enabled_addons

The link should be:

https://glean-dictionary-dev.netlify.app/apps/firefox-android-release/metrics/addons-enabled_addons

But actually is:

https://glean-dictionary-dev.netlify.app/apps/firefox-android-release/metrics/addons_enabled_addons

Fenn-CS:
Replaced `.`s in metric ulrs with `_`s as a way to establish consistency with the metric names in BigQuery definitions.

Fixes : #313

### DIFF
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -10,11 +10,10 @@
   import { pageTitle } from "../state/stores";
 
   export let params;
-  let metricName = params.metric;
 
-  const metricDataPromise = getMetricData(params.app, metricName);
+  const metricDataPromise = getMetricData(params.app, params.metric);
 
-  pageTitle.set(`${metricName} | ${params.app} `);
+  pageTitle.set(`${params.metric} | ${params.app} `);
 
   function getGlamUrlTemplate(app) {
     const map = {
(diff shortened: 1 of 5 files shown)


src/pages/MetricDetail.svelte
```diff
@@ -10,7 +10,7 @@
   import { pageTitle } from "../state/stores";
 
   export let params;
-  let metricName = params.metric.replaceAll("-", ".");
+  let metricName = params.metric;
```
wlach, line 13: I think we should just take out this variable and replace it with "params.metric" everywhere. I think `metricName` falsely implies that this is the real name of the metric, rather than an identifier.

wlach (changes requested):
There's a minor unit test failure (see the circle ci logs) and I have one minor requested change, other than that, looks great!

src/pages/MetricDetail.svelte
Fenn-CS: Yep! Makes sense.

wlach (commented):
👍

PR #355, opened 2021-01-25
Add some extra stuff about Data @ Mozilla to footer

wlach:
Add links to TMO, Data Blog, Data @ Mozilla vision statement in advance of public launch. Still need to figure out if this is what we want.

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/Footer.svelte
+++ b/src/components/Footer.svelte
@@ -34,6 +34,15 @@
           <a
             href="https://github.com/mozilla/glean-dictionary/tree/{rev}">{rev.substring(0, 10)}</a>
         </p>
+        <p>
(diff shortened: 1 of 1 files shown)


mcfunley (commented):
👍

Iinh (approved):
Looks good, although I do wonder if we should keep this, along with more info about Glean Dictionary, in a short About Page.

wlach:
> Looks good, although I do wonder if we should keep this, along with more info about Glean Dictionary, in a short About Page.

Seems like a good idea, but I think it might make sense to wait a bit on that. I think we'll learn a lot more about common questions people want answered, etc. after this product a bit more use in the field.

PR #357, opened 2021-01-27
Add more help to metric and ping pages (fixes #323)

### LINKED ISSUE #323: Help strings everywhere
We have the capability of adding help hovers since #300, but are only using them sporadically. We should use them all over the UI to explain what the various fields mean (and provide links to documentation where possible). I'm going to work on this myself, since it requires a fair amount of domain knowledge. Will get the glean team to review.

wlach:
This is not ready for review, but just putting it up so people (esp. @Iinh) are aware of it.

### Pull Request checklist

      fixes, if applicable)


wlach:
Apologies for mixing a bunch of stuff in this PR (in particular a move away from the sveltejs-tippy library we were using earlier, which has a weird interaction with storybook), but I think it's easier to just tackle all of this at once. Basically this revamps our metadata display, adding more help everywhere and moving the definition of what we display (for apps, metrics, and pings) to a schema file. Looking to @Iinh to review the code/display logic and @mdboom / @badboy to validate look at the help text (but anyone should feel free to critique the latter). You can view a deploy preview here:

https://deploy-preview-357--glean-dictionary-dev.netlify.app/

src/pages/MetricDetail.svelte
```diff
@@ -92,6 +91,9 @@
 <style>
   @import "../main.scss";
   @include metadata-table;
+  h2 {
```
wlach, line 94: @Iinh I can't remember if we figured out a way of setting this globally? It would be nice to use a universal value here.

src/components/MetadataTable.svelte
```diff
@@ -0,0 +1,48 @@
...
+          {/if}
+        </td>
+        <td>
+          {#if schemaEntry.type === 'link'}
+            <a
+              href={format(item[schemaEntry.id], schemaEntry.linkFormatter)}>{format(item[schemaEntry.id], schemaEntry.valueFormatter)}</a>
+          {:else if schemaEntry.type === 'links'}
+            {#each item[schemaEntry.id] as ref}
+              <a
+                href={format(ref, schemaEntry.linkFormatter)}>{format(ref, schemaEntry.valueFormatter)}</a>&nbsp;
```
Iinh, line 40: I think for better visibility we should wrap each link in a div, so that it is its own line, rather than all links in one line that’s only separated by `&nbsp`.

```suggestion
<div>
<a href={format(ref, schemaEntry.linkFormatter)}>{format(ref, schemaEntry.valueFormatter)}</a>
</div>
```

src/pages/MetricDetail.svelte
Iinh: You could define this as a mixin in main.scss, then `@include` it anywhere you want to use it.
For example:
```
// main.scss
@mixin table-label {
font-size: 24px; }
```
Then, to use it in this component:
```suggestion
  h2 {
    @include table-label;
  }
```

src/pages/MetricDetail.svelte
```diff
@@ -113,132 +115,41 @@
     in
     <a href={`/apps/${params.app}`}>{params.app}</a>
```
Iinh, line 116: Should we include the app name here? I feel it's a bit redundant as the user can see the app name right on top of this with the breadcrumb.

src/pages/MetricDetail.svelte
```diff
@@ -113,132 +115,41 @@
...
     that
-    {getExpiryInfo(metric.expires)}
+    {getExpiryInfo(metric.expires)}.
```
Iinh, line 118: I think we could move this expiry info to the Metadata table. For the metric type, maybe make it a clickable pill to highlight it?

Iinh (commented):
This looks great! The added help texts are so useful, and thank you for making the metadata table more reusable! Regarding the display, I wonder if we could simplify the Metric page even more, right now the mixture of texts and metadata tables with various width sizes does not look as clean as I think it could be. I tested it out a bit and this is what I came up with, let me know if you think it's a good idea.

**Before**

image

**After**
Metric type as a clickable pill, simplified description field, expiry info moved to Metadata, and readjust the table layout.

image

src/pages/MetricDetail.svelte
wlach: This is definitely better than hardcoding 24px all over the place, but I was thinking more in terms of setting the desired height for `h2` globally (overriding what's defined by protocol), so we don't even have to do this -- I remember we considered using the condensed type scale defined here but it seemed a little small -- https://protocol.mozilla.org/fundamentals/typography.html

I fiddled a bit and think I found something that works-- I think this is better left for a subsequent PR though. For this one, I'll try your approach.

src/pages/MetricDetail.svelte
wlach: I think you're right, let's take out the app name.

src/pages/MetricDetail.svelte
wlach: wrt expiry, this makes sense now that we have a warning when something actually has expired. I'll make this change.

I think it's worth calling out the metric type explicitly in the header, as it's rather vitally important -- arguably the most important piece of information about a metric aside from the name.

src/components/MetadataTable.svelte
wlach: Makes sense, I'll make this change.

wlach:
@Iinh Ok have a new revision which addresses the issues mentioned. I really like the idea of having a two-column layout for the tables (where screen sizes permit), but maybe let's leave that to a followup? This PR is already pretty large. :)

Iinh (approved):
> @Iinh Ok have a new revision which addresses the issues mentioned. I really like the idea of having a two-column layout for the tables (where screen sizes permit), but maybe let's leave that to a followup? This PR is already pretty large. :)

Yes I'm happy to work on the layout as a followup :)

src/data/schemas.js
```diff
@@ -0,0 +1,195 @@
...
+    id: "url",
+    type: "link",
+    helpText: "Where the source code for this application lives.",
+  },
+  {
+    title: "Application ID",
+    id: "app_id",
+    type: "value",
+    helpText:
+      "The app's identifier exactly as it appears in the relevant app store listing (for relevant platforms) or in the app's Glean initialization call (for other platforms). In the case of the Google Play store, it may contain both dots and dashes. For applicable platforms, you should be able to construct an app store URL from this value.",
```
badboy, line 23:
```
In the case of the Google Play store, it may contain both dots and dashes.
```
Is that much detail needed here? Here the app id is read-only. They are defined in some other place, so I'm not sure how much that detail here really helps.

src/data/schemas.js
```diff
@@ -0,0 +1,195 @@
...
+    type: "value",
+    helpText:
+      "True if data collection for this metric is disabled. Sometimes this is used to temporarily disable the collection for a specific metric without removing references to it in the source code.",
+  },
+  {
+    title: "Labels",
+    id: "labels",
+    type: "list",
+    helpText:
+      "Different label names for this metric. FIXME: better explain this.",
```
badboy, line 112: Based on https://mozilla.github.io/glean_parser/metrics-yaml.html#metric-parameters,
maybe:

```
Only used for labeled metrics. If provided, the labels are enforced at runtime, and recording to an unknown label is recorded to the special label `__other__`.
```

badboy (approved):
Only 2 minimal comments, but otherwise good to go!

src/data/schemas.js
wlach: Yeah, agreed, we can skip the google play store part. This was taken from the definition in probe-scraper

wlach:
Thanks all!

PR #358, opened 2021-01-28
Fix logo font-size

Iinh:
Something I noticed: because Protocol's headers are responsive, if screen size goes to below 700px, the logo text "Dictionary" gets smaller while "Glean" stays the same size because it is an image and has a fixed width.

**Before**
Screen Shot 2021-01-28 at 8 11 42 AM

To prevent this, we need to hard code the font size value of "Dictionary" so that it is fixed regardless of screen size.

**After**
Screen Shot 2021-01-28 at 8 11 46 AM

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -120,6 +120,7 @@
         color: #e0e3e0;
         line-height: 0.7;
         padding-left: $spacing-xs;
+        font-size: 38px;
       }
     }
   }


wlach (approved):

wlach:
Thank you!

PR #359, opened 2021-01-28
Add a warning to Searchfox when the metric is expired

Iinh:
### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -7,6 +7,7 @@
   import HelpHoverable from "../components/HelpHoverable.svelte";
   import helpText from "../data/help";
   import PageTitle from "../components/PageTitle.svelte";
+  import WarningIcon from "../components/icons/WarningIcon.svelte";
 
   import { getSearchfoxLink } from "../formatters/searchfox";
   import Pill from "../components/Pill.svelte";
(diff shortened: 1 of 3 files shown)


src/pages/MetricDetail.svelte
```diff
@@ -219,6 +220,11 @@
           <HelpHoverable
             content={helpText.searchFox.text}
             link={helpText.searchFox.link} />
+          {#if isExpired(metric.expires)}
+            <HelpHoverable content={'This metric has expired.'}>
```
wlach, line 224:
```suggestion
            <HelpHoverable content={'This metric has expired and will likely not appear in Searchfox.'}>
```

src/pages/MetricDetail.svelte
```diff
@@ -219,6 +220,11 @@
...
+              <WarningIcon width={'15px'} height={'15px'} />
```
wlach, line 225: It would be nice to not hardcode this here, but maybe that's a problem to solve another time.

wlach (approved):
Approve with one suggestion. Thanks @Iinh!

src/pages/MetricDetail.svelte
Iinh: It was not possible to pass these values down as props in `HelpHoverable`. I hope that this use case is rare enough that hard coding is okay, for now.

PR #361, opened 2021-01-29
Warn more prominently when a metric is expired

wlach:
While working on #357, I realized that we should probably be generally much more explicit when a metric is expired. Many things will no longer work as one might expect.

image

Note that I expect to remove the searchbox-specific warning in #357 (I think it's redundant with the changes here), but leaving it in for now.

### Pull Request checklist

      fixes, if applicable)


Iinh (approved):
👍

Issue #223, opened 2020-11-24
Link back to metric and ping definitions in source control


wlach:
It is sometimes handy to be able to go *back* to the original definitions in GitHub (or other source repositories) where glean metrics and pings are defined (e.g. to fix a typo or make an edit). 

I would propose adding a "Defined in" field inside the metrics and ping pages that linked back to the metric and ping files at the appropriate line number. For example, for [the usage ping for mozregression](https://glean-dictionary-dev.netlify.app/#!/apps/mozregression/pings/usage), we would link back to:

https://github.com/mozilla/mozregression/blob/master/mozregression/pings.yaml#L4

The link name would be the unique part of the page after `/master/` (`mozregression/pings.yaml`)

Solving this issue would involve several steps:

1. Update the [glean_parser](https://github.com/mozilla/glean_parser/) to gather this information (this article on getting line number information from yaml files may prove useful: https://stackoverflow.com/questions/13319067/parsing-yaml-return-with-line-number) 
2. Gather this newly exposed data using glean-parser inside [probe-scraper](https://github.com/mozilla/probe-scraper/) and expose it.
3. Update our [own metadata gathering script](https://github.com/mozilla/glean-dictionary/blob/main/scripts/build-glean-metadata.py) to make use of this information.
4. Finally (the easy part) expose this data in the glean dictionary UI.

Iinh:
@wlach Can I work on this?

wlach:
> @wlach Can I work on this?

For sure! For (1) I would recommend reaching out to the Glean team on https://chat.mozilla.org/#/room/#glean:mozilla.org

PR #371, opened 2021-02-02
Only include current pings for a metric (fixes #354)

### LINKED ISSUE #354: Glean Dictionary incorrectly stating historical data
[This probe](https://dictionary.protosaur.dev/apps/fenix/metrics/browser_search_with_ads) is an example problem. It indicates that it is sent in two pings: baseline and metrics. However the [probe-info-service](https://probeinfo.telemetry.mozilla.org/glean/fenix/metrics) tells a different story, that it was only sent on the baseline ping in the spring of 2020, which means that now almost _no_ baseline pings would include it (only those from very old clients).

@wlach suggested that we should be matching the definition to versions, which would require use of the `git-commits` (but only for a linear history, I believe).

wlach:
### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/scripts/glean.py
+++ b/scripts/glean.py
@@ -49,11 +49,7 @@ class GleanMetric(GleanObject):
         self._set_definition(definition)
         self._set_description(self.definition)
 
-        defn_pings = set(
-            [p for d in definition[self.HISTORY_KEY] for p in d.get("send_in_pings", ["metrics"])]
-        )
-        self.definition["send_in_pings"] = defn_pings
-        self.is_client_info = "glean_client_info" in defn_pings
+        self.is_client_info = "glean_client_info" in self.definition["send_in_pings"]
         if ping_names is not None:
             self._update_all_pings(ping_names)
 


scripts/glean.py
```diff
@@ -49,11 +49,7 @@ def __init__(self, identifier: str, definition: dict, *, ping_names: List[str] =
         self._set_definition(definition)
         self._set_description(self.definition)
 
-        defn_pings = set(
-            [p for d in definition[self.HISTORY_KEY] for p in d.get("send_in_pings", ["metrics"])]
-        )
```
wlach, line 54: I am pretty sure this type of logic is unnecessary-- by the time probe-scraper outputs the files for us to consume, the defaults should already be set:

https://github.com/mozilla/glean_parser/blob/f9b676646f82d93ca2698540646a6d0bd29ab3ad/glean_parser/parser.py#L320

fbertsch (approved):
LGTM - we could keep a separate issue open with a per-version system. That seems like a larger design overhaul that affects the glean-dictionary in multiple places

wlach:
> LGTM - we could keep a separate issue open with a per-version system. That seems like a larger design overhaul that affects the glean-dictionary in multiple places

Yeah I really don't know what to do there. The probe dictionary shows the full json definition of what's in probe scraper, but that seems pretty half-baked to me and I'm not sure if that's much better than providing a link to the source (which we'll be doing in #223)

fbertsch:
> Yeah I really don't know what to do there. The probe dictionary shows the full json definition of what's in probe scraper, but that seems pretty half-baked to me and I'm not sure if that's much better than providing a link to the source (which we'll be doing in #223)

It is definitely half-baked. That was my idea because it became confusing trying to figure out what the probe dictionary was doing, and is certainly only useful to data engineers who understand the probe-scraper.

PR #372, opened 2021-02-02
Show all metrics in the item list (don't scroll)

wlach:
On request from @tdsmith. I do agree this seems better and more
in line with how other applications work. To prevent the list from
jumping around, I set a maximum height for each table cell (via a div
wrapper for the content). To make the full metric description easily
accessible, added a title tooltip which people can hover over.

image

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -5,7 +5,6 @@
 
   import { getItemURL } from "../state/urls";
 
-  import tippy from "./tippy";
   import Pagination from "./Pagination.svelte";
   import FilterInput from "./FilterInput.svelte";
   import Markdown from "./Markdown.svelte";
@@ -61,10 +60,8 @@
   }
 
   .item-property {
-    height: 24px;
+    height: 36px;
     overflow: hidden;
-    white-space: nowrap;
-    text-overflow: ellipsis;
   }
 
   table {
@@ -91,7 +88,7 @@
         background: rgba($color-link-hover, 0.1);
       }
       .description {
-        @include text-body-sm;
+        @include text-body-xs;
       }
     }
   }
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -62,6 +62,7 @@
   .item-property {
     height: 40px;
     overflow-y: auto;
+    margin: -0.25rem;
   }
 
   table {
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -121,8 +121,8 @@
       <!-- We have to do inline styling here to override Protocol CSS rules -->
       <!-- https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity -->
       <col width="35%" />
-      <col width={itemType === 'metrics' ? '25%' : '65%'} />
-      <col width={itemType === 'metrics' ? '40%' : '0'} />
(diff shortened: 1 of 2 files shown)


tdsmith:
Thanks, that's helpful!

I would probably bias towards displaying the entire description instead of truncating it -- if I'm looking at more than one result here, I probably need the full context to decide which, if any, of these probes are the thing I need. It's not visually obvious that truncated text will have a tooltip with the full string; I'd probably assume I had to click on the probe names one-by-one. I don't know if copy-pasting into a Google Sheet works at all but I'd wonder how this would interact with that.

wlach:
> Thanks, that's helpful!
> 
> I would probably bias towards displaying the entire description instead of truncating it -- if I'm looking at more than one result here, I probably need the full context to decide which, if any, of these probes are the thing I need. It's not visually obvious that truncated text will have a tooltip with the full string; I'd probably assume I had to click on the probe names one-by-one. I don't know if copy-pasting into a Google Sheet works at all but I'd wonder how this would interact with that.

The problem with showing the full description is that it will cause the pagination to jump around as the user navigates, since the amount of description is highly variable (of course we could set the height of each row to a ridiculously large number to ensure that's never a problem).

One compromise might be to use a smaller font for the description text and make the rows slightly larger. What do you think of this? 
image

wlach:
@tdsmith do you have any further feedback on the approach above, or should I go ahead and merge? You can play with it a bit at: https://deploy-preview-372--glean-dictionary-dev.netlify.app/apps/fenix

Also see @Iinh's followup in #375

wlach:
I'm pretty sure this is an improvement, let's go with this. @Iinh any thoughts?

src/components/ItemList.svelte
```diff
@@ -85,7 +88,7 @@
         background: rgba($color-link-hover, 0.1);
       }
       .description {
-        @include text-body-sm;
+        @include text-body-xs;
```
Iinh, line 91: This is too small imo, to the point where it's slightly hard to read. I think we should keep the old size, and make a few compromises in other places. 
```suggestion
        @include text-body-sm;
```

src/components/ItemList.svelte
```diff
@@ -54,13 +54,16 @@
...
-    overflow: scroll;
     a {
       text-decoration: none;
     }
   }
 
+  .item-property {
+    height: 36px;
+    overflow: hidden;
+  }
```
Iinh, line 65: Let's make the div a little bit taller. Also, if we set `overflow-y` to be auto here the content is still hidden when overflown, but with a scrollbar visible for users to read the rest of the content

```suggestion
  .item-property {
    height: 40px;
    overflow-y: auto;
  }
```

Iinh (commented):
I like the new layout. To give the description column more space, we could also increase its width.
```
      <col width="30%" />
      <col width={itemType === 'metrics' ? '20%' : '70%'} />
      <col width={itemType === 'metrics' ? '50%' : '0'} />
```
https://github.com/mozilla/glean-dictionary/blob/c784f551f713915bae79cf4dc89eb78f8742a299/src/components/ItemList.svelte#L120-L122

tdsmith:
I like this a lot; super sorry for the delayed feedback here.

Iinh (approved):

PR #375, opened 2021-02-05
[WIP] Add items per page input

Iinh:
Something that I thought about after reading #372: since users have different preferences when it comes to how they want the metric list to look like (more vs less pagination, longer vs shorter list, etc.), we should make the display more customizable. In this PR I experimented with letting users choose how many items they want shown per page. @wlach If you think this could be helpful I'll work on improving the input UI.

Screen Shot 2021-02-05 at 9 50 24 AM

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/__snapshots__/storyshots.test.js.snap
+++ b/src/__snapshots__/storyshots.test.js.snap
@@ -462,13 +462,28 @@ exports[`Storyshots ItemList Default 1`] = `
   class="storybook-snapshot-container"
 >
   <span
-    class="expire-checkbox svelte-1wy46sp"
+    class="output-control svelte-o23uf0"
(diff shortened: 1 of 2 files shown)


wlach:
I think this is a cool idea, however I might just provide a "show all" button on the bottom which removes the pagination (I think that's more in the spirit of what @tdsmith was asking for).

image

Iinh:
Thank you for the feedback. I'll change this to a "Show All" button

tdsmith:
A "no pagination" option would be awesome!

Iinh:
I've opened #403 for the "Show All"/"No Paginate" option.

PR #403, opened 2021-02-16
Add "Paginate" checkbox

Iinh:
From the discussion in #375, I add a "Paginate" checkbox so users can choose whether or not they want the metric display to be paginated.
CleanShot 2021-02-16 at 11 17 26@2x

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -12,7 +12,7 @@
 
   import { isExpired } from "../state/metrics";
 
-  let DEFAULT_ITEMS_PER_PAGE = 20;
+  let itemsPerPage = 20;
 
   export let appName;
   export let items;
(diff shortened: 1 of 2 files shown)


src/components/ItemList.svelte
```diff
@@ -111,6 +112,10 @@
         <input type="checkbox" bind:checked={showExpired} />
         Show expired metrics
       </label>
+      <label>
+        <input type="checkbox" bind:checked={paginated} />
+        Paginate
+      </label>
```
Iinh, line 118: - I opted to put this checkbox on top instead of below the Pagination bar because I thought that if the user unclicked this box but changed their mind, they would not have to scroll past hundreds of items to get to the bottom to find it. 

- "Paginate" instead of "Show All" because next to the "Show expired metrics" box, "Show All" might not be as clear a name as "Paginate"? 

What do you think @wlach?

wlach (approved):
Looks reasonable! Let's land it and see how it works for people.

PR #455, opened 2021-03-08
Markdown improvements

wlach:
* Add option to not display inline, which is what we want for larger
  free-form text
* Fix display of lists by using protocol's styles


wlach:
This substantially improves the display of metric descriptions that involved lists. E.g. compare:

https://deploy-preview-455--glean-dictionary-dev.netlify.app/apps/firefox-ios-dev/metrics/bookmarks_add

With:

https://dictionary.protosaur.dev/apps/firefox-ios-dev/metrics/bookmarks_add

src/components/Markdown.svelte
```diff
@@ -1,9 +1,27 @@
 <script>
-  import { parseInline } from "marked";
+  // you would think we could just do:
+  // import marked from "marked", but that seems to cause
+  // weird problems with jest and storyshots (probably because
+  // of how it imports/compiles ES modules). this is the only
+  // workaround I found
```
wlach, line 6: This makes me grumpy, but I couldn't find any alternative. :( I suspect this is an issue with marked (it came up earlier as well, though I can't find the reference)

src/components/Markdown.svelte
```diff
@@ -1,9 +1,27 @@
...
+  export let inline = true;
 </script>
 
-<slot>
+{#if inline}
   {@html parseInline(text)}
-</slot>
+{:else}
+  {@html parse(text)}
+{/if}
```
robhudson, line 27: Maybe a code comment about why this was needed or what it accomplishes.

src/pages/MetricDetail.svelte
```diff
@@ -90,9 +90,7 @@
 
   <PageTitle text={metric.name} />
 
-  <p>
-    <Markdown text={metric.description} />
-  </p>
+  <Markdown text={metric.description} inline={false} />
```
robhudson, line 93: The `text` seems more "HTML"-like if it were in a `slot`, perhaps?

src/components/Markdown.svelte
robhudson: Bummer, but maybe a comment to that effect for future code readers is fine.

robhudson (changes requested):
r+wc

src/pages/MetricDetail.svelte
wlach: Yeah, this is the way we implemented it originally but it needed some hacks to work and had some problems. @Iinh reverted this behaviour in #192

wlach: (as such I'm going to merge this in, but if you want to follow up on this let me know)

PR #468, opened 2021-03-17
Add Sourcegraph links for metrics not indexed by Searchfox

### LINKED ISSUE #404: Use sourcegraph to search for code in repos not indexed by searchfox
In #339 @Iinh added support for searching for glean metric definitions using searchfox. However, not everything we care about is currently indexed by searchfox.

Seperately, @Iinh found a great service called sourcegraph which indexes at least some of these repositories. For example, mozregression:

https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/mozilla/mozregression%24+&patternType=literal

For these cases, we should provide links to sourcegraph. 

This is on the complex end of things for a good first bug, but should be doable. Fixing this issue would involve:

* Looking at how @Iinh implemented things in #339
* Implementing similar functionality for getting links to sourcegraph (note we would *only* return links to sourcegraph where we do not have a searchfox link, since searchfox is more familiar and probably better for those cases) and putting them in the metadata table.

jcads:
Fixes #404 

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/data/schemas.js
+++ b/src/data/schemas.js
@@ -5,7 +5,7 @@ import {
   getBugLinkTitle,
   getSourceUrlTitle,
 } from "../formatters/links";
-import { getSearchfoxLink } from "../formatters/searchfox";
+import { getMetricLink } from "../formatters/searchfox";
 
 const REQUIRED_METRIC_PARAMS_DOCS =
   "https://mozilla.github.io/glean/book/user/metric-parameters.html#required-metric-parameters";
@@ -64,7 +64,7 @@ export const METRIC_DEFINITION_SCHEMA = [
     helpText:
       "Finds uses of this metric using simple matching. Metrics used dynamically may not appear in the results, only metrics defined in the application will be found.",
     linkFormatter: (metricId, appName) => {
-      return getSearchfoxLink(appName, metricId);
+      return getMetricLink(appName, metricId);
     },
   },
 ];
--- a/src/data/schemas.js
+++ b/src/data/schemas.js
@@ -5,7 +5,7 @@ import {
   getBugLinkTitle,
   getSourceUrlTitle,
 } from "../formatters/links";
-import { getMetricLink } from "../formatters/searchfox";
+import { getCodeSearchLink } from "../formatters/codesearch";
 
 const REQUIRED_METRIC_PARAMS_DOCS =
   "https://mozilla.github.io/glean/book/user/metric-parameters.html#required-metric-parameters";
@@ -58,13 +58,13 @@ export const METRIC_DEFINITION_SCHEMA = [
     valueFormatter: getSourceUrlTitle,
   },
   {
-    title: "Searchfox",
+    title: "Code Search",
     id: "name",
     type: "link",
     helpText:
       "Finds uses of this metric using simple matching. Metrics used dynamically may not appear in the results, only metrics defined in the application will be found.",
     linkFormatter: (metricId, appName) => {
-      return getMetricLink(appName, metricId);
+      return getCodeSearchLink(appName, metricId);
     },
   },
 ];
--- a/src/formatters/searchfox.js
+++ b/src/formatters/searchfox.js
@@ -30,7 +30,21 @@ const searchfoxMap = {
   reference_browser: "reference-browser",
 };
 
-export const getSearchfoxLink = (app, metric) => {
+const sourcegraphMap = {
+  klar_ios: "focus-ios",
+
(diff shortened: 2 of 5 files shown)


src/formatters/searchfox.js
```diff
@@ -30,7 +30,21 @@ const searchfoxMap = {
   reference_browser: "reference-browser",
 };
 
-export const getSearchfoxLink = (app, metric) => {
+const sourcegraphMap = {
```
wlach, line 33: Could you rename this file to something like codesearch.js?

src/formatters/searchfox.js
```diff
@@ -30,7 +30,21 @@ const searchfoxMap = {
...
+
+  mozphab: "mozphab",
+
+  mozregression: "mozregression",
+};
+
+export const getMetricLink = (app, metric) => {
   const [category, name] = metric.split(/\.(?=[^.]+$)/);
 
   // Generate all possible casings for a Searchfox query:
```
wlach, line 50:
```suggestion
  // Generate all possible casings for a search query:
```

src/formatters/searchfox.js
```diff
@@ -63,5 +77,9 @@ export const getSearchfoxLink = (app, metric) => {
       ? `https://searchfox.org/mozilla-central/search?q=${allLanguagePatterns}&regexp=true`
       : `https://searchfox.org/mozilla-mobile/search?q=${allLanguagePatterns}&path=${searchfoxMap[app]}&regexp=true`;
   }
-  return undefined;
+
+  return (
```
wlach, line 81: Let's check to make sure there is an entry for this app in sourcegraph, just as we do above for searchfox.

There may be some cases where we don't have a code search service for a repository.

src/formatters/searchfox.js
```diff
@@ -30,7 +30,21 @@ const searchfoxMap = {
...
+  lockwise_ios: "lockwise-ios",
+
+  mach: "mach",
+
+  mozphab: "mozphab",
+
+  mozregression: "mozregression",
+};
+
+export const getMetricLink = (app, metric) => {
```
wlach, line 47:
```suggestion
export const getCodeSearchLink = (app, metric) => {
```

src/data/schemas.js
```diff
@@ -64,7 +64,7 @@ export const METRIC_DEFINITION_SCHEMA = [
     helpText:
       "Finds uses of this metric using simple matching. Metrics used dynamically may not appear in the results, only metrics defined in the application will be found.",
     linkFormatter: (metricId, appName) => {
-      return getSearchfoxLink(appName, metricId);
+      return getMetricLink(appName, metricId);
```
wlach, line 67: Could you rename this entry to `Code Search`? Seems more intuitive to me.

wlach (changes requested):
Thanks @jcads! Just have a few requests for changes before this lands.

wlach (approved):
Thanks for this!

The only bug I can see is that it seems like we're still showing the "code search" option (with no link) even when it's unavailable. e.g. https://deploy-preview-468--glean-dictionary-dev.netlify.app/apps/bergamot/metrics/errors_marian

image

However that seems like an old issue, so I'm going to merge this as-is. 

@jcads would you be interested in investigating that as a followup?

jcads:
You're welcome, @wlach! :tada: 

As for the bug, I'll gladly look into that.

jcads:
I did some digging and I think it's because `bergamot` has just recently been added to https://probeinfo.telemetry.mozilla.org/v2/glean/app-listings, where the glean-dictionary data is pulled from. Therefore, it wasn't added to `codesearch.js` to be indexed.

And just a sidenote: bergamot can't be currently indexed on both searchfox and sourcegraph, so adding it in `codesearch.js` will not fix the bug.

Should the "code search" option be removed when the link is unavailable?

wlach:
> And just a sidenote: bergamot can't be currently indexed on both searchfox and sourcegraph, so adding it in `codesearch.js` will not fix the bug.

Right :) I asked about this on the #github-admin channel on Matrix today. I think searchgraph is not yet indexing the `mozilla-extensions` organization. I can ping them later about this.
 
> Should the "code search" option be removed when the link is unavailable?

I think it would be better to just say something like `Unavailable: no search index specified for this application.`. That way people know the reason why it's missing.

jcads:
> I think it would be better to just say something like Unavailable: no search index specified for this application.. That way people know the reason why it's missing.

Got it! Opening a new issue.

PR #480, opened 2021-03-25
Fix repo names in Sourcegraph map

Iinh:
Something I noticed while playing around with the new Sourcegraph implementation, which looks great by the way @jcads!

Example:

Old code search for [`mach.command`](https://dictionary.protosaur.dev/apps/mach/metrics/mach_command):
https://sourcegraph.com/search?q=repo:%5Egithub%5C.com%5C/%5BMm%5Dozilla%28.*%29%5C/mach%24+mach.command|Mach.command|mach.command|mach::command&patternType=regexp

New code search: 
https://sourcegraph.com/search?q=repo:%5Egithub%5C.com%5C/%5BMm%5Dozilla%28.*%29%5C/gecko-dev%24+mach.command|Mach.command|mach.command|mach::command&patternType=regexp
 
### Pull Request checklist

- [x] The pull request has a descriptive title (and a reference to an issue it
      fixes, if applicable)
- [x] All tests and linter checks are passing
- [x] The pull request is free of merge conflicts

PR #481, opened 2021-03-25
Re-add mozphab repo to sourcegraph repos

wlach:
@Iinh correctly pointed out in #480 that we do have sourcegraph indexing for mozphab. So this PR re-adds it. However the search string currently fails because it winds up being something like:

https://sourcegraph.com/search?q=repo:%5Egithub%5C.com%5C/%5BMm%5Dozilla%28.*%29%5C/mozphab%24+mozphabEnvironment.distributionVersion|MozphabEnvironment.distributionVersion|mozphab_environment.distribution_version|mozphab_environment::distribution_version&patternType=regexp

for `mozphab.environment.distribution_version`

Note that this (adhoc) query works fine:

https://sourcegraph.com/search?q=repo:%5Egithub%5C.com%5C/%5BMm%5Dozilla%28.*%29%5C/review%24+mozphab.environment.distribution_version&patternType=regexp

I think the code to create the search regex wasn't designed to handle metrics with three levels of nesting like this. Not sure what the right solution is here, @badboy any ideas? Code in question is here:

https://github.com/mozilla/glean-dictionary/blob/0de7cbab440aa683275fcef01bc969927dbc9e3b/src/formatters/codesearch.js#L46

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/formatters/codesearch.js
+++ b/src/formatters/codesearch.js
@@ -39,6 +39,8 @@ const sourcegraphMap = {
 
   mach: "gecko-dev",
 
+  mozphab: "review",
+
   mozregression: "mozregression",
 };
 
--- a/src/formatters/codesearch.js
+++ b/src/formatters/codesearch.js
@@ -70,7 +70,7 @@ export const getCodeSearchLink = (app, metric) => {
     ? `${snakeCase(category)}::${snakeCase(name)}`
     : snakeCase(category);
 
-  const allLanguagePatterns = `${camelCased}|${capitalizedCamelCased}|${snakedCased}|${dblColonSnakeCased}`;
+  const allLanguagePatterns = `${metric}|${camelCased}|${capitalizedCamelCased}|${snakedCased}|${dblColonSnakeCased}`;
 
   /* eslint no-else-return: "error" */
 
--- a/tests/formatters.codesearch.test.js
+++ b/tests/formatters.codesearch.test.js
@@ -3,12 +3,12 @@ import { getCodeSearchLink } from "../src/formatters/codesearch";
 describe("Searchfox links for firefox-desktop", () => {
   it("works as expected", () => {
     expect(getCodeSearchLink("firefox_desktop", "test.metric_name")).toEqual(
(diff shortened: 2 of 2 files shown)


Iinh:
> I think the code to create the search regex wasn't designed to handle metrics with three levels of nesting like this. Not sure what the right solution is here, @badboy any ideas? Code in question is here:

https://github.com/mozilla/glean-dictionary/blob/0de7cbab440aa683275fcef01bc969927dbc9e3b/src/formatters/codesearch.js#L72
I think we need to add the original metric name to the search regex above. I tested it locally and it seems to work.
`${metric}|${camelCased}|${capitalizedCamelCased}|${snakedCased}|${dblColonSnakeCased};`

wlach:
>  I think we need to add the original metric name to the search regex above. I tested it locally and it seems to work.
> `${metric}|${camelCased}|${capitalizedCamelCased}|${snakedCased}|${dblColonSnakeCased};`

Yep, I think that would work for python, but not sure about (e.g.) Kotlin. Wouldn't mind getting some guidance from @badboy or maybe @mdboom (when he returns) on this.

badboy:
> > I think we need to add the original metric name to the search regex above. I tested it locally and it seems to work.
> > `${metric}|${camelCased}|${capitalizedCamelCased}|${snakedCased}|${dblColonSnakeCased};`
> 
> Yep, I think that would work for python, but not sure about (e.g.) Kotlin. Wouldn't mind getting some guidance from @badboy or maybe @mdboom (when he returns) on this.

I think it should be fine.
For Kotlin, Swift and Rust we do convert things like `mozphab.environment` into a valid identifier, by either camelCasing or snake_casing it.
It's just in Python that we keep the nesting as is (and tbh: I wasn't aware that we do until I looked here).

wlach:
Cool! I just submitted a PR that adjusts things how @Iinh suggested. It works great.

https://deploy-preview-481--glean-dictionary-dev.netlify.app/apps/mozphab/metrics/mozphab_environment_distribution_version

Iinh (approved):
LGTM!

PR #493, opened 2021-03-29
Ping Annotations

wlach:
Depends on https://github.com/mozilla/glean-annotations/pull/8

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -9,7 +9,7 @@ import requests
 import stringcase
 
 OUTPUT_DIRECTORY = os.path.join("public", "data")
-ANNOTATIONS_URL = "https://mozilla.github.io/glean-annotations/api.json"
+ANNOTATIONS_URL = "https://deploy-preview-8--blissful-roentgen-ca62bb.netlify.app/api.json"
 
 
 def _serialize_sets(obj):
@@ -139,7 +139,9 @@ for (app_name, app_group) in app_groups.items():
                 app_metrics[metric.identifier] = dict(
                     metric.definition,
                     name=metric.identifier,
-                    annotation=annotations_index.get(app_name, {}).get(metric.identifier),
+                    annotation=annotations_index.get(app_name, {})
+                    .get("metrics", {})
+                    .get(metric.identifier),
                     repo_url=app.app["url"],
                     variants=[],
                 )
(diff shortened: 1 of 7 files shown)


scripts/build-glean-metadata.py
```diff
@@ -9,7 +9,7 @@
 import stringcase
 
 OUTPUT_DIRECTORY = os.path.join("public", "data")
-ANNOTATIONS_URL = "https://mozilla.github.io/glean-annotations/api.json"
+ANNOTATIONS_URL = "https://deploy-preview-8--blissful-roentgen-ca62bb.netlify.app/api.json"
```
wlach, line 12: This is just for testing, it should be reverted before this lands.

acmiyaguchi: Maybe something to to pull from the environment via `.env`, if it's a common enough workflow.

scripts/build-glean-metadata.py
```diff
@@ -139,7 +139,9 @@ def etl_snake_case(line: str) -> str:
                 app_metrics[metric.identifier] = dict(
                     metric.definition,
                     name=metric.identifier,
-                    annotation=annotations_index.get(app_name, {}).get(metric.identifier),
+                    annotation=annotations_index.get(app_name, {})
+                    .get("metrics", {})
```
acmiyaguchi, line 143: This is black's behavior on long chainable arguments? Imo not the prettiest look (just throwing my opinion out there).

acmiyaguchi (approved):
Sorry I didn't see this alongside the other PR. LGTM

scripts/build-glean-metadata.py
wlach: Agreed, it's not great. Maybe we can find a better way of expressing this in the future.

scripts/build-glean-metadata.py
wlach: I'm not sure -- in this case I did it so people could preview the behaviour inside the deploy preview here. An environment variable wouldn't solve that case. 

It's worth bearing in mind though, I can see this being useful in some other cases. If this comes up again I'll definitely add this feature.

scripts/build-glean-metadata.py
acmiyaguchi: Maybe wrapping the whole expression in another set of parenthesis might add an indent. Nothing worth nitpicking over though.

wlach: Yeah, actually this helped a bunch! I was in the file anyway to revert the change for the annotations URL so it was no trouble.

PR #497, opened 2021-04-02
Translate markdown with angle brackets to HTML (fixes #476)

### LINKED ISSUE #476: Markdown with angle brackets in it doesn't get translated to html correctly
See: https://dictionary.protosaur.dev/apps/fenix/metrics/browser_search_with_ads

image

Underlying JSON: https://dictionary.protosaur.dev/data/fenix/metrics/browser_search_with_ads.json

My guess is that the problem is somewhere in our usage of the markdown library here: https://github.com/mozilla/glean-dictionary/blob/main/src/components/Markdown.svelte

After fixing, would be good to add a testcase to the story: https://github.com/mozilla/glean-dictionary/blob/main/stories/markdown.stories.js

@jcads Any interest in this one? 😄

jcads:
https://deploy-preview-497--glean-dictionary-dev.netlify.app/apps/fenix/metrics/browser_search_with_ads
image

Fixes #476.

***
### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -17,13 +17,29 @@
   use({ renderer });
 
   export let text;
-
-  // if inline is set, do not wrap the markdown in a paragraph -- useful for short snippits
+  // if inline is set, do not wrap the markdown in a paragraph -- useful for short snippets
   export let inline = true;
+
+  // Split using the regex "/\r?\n|\r/" which should be able to
+  // detect most common line ending formats (win / mac / unix).
+  // ex: "This\nis\n\nan example" -> ["This", "is", "", "an example"]
+  let lines = text.split(/\r?\n|\r/);
 </script>
 
-{#if inline}
-  {@html parseInline(text)}
+{#if text.includes('<')}
+  {#if inline}
+    {text}
+  {:else}
+    {#each lines as text}
+      <p style="margin: 0">{text}</p>
+    {/each}
+  {/if}
+  <!-- hack to get consistent margin -->
+  <p />
 {:else}
-  {@html parse(text)}
+  {#if inline}
+    {@html parseInline(text)}
+  {:else}
+    {@html parse(text)}
+  {/if}
 {/if}
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -20,7 +20,7 @@
   // if inline is set, do not wrap the markdown in a paragraph -- useful for short snippets
   export let inline = true;
 
-  let htmlText = text.replace("<", "&lt;")
+  let htmlText = text.replace("<", "&lt;");
 </script>
 
 {#if inline}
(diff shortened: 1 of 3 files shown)


src/components/Markdown.svelte
```diff
@@ -17,13 +17,29 @@
...
-{#if inline}
-  {@html parseInline(text)}
+{#if text.includes('<')}
+  {#if inline}
+    {text}
+  {:else}
+    {#each lines as text}
+      <p style="margin: 0">{text}</p>
+    {/each}
+  {/if}
```
Iinh, line 36: We should always use parse/parseInline to render `text` in this component as the goal is to display markdown content.

Iinh (commented):
Hi @jcads, thank you for working on this issue. This solution looks like a great start, I have some input below.

I believe the cause of this bug is since `<` and `>` are reserved characters in HTML, the browser mistook `<provider-name>` as an HTML tag. To fix this, we could use [character entities](https://developer.mozilla.org/en-US/docs/Glossary/Entity) to display these special characters instead. 

For example, if we pass:

`htmlText = text.replace("<", "&lt;");` 

to `parse()`, instead of just `text`, the browser would then be able to read "<".

Wondering what @wlach thinks about this approach. If this seems reasonable, should we also take care of the other edge cases, such as `&`?

src/components/Markdown.svelte
```diff
@@ -17,13 +17,29 @@
   use({ renderer });
```
wlach, line 17: I think we need to enhance the above renderer object to "escape" angle brackets found in HTML tags, as I don't think we want to allow people to specify HTML in their markdown descriptions.

wlach (changes requested):
Hi @jcads -- thanks for the PR I think this solution is not quite right. As @Iinh, we want to continue using parse/parseInline to render the text.

What's going on here is the markdown renderer is creating `<provider-name>` as an HTML element. We can confirm this by looking at the developer console:

image

I think a shorter/cleaner solution would be to enhance the renderer object to escape these objects. 

https://marked.js.org/using_pro#renderer

I think you might need to add a render for `tag` types, but am not 100% sure. Give it a try and see!

wlach:
> I believe the cause of this bug is since `<` and `>` are reserved characters in HTML, the browser mistook `<provider-name>` as an HTML tag. To fix this, we could use [character entities](https://developer.mozilla.org/en-US/docs/Glossary/Entity) to display these special characters instead.
> 
> For example, if we replace the special characters with their equivalent HTML entities, such as:
> 
> `htmlText = text.replace("<", "&lt;");`
> 
> and then pass it to `parse(htmlText)`, the browser would then be able to read "<".
> 
> Wondering what @wlach thinks about this approach. If this seems reasonable, should we also take care of the other edge cases, such as `&`?

Thanks @Iinh! This is actually a better approach than what I suggested. I think the markdown parser will automatically handle characters like `&` ok -- it's just HTML-like tags that confuse it. @jcads do you want to give her approach a try?

src/components/Markdown.svelte
jcads: Got it @Iinh. Thank you.

wlach (approved):
Looks great, thanks for persisting on this @jcads. Let me know if / when you're interested in working on something else.

PR #509, opened 2021-04-05
Make variable declaration in markdown component reactive

wlach:
This fixes pagination when navigating through long lists of metrics (the descriptions weren't being updated after #497). I missed this in review (/cc @jcads)

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -20,6 +20,8 @@
   // if inline is set, do not wrap the markdown in a paragraph -- useful for short snippets
   export let inline = true;
 
+  // escape `<` so it doesn't get confused as an HTML tag, see:
+  // https://github.com/mozilla/glean-dictionary/pull/497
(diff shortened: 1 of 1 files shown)


src/components/Markdown.svelte
```diff
@@ -20,7 +20,7 @@
   // if inline is set, do not wrap the markdown in a paragraph -- useful for short snippets
   export let inline = true;
 
-  let htmlText = text.replace("<", "&lt;");
+  $: htmlText = text.replace("<", "&lt;");
```
Iinh, line 23: Should we add a comment here to explain why we needed to replace this string?

Iinh (approved):
👍

src/components/Markdown.svelte
wlach: Yes, this makes sense. I'll add a quick comment before merging.

PR #510, opened 2021-04-06
Fix extraction of css to bundle.css

wlach:
The behaviour changed in a recent version of the rollup plugin, updating
our code to match. This also removes our usage of postcss, which I
believe is no longer needed.

### Pull Request checklist

      fixes, if applicable)


rollup.config.js
```diff
@@ -34,6 +34,10 @@ function serve() {
 }
 
 export default {
+  onwarn: (warning, handler) => {
+    if (warning.message.indexOf("Unused CSS selector") !== -1) return;
+    handler(warning);
+  },
```
wlach, line 40: This is strictly unrelated to this PR, but prevents the console from filling up with warnings while we buid the project.

wlach:
@Iinh I must confess I never really understood why we needed postcss (it came in through the [app template](https://github.com/graph-paper-org/app-template/blob/main/rollup.config.js) I used to start this project), however removing it makes things work as expected: if I leave it in, it seems to compile CSS to JavaScript, which is not what we want. Do you have any concerns?

rollup.config.js
Iinh: This is great! We also do something similar in storybook https://github.com/mozilla/glean-dictionary/blob/8d9a6a2b6aed9648c025210ee2ca30066b54165f/.storybook/main.js#L32-L36

Iinh (changes requested):
> @Iinh I must confess I never really understood why we needed postcss (it came in through the app template I used to start this project), however removing it makes things work as expected: if I leave it in, it seems to compile CSS to JavaScript, which is not what we want. Do you have any concerns?

Initially we needed postcss because we were using Tailwind, and Tailwind is a postcss plugin. If I remember correctly we were also using several other plugins such as autoprefixer. After migrating to Protocol, the only postcss plugin we need now is postcss-import, which we use to import css libraries to use as global styles:
https://github.com/mozilla/glean-dictionary/blob/8d9a6a2b6aed9648c025210ee2ca30066b54165f/src/GlobalStyles.svelte#L1-L4

If we remove postcss, postcss-import would no longer work. The tippy library now doesn't work anymore in the preview https://deploy-preview-510--glean-dictionary-dev.netlify.app/apps/focus_ios

CleanShot 2021-04-06 at 10 36 17@2x

wlach:
> Initially we needed postcss because we were using Tailwind, and Tailwind is a postcss plugin. If I remember correctly we were also using several other plugins such as autoprefixer. After migrating to Protocol, the only postcss plugin we need now is postcss-import, which we use to import css libraries to use as global styles:

Got it, it looks like we can get the svelte plugin to handle postcss just by passing `postcss: true` as an option and this seems to be what's recommended these days (as opposed to using rollup-plugin-postcss):

https://github.com/sveltejs/sapper/issues/699#issuecomment-742841398

With that updated, tooltips seem to show as expected.

Iinh (approved):
Looks good!

PR #511, opened 2021-04-06
Pull in feature annotations from glean-annotations (where available)

wlach:
This is a quick draft / checkpoint of feature annotations to go along with mozilla/glean-annotations#12. Expect changes.

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -12,7 +12,7 @@ import stringcase
 import yaml
 
 OUTPUT_DIRECTORY = os.path.join("public", "data")
-ANNOTATIONS_URL = "https://deploy-preview-12--glean-annotations.netlify.app/api.json"
+ANNOTATIONS_URL = "https://mozilla.github.io/glean-annotations/api.json"
 NAMESPACES_URL = "https://raw.githubusercontent.com/mozilla/looker-hub/main/namespaces.yaml"
 
 # Priority for getting metric data (use the later definitions of nightly over release)
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -173,8 +173,7 @@ open(os.path.join(OUTPUT_DIRECTORY, "apps.json"), "w").write(json.dumps(list(app
 for (app_name, app_group) in app_groups.items():
     app_dir = os.path.join(OUTPUT_DIRECTORY, app_name)
     (app_id_dir, app_ping_dir, app_table_dir, app_metrics_dir) = (
-        os.path.join(app_dir, subtype)
-        for subtype in ("app_ids", "pings", "tables", "metrics")
+        os.path.join(app_dir, subtype) for subtype in ("app_ids", "pings", "tables", "metrics")
     )
     for directory in (app_id_dir, app_ping_dir, app_table_dir, app_metrics_dir):
         os.makedirs(directory, exist_ok=True)
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -451,30 +451,6 @@ for (app_name, app_group) in app_groups.items():
             )
         )
 
-    # write labels (if any)
-    labels = [
-        {"name": k, "description": v}
-        for (k, v) in annotations_index.get(app.app_name, {}).get("labels", {}).items()
-    ]
-    if labels:
-        app_data["labels"] = labels
-        for label in labels:
-            label_metrics = [
-                metric
-                for metric in app_data["metrics"]
-                if label["name"] in metric.get("labels", [])
-            ]
-            label["metric_count"] = len(label_metrics)
-            open(os.path.join(app_labels_dir, f"{label['name']}.json"), "w").write(
-                json.dumps(
-                    dict(
-                        label,
-                        origin=app.app_name,
-                        metrics=label_metrics,
-                    )
-                )
-            )
-
     # sort the information in the app-level summary, then write it out
     # (we don't sort application id information, that's already handled
     # above)
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -451,6 +451,21 @@ for (app_name, app_group) in app_groups.items():
             )
         )
 
+    # write labels (if any)
+    labels = [
+        {"name": k, "description": v}
+        for (k, v) in annotations_index.get(app.app_name, {}).get("labels", {}).items()
+    ]
+    if labels:
+        app_data["labels"] = labels
+        for label in labels:
+            label_metrics = [
+                metric
+                for metric in app_data["metrics"]
+                if label["name"] in metric.get("labels", [])
+            ]
+            label["metric_count"] = len(label_metrics)
+
     # sort the information in the app-level summary, then write it out
     # (we don't sort application id information, that's already handled
     # above)
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -12,7 +12,7 @@ import stringcase
 import yaml
 
 OUTPUT_DIRECTORY = os.path.join("public", "data")
-ANNOTATIONS_URL = "https://mozilla.github.io/glean-annotations/api.json"
+ANNOTATIONS_URL = "https://deploy-preview-12--glean-annotations.netlify.app/api.json"
 NAMESPACES_URL = "https://raw.githubusercontent.com/mozilla/looker-hub/main/namespaces.yaml"
 
 # Priority for getting metric data (use the later definitions of nightly over release)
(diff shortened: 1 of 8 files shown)


wlach:
Current implementation resembles GitHub's labels feature:

image

scripts/build-glean-metadata.py
```diff
@@ -9,7 +9,7 @@
 import stringcase
 
 OUTPUT_DIRECTORY = os.path.join("public", "data")
-ANNOTATIONS_URL = "https://mozilla.github.io/glean-annotations/api.json"
+ANNOTATIONS_URL = "https://deploy-preview-12--blissful-roentgen-ca62bb.netlify.app/api.json"
```
wlach, line 12: We'll need to restore this before we land

wlach:
Hey @Iinh, I was wondering if you'd mind having a look at this. A couple of points about the implementation:

1. This adds feature annotations for Fenix, visible at: https://deploy-preview-511--glean-dictionary-dev.netlify.app/apps/fenix I'm just using standard pills here, which I'm not sure is the right "look" (although maybe it's ok for a first implementation).
2. Clicking on a pill for a feature will filter results to just that feature (there is some new logic in the item list to handle this). At the same time, I realized it would be nice for history to work for this case, so I modified some of the code around handling the URL state to allow for pushing the state in cases where it makes sense.
3. I wound up refactoring the store/context code to just use the store in most cases, since I needed that for point 2.

src/components/ItemList.svelte
```diff
@@ -20,44 +21,75 @@
...
+    filteredItems = features.has(search)
+      ? items.filter(
+          (item) =>
+            item.features &&
+            item.features.some(
+              (feature) => feature.toLowerCase() === search.toLowerCase()
+            )
+        )
+      : items.filter(
+          (item) => item.name.includes(search) || featureMatch(item)
```
Iinh, line 63: Maybe I'm not following this right, but why do we need `|| featureMatch(item)` when we only need to filter out results that match search text?

src/components/ItemList.svelte
```diff
@@ -20,44 +21,75 @@
...
-    );
-    // show the first page of result
-    currentPage.set(1);
-    // even if currentPage is already 1, we need to manually call goToPage() to get the first page
-    goToPage(1);
+    const perPage = paginated ? DEFAULT_ITEMS_PER_PAGE : filteredItems.length;
+    pagedItems =
+      filteredItems.length > 0
+        ? chunk([...filteredItems], perPage)[$currentPage - 1]
+        : [];
```
Iinh, line 85: Do you think we should update pageState to persist pagination state too?

Iinh (approved):
Looks really good! You seem to have taken care of all of the use cases for this feature 👍 

Style-wise, I think the Pill should be a little bit smaller with thinner font, probably a softer color too. I feel like the UI for ItemList is starting to look quite cluttery, what do you think? But I agree, for first pass this is ok.

src/components/ItemList.svelte
wlach: I think we probably should, yes. Let's save it for a followup though, I don't think it's too pressing. Filed #528

src/components/ItemList.svelte
wlach: We also want to match on features, even if only part of their string is matched (e.g. "sea" should match all metrics which have "search" as a feature, even if they do not have search in their name). I can add a comment about this.

wlach:
Feeling pretty good about where this is at. About the only thing that feels really wrong is the fact that the scroll position doesn't reset after clicking on a label in the "labels" view (or inside the metrics view itself). Have a plan to fix that in a followup, since it's not technically a new issue.

src/pages/AppDetail.svelte
```diff
@@ -73,8 +74,15 @@
   >
     <Tab key="metrics">Metrics</Tab>
     <Tab key="pings">Pings</Tab>
+    {#if app.labels && app.labels.length}
+      <Tab key="labels">Labels</Tab>
```
Iinh, line 78: Personally I find "Features", instead of "Labels", more idiomatic. I think calling these "labels" is too general and it could be confusing at first to understand what the labels mean, as we also use labels for the glean libraries, item status, etc. The convention for these labels in the fenix repo uses `Features:` as a prefix, so maybe we should call these labels "Features" too?

src/state/api.js
```diff
@@ -13,6 +13,10 @@ export async function getAppIdData(appName, appId) {
   return fetchJSON(`/data/${appName}/app_ids/${appId}.json`);
 }
 
+export async function getLabelData(appName, labelName) {
```
Iinh, line 16: Why do we need this function? I don't see it being used anywhere?

src/components/ItemList.svelte
```diff
@@ -16,6 +16,7 @@
   export let appName;
   export let items;
   export let itemType;
+  export let containerName;
```
Iinh, line 19: Should probably remove this.

src/App.svelte
```diff
@@ -38,6 +38,7 @@
       ...(appId
         ? [{ url: `/apps/${app}/app_ids/${appId}/`, name: appId }]
         : []),
+      ...(label ? [{ url: `/apps/${app}/labels/${label}/`, name: label }] : []),
```
Iinh, line 41: Why do we need this?

scripts/build-glean-metadata.py
```diff
@@ -443,9 +451,34 @@ def get_app_variant_description(app):
...
+            label["metric_count"] = len(label_metrics)
+            open(os.path.join(app_labels_dir, f"{label['name']}.json"), "w").write(
+                json.dumps(
+                    dict(
+                        label,
+                        origin=app.app_name,
+                        metrics=label_metrics,
+                    )
+                )
+            )
```
Iinh, line 476: Why do we need to write label data to separate json files? It looks like write the data here but never use it, as the label data needed was pulled from the main `index.json` file rather than `[label].json`

Iinh (changes requested):
A few things to clean up, other than that this looks good!

src/pages/AppDetail.svelte
wlach: Yeah, I think it's not going to be the same for every application. But maybe we can make the user-visible name we show to people configure. "Features" definitely makes the most sense for Fenix. Will investigate.

src/state/api.js
wlach: I think it's a leftover from an earlier iteration, will remove.

src/App.svelte
wlach: Leftover from an earlier experiment. Will take it out.

scripts/build-glean-metadata.py
wlach: Yup, another leftover. Will take this out.

src/pages/AppDetail.svelte
wlach: Ok, I'm just going to call them "Features" for this iteration, since this only applies to Fenix.

Iinh: 👍

scripts/build-glean-metadata.py
```diff
@@ -12,7 +12,7 @@
 import yaml
 
 OUTPUT_DIRECTORY = os.path.join("public", "data")
-ANNOTATIONS_URL = "https://mozilla.github.io/glean-annotations/api.json"
+ANNOTATIONS_URL = "https://deploy-preview-12--glean-annotations.netlify.app/api.json"
```
Iinh, line 15: I think we need to change this back before landing.

scripts/build-glean-metadata.py
```diff
@@ -172,10 +172,11 @@ def get_app_variant_description(app):
 # Write out some metadata for each app group (for the app detail page)
 for (app_name, app_group) in app_groups.items():
     app_dir = os.path.join(OUTPUT_DIRECTORY, app_name)
-    (app_id_dir, app_ping_dir, app_table_dir, app_metrics_dir) = (
-        os.path.join(app_dir, subtype) for subtype in ("app_ids", "pings", "tables", "metrics")
+    (app_id_dir, app_ping_dir, app_table_dir, app_metrics_dir, app_labels_dir) = (
```
Iinh, line 175:
```suggestion
    (app_id_dir, app_ping_dir, app_table_dir, app_metrics_dir) = (
```

scripts/build-glean-metadata.py
```diff
@@ -172,10 +172,11 @@ def get_app_variant_description(app):
...
+        os.path.join(app_dir, subtype)
+        for subtype in ("app_ids", "pings", "tables", "metrics", "labels")
```
Iinh, line 177:
```suggestion
        for subtype in ("app_ids", "pings", "tables", "metrics")
```

scripts/build-glean-metadata.py
```diff
@@ -172,10 +172,11 @@ def get_app_variant_description(app):
...
     )
-    for directory in (app_id_dir, app_ping_dir, app_table_dir, app_metrics_dir):
+    for directory in (app_id_dir, app_ping_dir, app_table_dir, app_metrics_dir, app_labels_dir):
```
Iinh, line 179:
```suggestion
    for directory in (app_id_dir, app_ping_dir, app_table_dir, app_metrics_dir):
```

Iinh (approved):
r+wc

PR #512, opened 2021-04-06
Persist tab state in URL on app detail page, search state everywhere

wlach:
This will pave the way for "search by feature label" (see #511)
but should also be useful on its own.

### Pull Request checklist

      fixes, if applicable)


wlach:
Some demos:

* https://deploy-preview-512--glean-dictionary-dev.netlify.app/apps/fenix/app_ids
* https://deploy-preview-512--glean-dictionary-dev.netlify.app/apps/fenix/metrics?search=addons

src/pages/AppDetail.svelte
```diff
@@ -40,20 +53,39 @@
...
-    <TabContent key="Pings">
-      <ItemList itemType="pings" items={app.pings} appName={app.app_name} />
+    <TabContent key="pings">
+      <ItemList
+        itemType="pings"
+        items={app.pings}
+        appName={app.app_name}
+        filterText={queryString}
+        on:filterTextChanged={({ detail }) => {
+          updateURL(detail.filterText);
```
wlach, line 73: We need to be careful to pass this value up when updating the URL-- if we use the local `queryString` parameter, it will be out of date.

wlach:
Realized more cleanup was required the next day while working on feature annotations, I think we're good now though.

src/pages/AppDetail.svelte
wlach: This turns out not to be required if we use a binding!

src/components/tabs/TabGroup.svelte
```diff
@@ -1,10 +1,13 @@
...
   export let active = 0;
   import { writable } from "svelte/store";
-  import { setContext } from "svelte";
+  import { createEventDispatcher, setContext } from "svelte";
 
   const store = writable(active);
   setContext("activeTab", store);
+
+  const dispatch = createEventDispatcher();
+  store.subscribe((value) => dispatch("tabChanged", { active: value }));
```
Iinh, line 10: This took a while for me to understand. If the goal is to dispatch `tabChanged` when the value of `active` changes, I think we could make `active` a store and set a watcher to dispatch `tabChanged`. If you prefer to keep `store`, let's change it to a more descriptive name rather than a generic `store`.
(unrelated but not sure why the default value of active is 0, isn't it supposed to be a string?)

Iinh (changes requested):
I admit this was kinda hard to follow, but it's to be expected imo as handling state sharing across multiple components is not an easy problem.

From reading the code, I understand the goal is to update the url when the active tab changes, and when `search` changes in AppDetail and TableDetail. There is a clear reactive relationship between these 3 states, and as these states are shared across multiple components, I wonder if it's better to group them in a store object to better manage the relationship, instead of having 3 separate dispatch events that affect data binding. I'd hope that doing so should improve readability and avoid side effects. 

Speaking of side effects, I ran into some strange behaviors.
- `search` not reset properly when switching tabs: If you input some filter text in a tab and then switch to another tab, the item list of the new tab is affected. For example, in [burnham](https://deploy-preview-512--glean-dictionary-dev.netlify.app/apps/burnham/metrics), search for "test" in Metrics tab, then switch to Pings tab, the ping list appears empty. 
- Using back button after a search query: if I search for something, say "test", then use the back button to go back to a previous page. Instead of showing the previous page, it would show all the previous queries: "test", "tes", "te", "t" before directing to the previous page. I think this is quite unintuitive, but I understand that this might be out of scope for this PR.

wlach:
Thanks @Iinh! I could have sworn the issues you mentioned were fixed in the latest revision, but apparently not :( This does indeed seem to need some more work: I think I am going to try using a store here, as you suggest.

wlach:
Getting closer, but still need to work on this more.

wlach:
Hey @Iinh, ok so this turned into something bigger than I originally anticipated, but I think this will set us up for the future. 
So here's a new approach:

1. Search state is always set as context on pages that have a search box (along with various other parameters that people can set)
2. There is top-level persistence handling in App.svelte which uses a "pageState" store, that then propagates down to components. This allows us to define persistence inside each page in a relatively uniform way.
3. We now use the [replaceState API](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState) for modifying the URL, which lets us avoid polluting browser history every time someone changes an option

I think there might be further cleanup we can do here, but I'm pretty happy with where things are at now. Let me know what you think.

src/state/stores.js
```diff
@@ -1,3 +1,4 @@
 import { writable } from "svelte/store";
 
 export const pageTitle = writable("");
+export const pageState = writable({});
```
Iinh, line 4: Out of curiosity, why did you decide to set up an empty store object, instead of populating them with properties and initial values? I'm used to seeing stores being more definitive, such as [this](https://svelte.dev/repl/1305d426ddef44578ae101bb1bc5cc8e?version=3.17.1). I think doing so is helpful because it tells us at first glance what states we're tracking across the app by looking at this store.

```suggestion
export const pageState = writable({
  itemType: "metrics",
  search: "",
  showExpired: "false",
  activeTab: "metrics",
  ...
});;
```

src/App.svelte
```diff
@@ -87,6 +86,22 @@
...
+  // initial state before doing this)
+  $: {
+    // create nice, short urls that don't incorporate falsely values
+    // and convert booleans to integers
+    const query = stringify(
+      mapValues(
+        pickBy($pageState, (v) => (typeof v !== "string" && v) || v.length > 0),
+        (v) => (typeof v === "boolean" ? +v : v)
+      )
+    );
```
Iinh, line 100: The comments are definitely helpful, but I think we should split the logic to smaller functions to improve readability (tbh I'm asking for a selfish reason, my brain can't comprehend too many logic circuits at once lol)

src/components/tabs/TabGroup.svelte
```diff
@@ -1,10 +1,13 @@
...
-  import { setContext } from "svelte";
+  import { createEventDispatcher, setContext } from "svelte";
 
-  const store = writable(active);
-  setContext("activeTab", store);
+  const activeTab = writable(active);
+  setContext("activeTab", activeTab);
+
+  const dispatch = createEventDispatcher();
+  $: dispatch("tabChanged", { active: $activeTab });
```
Iinh, line 10: This seems like an overkill: we're using 3 separate state management systems — store, context, event dispatcher — to handle active tab state. Is it possible to reduce the complexity here?

src/pages/AppDetail.svelte
```diff
@@ -9,12 +12,21 @@
...
 
   export let params;
 
   const appDataPromise = getAppData(params.app);
 
+  let itemType = $pageState.itemType || "metrics";
+  const searchText = writable($pageState.search || "");
+  setContext("searchText", searchText);
+  const showExpired = writable($pageState.showExpired || false);
+  setContext("showExpired", showExpired);
```
Iinh, line 25: If `pageState` store already tracks search, itemType, showExpired, (and maybe more that I'm missing), why is it necessary to create context for these properties? Aren't `searchText` and `$pageState.search` always referring to the same value? 

> There is top-level persistence handling in App.svelte which uses a "pageState" store, that then propagates down to components. This allows us to define persistence inside each page in a relatively uniform way.

If ^ is the underlying principle to handle states in these pages, wouldn't it be simpler to retrieve and update the states directly from the store, instead of creating another access layer of context on top? 

What I'm trying to say is, wherever we need to use state variables, we could just call directly from the store. And same goes for when we need to update them. I feel like using context for a store property is a bit redundant, but I could be missing something major here.

src/App.svelte
```diff
@@ -87,6 +86,22 @@
...
+    const path = `${window.location.pathname}${query ? `?${query}` : ""}`;
+    window.history.replaceState(null, undefined, path);
```
Iinh, line 102: This replaceState method is so cool! Such a nice way of handling url history.

Iinh (changes requested):
This looks great. I tested the app using the deploy preview, and everything seems to work as expected. There is a small thing: using back button after a search query works now, but if I use the forward button after, the search query doesn't seem to reset correctly. 

CleanShot 2021-04-08 at 17 36 59@2x

I was able to follow the code pretty well with this new approach (thank you for providing more context in the comments btw!) I left a few comments and questions where I think the code could be simplified a bit more. Other than that, I think this looks great. Having shareable URLs is an important feature that I think Glean Dictionary is missing.

src/pages/AppDetail.svelte
wlach: I think you're right. I think my reasoning was that we should use context to make the components more reusable, but in practice it just makes things more complicated and these components are pretty context-dependent in any case.

I'll give what you suggest a try.

src/state/stores.js
wlach: Hmm, I wonder if this makes sense given that the applicable pageState depends on the page you're on?

wlach:
> This looks great. I tested the app using the deploy preview, and everything seems to work as expected. There is a small thing: using back button after a search query works now, but if I use the forward button after, the search query doesn't seem to reset correctly.

I think this is actually the desired behaviour 😄 Seems handy to preserve the search when navigating back & forth.

src/App.svelte
wlach: I think keeping it in one function is for the best, but I'll write a comment to say more about what's going on here.

src/state/stores.js
wlach: On reflection, I think we might want to do this, but let's wait a bit until we've added a few more pages (e.g. derived dataset documentation / searching). I think it will be more clear how to approach page state in a principled manner when we have a few more examples to work from. I think the current code is easy enough to understand.

src/App.svelte
wlach: I split out the state handling from the stringify step and added a couple more comments, hopefully that helps.

src/pages/AppDetail.svelte
wlach: Ok, I dug more into this. 

I think, unfortunately, that something like the current approach is probably about the best we can do if we want to continue making the tabs a reusable component (rather than something that is tightly coupled to a notion of displaying an item list in the application page, which is what we're currently using it for).

The most promising place to clean stuff up would be TabGroup.svelte, which uses its own store as well as context. It would seem logical to just pass in a store and have it manipulate that (and it's totally possible), but that makes testing it as a standalone component and forces you to use it as a store, which I think is not necessarily what we'd want in all cases. I gave this a try in but reverted it after realizing that it broke the stories, you can see what it looks like here: https://github.com/mozilla/glean-dictionary/pull/512/commits/a90db1457fc9575a8d678cf12fbd4fe78205f092

wlach:
Hey @Iinh, I think I'd like to go ahead and merge this. I'm sure there are improvements possible but I think this is still a big improvement over what we have now. We can always refine things in the future as we learn more about svelte (I wonder how this would change if we were using sveltekit?).

Iinh (approved):
This is definitely a big improvement over what we have currently!

PR #549, opened 2021-04-26
Fix angle brackets (hopefully for the last time)

wlach:
PR #497 fixed some cases, but not all. Angle brackets in
code blocks have their own escaping behaviour, which doesn't
work well with the adhoc work we're doing.
    
Strictly a string like this:
    
    <provider>
    
is considered HTML in markdown. However this *probably* isn't
how we want to interpret it in the Glean Dictionary. We'll
address this by modifying the renderer when it encounters an
HTML tag, rather than doing a naive regex replace against the
whole markdown string.

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -13,20 +13,19 @@
         start ? `start="${start}"` : ""
       } class="mzp-u-list-styled">${body}</${outerEl}>`;
     },
+    html(html) {
+      return `<code>${html.replace("<", "&lt;")}</code>`;
+    },
   };
   use({ renderer });
 
   export let text;
   // if inline is set, do not wrap the markdown in a paragraph -- useful for short snippets
   export let inline = true;
-
-  // escape `<` so it doesn't get confused as an HTML tag, see:
-  // https://github.com/mozilla/glean-dictionary/pull/497
-  $: htmlText = text.replace("<", "&lt;");
 </script>
 
 {#if inline}
-  {@html parseInline(htmlText)}
+  {@html parseInline(text)}
 {:else}
-  {@html parse(htmlText)}
+  {@html parse(text)}
 {/if}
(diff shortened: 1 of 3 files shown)


wlach:
Example of case which isn't currently handled well: https://dictionary.protosaur.dev/apps/fenix/metrics/metrics_search_count

wlach:
With deploy preview:

https://deploy-preview-549--glean-dictionary-dev.netlify.app/apps/fenix/metrics/metrics_search_count

Also here's the previous one we fixed with #497, still looks good:

https://deploy-preview-549--glean-dictionary-dev.netlify.app/apps/fenix/metrics/browser_search_ad_clicks

wlach:
Because I'm a pedant, I filed an issue for the upstream issue (just this once): https://github.com/mozilla-mobile/fenix/pull/19243

Iinh:
> With deploy preview:

https://deploy-preview-549--glean-dictionary-dev.netlify.app/apps/fenix/metrics/metrics_search_count

I was not able to see the above page with my adblocker on, just filed an issue about this: https://github.com/mozilla/glean-dictionary/issues/550

wlach:
I think that's a seperate issue from this 😄 @Iinh would you mind reviewing this one too?

src/components/Markdown.svelte
```diff
@@ -13,20 +13,19 @@
         start ? `start="${start}"` : ""
       } class="mzp-u-list-styled">${body}</${outerEl}>`;
     },
+    html(html) {
+      return `<code>${html.replace("<", "&lt;")}</code>`;
+    },
```
Iinh, line 18: Along with replacing `<`, should we also replace `>`?
```
  const mapChar = {
    "<": "&lt;",
    ">": "&gt;",
  };
```

```suggestion
    html(html) {
      return `<code>${html.replace(/<|>/g, (char) => mapChar[char])}</code>`;
    },
```

Iinh (approved):
Looks good, and I'm optimistic that this will be the last fix 😄

src/components/Markdown.svelte
wlach: I'm pretty sure this isn't necessary. `>` by itself should parse fine in almost all cases: there might be one or two cases where weird things might happen, but I think they're unlikely enough that I'd rather keep this code simple.

Issue #537, opened 2021-04-22
Update deploy / CI logic for production


wlach:
Shortly we'll be [deploying a production version of the glean dictionary](https://bugzilla.mozilla.org/show_bug.cgi?id=1704527) at dictionary.telemetry.mozilla.org. A couple of follow-on tasks to do inside the glean dictionary for that to be smooth:

* [x] Configure site logic via netlify.toml rather than using the netlify deploy panel (easier to audit)
* [x] Stop deploying dev site to dictionary.protosaur.dev (to minimize confusion, we'll turn that into a set of redirects to dictionary.tmo, and use the dev version of the site at glean-dictionary-dev.netlify.app for keeping track of the state of the main branch)

wlach:
Let's also do a redirect for https://glean-dictionary.netlify.app/ (legacy site from last summer which was never updated)

wlach:
> Let's also do a redirect for https://glean-dictionary.netlify.app/ (legacy site from last summer which was never updated)

This is done now.

PR #551, opened 2021-04-26
Configure netlify via netlify.toml

wlach:
This is part of what we need to do to prepare for a production deploy (#537)

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,19 @@
+[build]
+command = """
+    scripts/build-glean-metadata.py
+    npm run build
+    npm run build-storybook
+    mv storybook-static public/storybook
+"""
+environment = { PYTHON_VERSION = "3.7" }
+publish = "public"
+
+[context.production]
+  # We don't build storybook in production
+  [context.production.build]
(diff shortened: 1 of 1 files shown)


wlach:
@robotblake You may want to unset the `Build command` and `Publish directory` settings in the netlify gui for dictionary.tmo, as the values in this file will now take precedence and it might be confusing to people in the future. Same with the `PYTHON_VERSION` environment variable.

netlify.toml
```diff
@@ -0,0 +1,15 @@
+[build]
+command = """
+    scripts/build-glean-metadata.py
+    npm run build
+    if [ "$STORYBOOK" ]; then
```
wlach, line 5: For `glean-dictionary-dev` (only), I've set this variable to `1` in the GUI. This seems like the most straightforward way of getting the right behaviour for our use case (multiple netlify "apps" for the same repo)

robotblake (approved):
I've gone ahead and removed the config from the Netlify Web UI. LGTM! 👍

PR #555, opened 2021-04-27
Bypass cache when accessing probeinfo.telemetry.mozilla.org

wlach:
In discussion on #glean-dictionary, there's some question about what will disable cloudfront's cache. But per https://aws.amazon.com/premiumsupport/knowledge-center/cloudfront-custom-object-caching/:

> Review the Cache-Control header in the response. If the value for Cache-Control is "no-store," then the header is directing CloudFront to not cache the response. If the value for Cache-Control is "no-cache," then the header is directing CloudFront to verify with the origin before returning a cached response. These directives supersede the Maximum TTL and Default TTL CloudFront settings.

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 autoflake
 black
-cachecontrol
 flake8
 flake8-black
 isort
--- a/scripts/glean.py
+++ b/scripts/glean.py
@@ -4,7 +4,6 @@ import logging
 from datetime import datetime
 from typing import List
 
-import cachecontrol
 import requests
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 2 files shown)


fbertsch (changes requested):
Unfortunately those docs are for response headers, not request headers. See [here](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/header-caching.html#header-caching-web) for request caching. Particularly:

```
By default, CloudFront doesn't consider headers when caching your objects in edge locations. If your origin returns two objects and they differ only by the values in the request headers, CloudFront caches only one version of the object. 
```

wlach:
Grump, I can't believe I missed that. Back to the drawing board.

fbertsch:
> Grump, I can't believe I missed that. Back to the drawing board.

Why not utilize the [existing method](https://github.com/mozilla/mozilla-schema-generator/blob/main/mozilla_schema_generator/generic_ping.py#L215-L218) we currently use?

wlach:
> > Grump, I can't believe I missed that. Back to the drawing board.
> 
> Why not utilize the [existing method](https://github.com/mozilla/mozilla-schema-generator/blob/main/mozilla_schema_generator/generic_ping.py#L215-L218) we currently use?

Right, that's the correct solution, we just need to take care implementing it: I want to be able to do some local caching to make sure that we're not fetching the same resource more than once. I think we can get away with just doing away with cachecontrol and creating our own naive dict-based cache.

fbertsch (approved):
LGTM!

PR #571, opened 2021-05-07
Link to Looker ping explores

wlach:
This is a preliminary implementation and is a bit of a hack. We can
flesh this out later as we deepen our integration with Looker.

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/state/urls.js
+++ b/src/state/urls.js
@@ -16,3 +16,17 @@ export function getBigQueryURL(appName, appId, pingName, metricName) {
 
   return base + (metricName ? `?search=${metricName}` : "");
 }
+
+export function getLookerURL(appName, pingName) {
+  // this is currently only supported for a hardcoded set of app ids and is a bit of a
+  // hack (we should ideally be doing this somewhere in our upstream ETL)
+
+  if (appName === "fenix" || appName === "burnham") {
+    return `https://mozilladev.cloud.looker.com/explore/${appName}/${pingName.replace(
+      /-/g,
+      "_"
+    )}`;
+  }
+
+  return undefined;
+}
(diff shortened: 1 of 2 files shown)


src/state/urls.js
```diff
@@ -16,3 +16,17 @@ export function getBigQueryURL(appName, appId, pingName, metricName) {
 
   return base + (metricName ? `?search=${metricName}` : "");
 }
+
+export function getLookerURL(appName, pingName) {
+  // this is currently only supported for a hardcoded set of app ids and is a bit of a
+  // hack (we should ideally be doing this somewhere in our upstream ETL)
+
+  if (appName === "fenix" || appName === "burnham") {
+    return `https://mozilladev.cloud.looker.com/explore/${appName}/${pingName.replace(
```
wlach, line 25: Should change this when fenix lands on production.

Iinh (approved):

src/state/urls.js
Iinh: I don't have access to this dev site. is it restricted to only the people who are working on Looker?

wlach: Yes, though that will almost certainly include you in the near future-- I suspect we should add you to the list of developers for Looker.

fbertsch: This should be prod now.

fbertsch (approved):
r+wc
Can we do the same for ios, and the `unified_metrics` explore?

We should be using `namespaces.yaml` for this. We can start with https://bugzilla.mozilla.org/show_bug.cgi?id=1707727, and then move to reading that and using it to know about explores in Looker.

wlach:
>  Can we do the same for ios, and the `unified_metrics` explore?
> 
> We should be using `namespaces.yaml` for this. We can start with https://bugzilla.mozilla.org/show_bug.cgi?id=1707727, and then move to reading that and using it to know about explores in Looker.

Yes to both of these things. Let's see if we can do this the "right way" next week. 😄

PR #586, opened 2021-05-12
Upgrade Protocol

Iinh:
With the latest Protocol release, we can now use sass instead of node-sass, which is already deprecated. 

### Pull Request checklist

      fixes, if applicable)


src/App.svelte
```diff
@@ -167,7 +167,7 @@
     {/if}
   </header>
   <main>
-    <div class="mzp-l-content">
+    <div class="mzp-l-content" style="padding-top: 15px">
```
Iinh, line 170: they changed the container padding in the latest version of Protocol, so we need to do inline styling here to override it or we'd have a lot of extra padding, like the screenshot below.

Screen Shot 2021-05-12 at 3 50 06 PM

With SvelteKit I'm only importing what we need to make it easier for customization like this (and less inline styling).

wlach: Awesome, sounds like a sensible compromise for now!

wlach (approved):

PR #608, opened 2021-05-21
Make it more clear that GLAM and Looker require LDAP

wlach:
Do this by creating an "authenticated link" class which adds a lock
icon next to the resources. This is the same approach we're using
with Glean Annotations.

### DIFF
--- a/.storybook/story.css
+++ b/.storybook/story.css
@@ -1,0 +1,1 @@
+@import "tippy.js/dist/tippy.css";
--- a/.storybook/story.css
+++ b/.storybook/story.css
@@ -1,1 +1,0 @@
-@import "tippy.js/dist/tippy.css";
--- a/src/components/AuthenticatedLink.svelte
+++ b/src/components/AuthenticatedLink.svelte
@@ -1,0 +1,18 @@
+<script>
+  import tippy from "./tippy";
+
+  export let href;
+</script>
+
+<style>
+  .lock-link {
+    text-decoration: none;
+  }
+</style>
+
+<a {href}><slot /></a>
+<a
+  class="lock-link"
+  href="https://docs.telemetry.mozilla.org/concepts/gaining_access.html"
+  use:tippy={{ content: "Access to this resource is limited to NDA'd Mozillians", placement: 'top' }}>
+  🔒</a>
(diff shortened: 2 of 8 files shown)


wlach:
Example:

image
https://deploy-preview-608--glean-dictionary-dev.netlify.app/apps/fenix/pings/activation

.storybook/story.css
```diff
@@ -0,0 +1 @@
+@import "tippy.js/dist/tippy.css";
```
Iinh, line 1: I don't think we need to do this to make tippy work in storybook, we already imported tippy.css with css loader in `.storybook/preview.js`. Also seems like this file is not imported to be used anywhere? It was an empty file before so I'm assuming it's safe to just delete it.

src/components/AuthenticatedLink.svelte
```diff
@@ -0,0 +1,18 @@
...
+  .lock-link {
+    text-decoration: none;
+  }
+</style>
+
+<a {href}><slot /></a>
+<a
+  class="lock-link"
+  href="https://docs.telemetry.mozilla.org/concepts/gaining_access.html"
+  use:tippy={{ content: "Access to this resource is limited to NDA'd Mozillians", placement: 'top' }}>
```
Iinh, line 17:
```suggestion
  use:tippy={{ content: "Access to this resource is limited to NDA'd Mozillians", placement: 'right' }}>
```
Let's change `placement` to right so that the tooltip hover doesn't cover the bigquery link above it.

CleanShot 2021-05-21 at 09 31 41@2x

Iinh (approved):
Other than the small nits that I commented this looks good. Thanks @wlach!

.storybook/story.css
wlach: You're right! I think we can get rid of this. I just put this in while I was trying to get things to work.

src/components/AuthenticatedLink.svelte
wlach: Hmm, I'm not sure -- this would be inconsistent with the behaviour of other on-hover tooltips (and tooltip behaviour in most apps) and I'm not sure if it's *too* much of a problem if the link is covered since you can just unhover the lock to see the link above. Do you feel strongly about this?

Iinh: I think a tooltip should not cover the object that it is attached to, which in this case it doesn't, so I don't feel strongly about this :) Tbh I never paid much attention to the tooltip placement so I wasn't aware that the default placement should be on top. Thanks for letting me know!

PR #610, opened 2021-05-21
a11y lock

wlach:
Use an a11y-friendly approach for the lock icon used to show limited access


wlach:
Sorry about the noise on this, I kind of realized after the fact that what I was doing wasn't great from an accessibility standpoint (something I'm trying to get better at).

Iinh (approved):
I should have caught this when reviewing it earlier. Thanks for making this change!

PR #619, opened 2021-05-24
persist pagination state in url (#528)

### LINKED ISSUE #528: Persist pagination state in URL
Just as we persist search strings and expired states (see #512), we should also persist the page that the user has navigated to (if any):

https://dictionary.protosaur.dev/apps/fenix?itemType=metrics&showExpired=1&page=3

This is a bit of a tricky one (it requires understanding the Glean Dictionary's URL persistence model, which is based on [svelte stores](https://svelte.dev/tutorial/writable-stores)) but might make an interesting issue for a motivated contributor to work on. Feel free to ping us in the channel if interested.

harnaman-hk:
Fixes #528 
The thought behind this PR is as follows:
1. Update pagination state if the user navigates to any
2. reset pagination state to 1 when input filter is changed, i.e. display the first page by default
3. Clear the pagination state from url if user jumps to some other page

Also, when un-checking the paginate checkbox on pages 2, 3, .. (any page other than 1), the app ran into `Uncaught (in promise) Error: {#each} only iterates over array-like objects`. This commit fixes this error as well.

Pull Request checklist

fixes, if applicable)

### DIFF
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -1,6 +1,4 @@
 <script>
-  import { setContext } from "svelte";
-  import { writable } from "svelte/store";
   import { chunk } from "lodash";
 
   import { getItemURL } from "../state/urls";
@@ -25,16 +23,20 @@
   let pagedItems;
   let paginated = true;
 
-  let currentPage = writable(1);
-  setContext("currentPage", currentPage);
+  // track change of filter input
+  let { search } = $pageState;
 
   // re-filter items when showExpired or search text changes
   $: {
-    const search = $pageState.search || "";
-    // don't use $currentPage, because we don't want to call this reactively
-    // when paginating
-    currentPage.set(1);
+    if (search !== $pageState.search) {
+      search = $pageState.search || "";
+      $pageState.page = 1;
+    }
+  }
 
+  // update pagedItems when either pagination changes or search text changes
+  // (above)
+  $: {
     const originMatch = (item) =>
       item.origin && item.origin.includes(search.toLowerCase());
 
@@ -47,15 +49,14 @@
     filteredItems = $pageState.showExpired
       ? filteredItems
       : filteredItems.filter((item) => !isExpired(item.expires));
-  }
 
-  // update pagination when either pagination changes or filtered list changes
-  // (above)
-  $: {
+    // update pagination
+    const currentPage = $pageState.page || 1;
     const perPage = paginated ? DEFAULT_ITEMS_PER_PAGE : filteredItems.length;
+    const chunkIndex = paginated ? currentPage - 1 : 0;
     pagedItems =
       filteredItems.length > 0
-        ? chunk([...filteredItems], perPage)[$currentPage - 1]
+        ? chunk([...filteredItems], perPage)[chunkIndex]
         : [];
   }
 
--- a/src/components/Pagination.svelte
+++ b/src/components/Pagination.svelte
@@ -1,5 +1,5 @@
 <script>
-  import { getContext } from "svelte";
+  import { pageState } from "../state/stores";
 
   import BackButton from "./icons/BackButton.svelte";
   import ForwardButton from "./icons/ForwardButton.svelte";
@@ -9,7 +9,7 @@
   let from;
   let to;
   let lastPage;
-  let currentPage = getContext("currentPage");
+  let currentPage = 1;
 
   const truncatedPagination = (current, last) => {
     // Source: https://gist.github.com/kottenator/9d936eb3e4e3c3e02598#gistcomment-3413141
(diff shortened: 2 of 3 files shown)


src/components/ItemList.svelte
```diff
@@ -28,39 +28,46 @@
...
     pagedItems =
       filteredItems.length > 0
-        ? chunk([...filteredItems], perPage)[$currentPage - 1]
+        ? chunk([...filteredItems], perPage)[chunkIndex]
         : [];
   }
 
   const originClicked = (origin) => {
-    $pageState = { ...$pageState, search: origin };
+    $pageState = { ...$pageState, page: origin, search: origin };
```
Iinh, line 70: Hmm this doesn't look right. `origin` in this case is the Glean label while page is supposed to be our current page number, which is already set to 1 when search changes in the reactive `$` block above.

```suggestion
    $pageState = { ...$pageState, search: origin };
```

Iinh (changes requested):
Hi @harnaman-hk, this looks like a great start! I tested using our [deploy preview](https://deploy-preview-619--glean-dictionary-dev.netlify.app/) and was able to see the pagination state updated in the URL when I changed pages 👍 However, when I refreshed the page, `$pageState` did not persist. 

E.g.:
https://deploy-preview-619--glean-dictionary-dev.netlify.app/apps/burnham?itemType=metrics&page=2&search=glean-core&showExpired=1

Looking at the above link I would assume that I'm going to see page 2 of `glean-core` metrics. But instead we get all metrics, and the page number was reset to 1.

You can see the correct behavior of persisting state in our production site (without pagination state): https://dictionary.telemetry.mozilla.org/apps/burnham?itemType=metrics&search=glean-core&showExpired=1

We should try to make pagination state persist the same way too. I suspect that this can be fixed by setting `search` and `$currentPage` to match `$pageState` on initial load. Wondering if @wlach has additional input? 

Thank you for working on this! Let us know in the [#glean-dictionary](https://chat.mozilla.org/#/room/#glean-dictionary:mozilla.org) channel if you have any questions :)

wlach:
> We should try to make pagination state persist the same way too. I suspect that this can be fixed by setting `search` and `$currentPage` to match `$pageState` on initial load. Wondering if @wlach has additional input?

Nope! That sounds exactly right to me. Looks like a great PR, looking forward to seeing some updates.

harnaman-hk:
Right, so since we are persisting pagination state in the url, we should not need to do getContext in [pagination.svelte](https://github.com/mozilla/glean-dictionary/blob/main/src/components/Pagination.svelte) for communication between the components. I refactored the code to exclude getContext. 
Also, I refactored the pagedItems updates in itemList. As it depends both on search text and pagination state, it can be done in one block.
Page refresh and pagination work in sync now.

Looking forward to your comments @Iinh , @wlach  😃

src/components/Pagination.svelte
```diff
@@ -101,22 +106,22 @@
 {#if itemsPerPage < totalItems}
   <div class="pagination-bar">
     <div
-      on:click|preventDefault={() => changePage($currentPage !== 1 ? $currentPage - 1 : 1)}
-      class="pagination-button {$currentPage === 1 ? 'current-page' : ''}">
+      on:click|preventDefault={() => changePage(currentPage !== 1 ? currentPage - 1 : 1)}
+      class="pagination-button {currentPage == 1 ? 'current-page' : ''}">
```
harnaman-hk, line 110: I wonder why == is giving the correct result instead of ===, as is the svelte syntax. I observed this by accident.

wlach:
Seems to work pretty well to me in my initial testing, but I'll leave it to @Iinh to do the final review. The linter issues will need to be fixed before this lands though.

src/components/Pagination.svelte
harnaman-hk: Oh okay, then it must be due to different data types. I'll update this.

wlach: Ah that makes sense, yes. You might need to cast the value to integer before doing the comparison

src/components/ItemList.svelte
```diff
@@ -25,16 +23,20 @@
   let pagedItems;
   let paginated = true;
 
-  let currentPage = writable(1);
-  setContext("currentPage", currentPage);
+  // track change of filter input
+  let search = $pageState.search;
```
Iinh, line 27: ESLint wants us to use object destructuring here:
```suggestion
  let search = { $pageState };
```

Iinh (changes requested):
Hi, this looks great. A couple of issues to resolve before this is ready:

- I was testing around and saw that page state was not reset properly. For example if I'm on [page 3 of `engine-gecko` search results](https://deploy-preview-619--glean-dictionary-dev.netlify.app/apps/fenix?itemType=metrics&page=3&search=engine-gecko&showExpired=1), then I switch tab to view the pings, the app breaks. This is due to `$pageState.page` was not reset properly, and I believe the code that causes this is in `AppDetail.svelte`.

- Since I'm seeing a merge conflict, you should rebase your branch onto our `main` branch. You need to add a remote upstream first: `git remote add upstream https://github.com/mozilla/glean-dictionary.git`. Then do `git pull --rebase upstream main`. Keep it mind will ask you to solve any existing merge conflicts before the rebase can complete. 

I know @wlach has a git cheatsheet that helped me a lot when I first started working on Glean Dictionary. If you're interested you could join our chat room and DM them for it :)

src/pages/AppDetail.svelte
```diff
@@ -58,7 +59,12 @@
   <TabGroup
     active={$pageState.itemType}
     on:tabChanged={({ detail }) => {
-      pageState.set({ ...$pageState, itemType: detail.active, search: "" });
+      pageState.set({
+        ...$pageState,
+        itemType: detail.active,
+        search: "",
+        page: "",
+      });
```
Iinh, line 67: Small nit: let's try to keep state resetting similar to the initial default state that we set above.

```suggestion
      pageState.set({
        ...$pageState,
        itemType: detail.active,
        search: "",
        page: 1,
      });
```

Iinh (approved):
This looks really good! One tiny nit left and then this is ready to land. Our Pagination component is quite complex to work with, so thank you for your patience and persistence in working on this issue!

Iinh:
Great work @harnaman-hk 🎉

PR #624, opened 2021-05-27
Refactor the way we represent some internal data

wlach:
* Avoid overly verbose "bigquery_names" dictionary (which deals with
  some things outside of BigQuery already, and will more in the future)
* Build and handle GLAM at the ETL stage rather than in the dictionary
  itself. This helps keep the complexity inside the dictionary down.

### DIFF
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -14,6 +14,12 @@ OUTPUT_DIRECTORY = os.path.join("public", "data")
 ANNOTATIONS_URL = "https://mozilla.github.io/glean-annotations/api.json"
 NAMESPACES_URL = "https://raw.githubusercontent.com/mozilla/looker-hub/main/namespaces.yaml"
 
+GLAM_PRODUCT_MAPPINGS = {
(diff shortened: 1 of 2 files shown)


scripts/build-glean-metadata.py
```diff
@@ -183,23 +189,29 @@ def etl_snake_case(line: str) -> str:
...
+                ),
             )
+            if metric_type != "event" and GLAM_PRODUCT_MAPPINGS.get(app.app_id):
+                (glam_product, glam_app_id) = GLAM_PRODUCT_MAPPINGS[app.app_id]
+                glam_metric_id = etl_snake_case(metric.identifier)
+                etl["glam_url"] = (
+                    "https://glam.telemetry.mozilla.org/"
+                    + f"{glam_product}/probe/{glam_metric_id}/explore"
+                    + f"?app_id={glam_app_id}"
+                )
```
wlach, line 207: Writing it this way partly to appease flake8, though I think it might be a little easier to read in any case.

Iinh (approved):

PR #626, opened 2021-05-28
Improve display of labels (fixes #602)

### LINKED ISSUE #602: Improve display of labels
We're still using a pretty primitive svelte component we've called "Pill" for rendering labels in the app:

image

https://github.com/mozilla/glean-dictionary/blob/9f27d440ceb6977997dd6fc87a7e0dd067a5b33e/src/components/Pill.svelte

It would be nice to improve the look of them to be more consistent with the protocol library we're using. I actually filed an issue about this here: https://github.com/mozilla/protocol/issues/689 -- but am also open to working on a Glean Dictionary implementation specifically if it looks like doing it upstream doesn't make sense for one reason or another.

Iinh:
Storybook for Label.svelte:
CleanShot 2021-05-28 at 08 28 44@2x

Some notes:
I wanted to find a color palette that is color-blind friendly. [This article](https://personal.sron.nl/~pault/#sec:qualitative) recommends a few different palettes, and I decided to go with the light colors as I find them fit the best with Protocol. The pill design is largely inspired by Github labels.

To make this component more reusable, such that the same text will get the same color, I generate the label color using an A1Z26 cipher. The text is first turned into a number from 1-9, then from this number we can get a color from the color palette map.

However, if for some reason we want to use a specific color for the pill, we can override the above method by setting a `labelNumber` number. I did this with the `deprecated` and `expired` labels.

I'm not sure if this approach is more complicated than necessary, and I'm open to feedback if there's any way we can improve it :)

### Pull Request checklist

      fixes, if applicable)


wlach:
@hamilton you have some expertise in this area, could you comment on the a11y side of things?

src/components/Label.svelte
```diff
@@ -0,0 +1,93 @@
...
+      background-color: $label-color-str;
+      color: set-text-color($label-color-str);
+    }
+  }
+
+  .label {
+    padding: 0;
+    display: inline-block;
+    line-height: 20px;
+    font-size: 14px;
```
hamilton, line 78:
```suggestion
    font-size: 12px;
```

hamilton: This gives the padding a bit more room to breath like the Github labels

src/components/Label.svelte
```diff
@@ -0,0 +1,93 @@
...
+    font-weight: 600;
+    border-radius: $spacing-sm;
+    background: silver;
+    padding: 0 5px;
```
hamilton, line 82:
```suggestion
    padding: 0 9px;
```

Since the font used doesn't support `500` which would probably be a bit easier to scan at a smaller text size, a bit of additional padding might help the thing breath a bit more

hamilton (commented):
@wlach asked me to take a look. From an accessibility standpoint, it looks like `label-5` has contrast issues with white text. It may be somewhat easier to not allow people to mix and match text + label with this in mind. I don't see any contrast issues with the other ones.

Screen Shot 2021-05-28 at 9 10 55 AM

I left some take-or-leave suggestions as well.

src/__snapshots__/storyshots.test.js.snap
```diff
@@ -1735,6 +1735,155 @@ exports[`Storyshots ItemList Default 1`] = `
 </section>
 `;
 
+exports[`Storyshots Label Default 1`] = `
```
wlach, line 1738: Love this

wlach (approved):
This looks great, I defer to @hamilton's suggestions on the tweaks.

I think we may want to define more "types" of labels in the future (ala what you did with expired and deprecated), but I think this implementation is more than good enough to start with.

Iinh:
Thanks @hamilton and @wlach for reviewing this PR. I fixed the contrast and padding issue, here are the updated labels:

CleanShot 2021-05-28 at 10 39 03@2x

@hamilton If you don't mind could you share the plugin that you used to catch that contrast issue? I couldn't search for the same one in the Firefox add-on store. Thanks :)

src/components/Label.svelte
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+  .label {
+    padding: 0;
+    display: inline-block;
+    line-height: 20px;
+    font-size: 12px;
+    font-weight: 600;
+    border-radius: $spacing-sm;
```
hamilton, line 80:
```suggestion
    border-radius: 2em;
```

Forgot this one, seeing your screenshot. Take or leave!

hamilton:
@linh my pleasure! It's actually built-in to Firefox devtools – look for "Accessibility" in the tabs.

src/components/Label.svelte
Iinh: taken! Great tip, thanks

Iinh: image

PR #638, opened 2021-05-31
Add a ping selector for "access"

wlach:
This narrows down the set of Looker and BigQuery options considerably.
By default, we show the `metrics` ping first, because that's the one
that people are usually most interested in (at least for Android and
iOS applications).


wlach:
Example:

image

https://deploy-preview-638--glean-dictionary-dev.netlify.app/apps/fenix/metrics/app_build

I tested this with a variety of other scenarios (e.g. an app like bergamot, where there is only one channel and/or a metric which only appears in one type of ping) and it seems to work well there too.

wlach:
GLAM is a bit of a confusing special case for this model because it doesn't seem to offer ping selection for Fenix. We probably need to figure this out eventually but I think this solution (just provide a link to GLAM if any ping has the metric) is good enough for now. /cc @robhudson @acmiyaguchi

robhudson (approved):
All looks great to me!

PR #644, opened 2021-06-02
Deduplicate metrics (fixes #627)

### LINKED ISSUE #627: Need to deduplicate metrics
Sometimes libraries have overlapping metrics if you look into the past. For example, `app_build` is present in `glean-core` and *was* present in `glean-android`). This can result in some confusing behaviour in the metric page because we wind up counting the metric twice for each variant:

image

The fix is to add some logic here to only take the latest version of a metric if we notice it occuring more than once when adding the dependency metrics:

https://github.com/mozilla/glean-dictionary/blob/d33ed7a8bfe67bcbf7be6338cc51615493e60dd8/scripts/glean.py#L205

Iinh:
Fixes #627. For example [`app_build`](https://deploy-preview-644--glean-dictionary-dev.netlify.app/apps/fenix/metrics/app_build) is now counted correctly: 

CleanShot 2021-06-03 at 10 31 46@2x

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/scripts/glean.py
+++ b/scripts/glean.py
@@ -196,6 +196,31 @@ class GleanApp(object):
         logging.info(f"For {self.app_id}, found Glean dependencies: {dependencies}")
         return dependencies
 
+    def deduplicate_metrics(self, metrics):
+        metric_names = set()
+        results = []
+
+        def get_index(name, list):
+            for i in range(len(list)):
+                if list[i][0] == name:
+                    return i
+            return None
+
+        def get_last_date(metric):
+            return metric[1]["history"][-1]["dates"]["last"]
+
+        for item in metrics:
+            if item[0] not in metric_names:
+                metric_names.add(item[0])
+                results.append(item)
+            # get the latest version of a metric if there's a duplicate
+            if item[0] in metric_names and get_last_date(item) > get_last_date(
+                results[get_index(item[0], results)]
+            ):
+                del results[get_index(item[0], results)]
+                results.append(item)
+        return results
+
     def get_metrics(self) -> List[GleanMetric]:
         data = _cache.get_json(GleanApp.METRICS_URL_TEMPLATE.format(self.app["v1_name"]))
         metrics = [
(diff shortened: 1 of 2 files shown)


wlach:
I think it would be better to do the deduplication a little earlier in the process, probably in this method that I linked to from the issue: https://github.com/mozilla/glean-dictionary/blob/d33ed7a8bfe67bcbf7be6338cc51615493e60dd8/scripts/glean.py#L199

scripts/glean.py
```diff
@@ -196,6 +196,31 @@ def get_dependencies(self):
...
+            if item[0] not in metric_names:
+                metric_names.add(item[0])
+                results.append(item)
+            # get the latest version of a metric if there's a duplicate
+            if item[0] in metric_names and get_last_date(item) > get_last_date(
+                results[get_index(item[0], results)]
+            ):
+                del results[get_index(item[0], results)]
+                results.append(item)
+        return results
```
wlach, line 222: This looks the right idea, however I'd suggest a slightly different implementation which I think will be both shorter and faster: basically create a dictionary from metric names -> metrics iteratively, then just get the values at the end. Here's a simple example of the approach:

```py
x = [
    {
    'id': 1,
    'date': 1
    },
    {
    'id': 1,
    'date': 2
    },
    {
    'id': 1,
    'date': 0
    }
]
y = {}
for x1 in x:
    if not y.get(x1['id']) or y[x1['id']]['date'] < x1['date']:
            y[x1['id']] = x1

y.values()    
```
https://alpha.iodide.io/notebooks/7408/

wlach (changes requested):
Looks good! I have some suggestions on how to improve the implementation

scripts/glean.py
Iinh: this is a much simpler approach! thank you!

scripts/glean.py
```diff
@@ -211,11 +211,20 @@ def get_metrics(self) -> List[GleanMetric]:
                 (d[0], {**d[1], "origin": dependency["library_name"]})
                 for d in dependency_metrics.items()
             ]
+            # only get the latest version of metrics
```
wlach, line 214: This should happen outside the for loop (it only needs to be run once, at the end) - maybe actually around line 226 would be better (just before the loop). I'd probably also give it another name, like `metric_map` or something (since it has a different structure than `metrics`, above).

wlach (changes requested):
Just a few small tweaks and we'll be good to land

wlach (approved):
💯

PR #662, opened 2021-06-08
Add platform logos to breadcrumb navigation

### LINKED ISSUE #659: Can be hard to tell offhand what product a metric or ping pertains to
Had one piece of feedback - it’s really hard, in general, to keep track of what probes exist for iOS vs Fenix glean… and using the glean dictionary for each, i had to keep looking up to the top banner to keep track of which product i was in, for a page (apps/fenix/etc.)
this was really time, and attention intensive.

it would be really great to encode which product we’re in (for like a probe, or search page) visually in some way that’s automatic and unconscious, so users can know what product they’re in almost instantly.

Some suggestions:
* have different background colors for Fenix vs iOS pages (so which product we’re in is color encoded in the background)
* have a giant logo in the white space of the product (not sure if this will work though, since the logo is the same…)

not sure how hard it is to implement… I’m a big fan of the first idea (background color encoding)

Iinh:
Fixes #659. 

Apple: https://deploy-preview-662--glean-dictionary-dev.netlify.app/apps/focus_ios?itemType=metrics&page=1&showExpired=1
CleanShot 2021-06-08 at 16 53 08@2x

Android: https://deploy-preview-662--glean-dictionary-dev.netlify.app/apps/fenix?itemType=metrics&page=1&showExpired=1
CleanShot 2021-06-08 at 16 51 31@2x

Amazon: https://deploy-preview-662--glean-dictionary-dev.netlify.app/apps/firefox_fire_tv?itemType=metrics&page=1&showExpired=1
CleanShot 2021-06-08 at 16 54 34@2x

### Pull Request checklist

      fixes, if applicable)


src/components/Breadcrumb.svelte
```diff
@@ -1,31 +1,65 @@
...
-      >
-    </li>
-  {/each}
-</ol>
+<div class="breadcrumb">
+  <ol>
+    {#each links as link}
+      <li>
+        <a class="link-name" href={link.url}>{link.name} </a>
+        <span>{links.indexOf(link) === links.length - 1 ? "" : "->"}</span>
```
Iinh, line 30: I changed to using the arrow "-->" instead of "/" here because I find it to look slightly cleaner. Happy to revert it though if it's better to use the original.

src/App.svelte
```diff
@@ -108,25 +108,17 @@
...
-      <div class="mzp-c-navigation-l-content">
-        <div class="mzp-c-navigation-container">
-          <div class="mzp-c-navigation-logo-glean">
-            <a class="glean-logo" href="/">
-              <img src="/glean_logo.png" alt="Glean Dictionary Logo" /></a
-            >
-            <a href="/"><h5>Dictionary</h5></a>
-          </div>
-        </div>
+      <div class="mzp-c-navigation-logo-glean">
```
Iinh, line 111: I cleaned up the redundant navigation divs in here and the breadcrumb below.

src/components/Breadcrumb.svelte
wlach: I like it! Let's keep it.

src/components/Breadcrumb.svelte
```diff
@@ -1,31 +1,65 @@
...
+    if (app.includes("fenix") || app.includes("android"))
+      return "/img/app-logos/android-breadcrumb.png";
+    if (
+      app.includes("amazon") ||
+      app.includes("echo") ||
+      app.includes("fire_tv")
+    )
+      return "/img/app-logos/amazon-breadcrumb.png";
+    return undefined;
+  };
```
wlach, line 22: We should eventually clean this up so the metadata is defined by the application itself or an annotation. That can wait though.

wlach (changes requested):
Looks great, my only feedback is that we should probably always put the platform identifier after the application, rather than at the end of the breadcrumb if you're looking at a metric, ping, etc. Right now it looks like the android logo pertains to the metric:

image

Also, kind of random, but what do you think about reducing the top and bottom padding on the breadcrumb while we're here? Maybe reduce it to 75% of what it is or something. I think it's a bit excessive right now.

src/App.svelte
```diff
@@ -172,7 +165,7 @@
     }
   }
 
-  .c-sub-navigation {
+  .breadcrumb {
```
wlach, line 168: 👍

wlach (approved):

wlach:
Thanks for this!

PR #668, opened 2021-06-10
Handle label edge case

wlach:
Some labels' styling wouldn't match one of the ones that we
generated, so we'd get a class that wasn't defined (label-0).
Fix this and simplify some logic while we're at it.

### DIFF
--- a/src/components/Label.svelte
+++ b/src/components/Label.svelte
@@ -16,20 +16,7 @@
     return num;
   };
 
-  const digitSum = (number) => {
-    let sum = 0;
-    let num = number;
-    while (num % 10 !== 0) {
-      sum += Math.floor(num % 10);
-      num /= 10;
-      if (num === 0 && sum >= 10) {
-        num = sum;
-        sum = 0;
-      }
-    }
-    return sum;
-  };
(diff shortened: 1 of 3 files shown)


src/components/Label.svelte
```diff
@@ -16,20 +16,7 @@
...
-      num /= 10;
-      if (num === 0 && sum >= 10) {
-        num = sum;
-        sum = 0;
-      }
-    }
-    return sum;
-  };
-  const getLabelNumber = (word) => digitSum(a1z26Decoder(word));
+  const getLabelNumber = (word) => a1z26Decoder(word) % 10;
```
Iinh, line 19: Seeing how the chunk of logic on the left is simplified to one single line is so satisfying

Iinh (approved):

src/components/Label.svelte
wlach: Sorry I didn't catch this in the original review! I really should have looked at the code a little more closely. The [modulus operator is indeed pretty useful for all sorts of things](https://blog.mattclemente.com/2019/07/12/modulus-operator-modulo-operation.html). E.g. It's used in our [sampling algorithm for telemetry data](https://docs.telemetry.mozilla.org/concepts/sample_id.html).

PR #678, opened 2021-06-14
Scroll item list back to top when clicking on labels

wlach:
Before, when clicking on the item, the scroll position would be
maintained, which was a bit disconcerting.

### DIFF
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -22,6 +22,8 @@
   let filteredItems = items.filter((item) => !isExpired(item.expires));
   let pagedItems;
   let paginated = true;
+  let topElement;
+  let scrollY;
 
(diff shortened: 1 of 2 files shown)


src/components/ItemList.svelte
```diff
@@ -66,116 +68,128 @@
...
-            <th scope="col" style="text-align: center;">Metric Count</th>
-          {/if}
-          <th scope="col" style="text-align: center;">Description</th>
-        </tr>
-      </thead>
-      <tbody>
-        {#each pagedItems as item}
+<svelte:window bind:scrollY />
+
+<div bind:this={topElement}>
```
wlach, line 80: This diff is long, but almost entirely due to a change in indentation

Iinh (approved):
Looks like we need to update the snapshot test.

PR #681, opened 2021-06-15
Various adjustments to margins and padding

wlach:
* Give more vertical padding for content (same as horizontal)
* Adjust vertical alignment of breadcrumb app logos slightly
* Make sure there is a little more space between labels

### DIFF
--- a/src/components/Label.svelte
+++ b/src/components/Label.svelte
@@ -61,7 +61,7 @@
 
   .label {
     padding: 0;
-    margin: 0.25em;
+    margin: 0.125em;
     display: inline-block;
     line-height: 20px;
     font-size: 12px;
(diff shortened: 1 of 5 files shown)


wlach:
Before:

image

After:

image

Before:

image

After:

image

src/components/Label.svelte
```diff
@@ -61,6 +61,7 @@
 
   .label {
     padding: 0;
+    margin: 0.25em;
```
Iinh, line 64:
```suggestion
    margin: 0.125em;
```
I find it looks better to keep the tags closer in this case, but no strong opinion about this.

Iinh (approved):

src/components/Label.svelte
wlach: Yep, agreed. I'll change this.

PR #691, opened 2021-06-21
Reset pagination if filter text ever changes

wlach:
Weird things can happen if you were on a subsequent page and changed
the filter text (up to and including being past the "edge" of where
one is allowed to paginate)

### DIFF
--- a/src/components/FilterInput.svelte
+++ b/src/components/FilterInput.svelte
@@ -10,6 +10,10 @@
     type="search"
     id="filter-input"
     bind:value={$pageState.search}
+    on:input={() => {
+      // we want to reset the page number to 1 if the filter text ever changes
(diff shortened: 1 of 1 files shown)


Iinh:
I found something strange: if I try to click on a page number after the filter text changes, the scroll position changes instead of the page number. Do you know why this is happening?

wlach:
> I found something strange: if I try to click on a page number after the filter text changes, the scroll position changes instead of the page number. Do you know why this is happening?

I can't reproduce this in the deploy preview. Can you give an exact set of steps?

Iinh:
Now I can't reproduce it either 🤦🏻‍♀️ I must have done something weird, sorry for the noise!

Iinh (approved):

PR #692, opened 2021-06-21
Update Glean Dictionary for new "tags" annotation format

wlach:
See mozilla/glean-annotations#46

### DIFF
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -12,8 +12,12 @@ import stringcase
 import yaml
 
 OUTPUT_DIRECTORY = os.path.join("public", "data")
-ANNOTATIONS_URL = "https://mozilla.github.io/glean-annotations/api.json"
-NAMESPACES_URL = "https://raw.githubusercontent.com/mozilla/looker-hub/main/namespaces.yaml"
+ANNOTATIONS_URL = os.getenv(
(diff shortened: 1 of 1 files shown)


scripts/build-glean-metadata.py
```diff
@@ -12,8 +12,12 @@
 import yaml
 
 OUTPUT_DIRECTORY = os.path.join("public", "data")
-ANNOTATIONS_URL = "https://mozilla.github.io/glean-annotations/api.json"
-NAMESPACES_URL = "https://raw.githubusercontent.com/mozilla/looker-hub/main/namespaces.yaml"
+ANNOTATIONS_URL = os.getenv(
+    "ANNOTATIONS_URL", "https://mozilla.github.io/glean-annotations/api.json"
```
Iinh, line 16: Why do we need to change these to environment variables?

Iinh (approved):

scripts/build-glean-metadata.py
wlach: It makes testing changes like this a little easier, since you can just do:

```
export ANNOTATIONS_URL=<x>
./scripts/build-glean-metadata.py
```

... without having to modify the source itself. We use a similar pattern in [lookml-generator](https://github.com/mozilla/lookml-generator)

PR #698, opened 2021-06-22
Broader matching when filtering

wlach:
Split search text by whitespace into a set of tokens, and consider the
term matching if *any* token matches one of metric text, tag, or origin.

Also, extract out filtering logic into a function and write a bunch of tests for it.

### DIFF
--- a/tests/state.filter.test.js
+++ b/tests/state.filter.test.js
@@ -18,7 +18,7 @@ const items = [
     tags: ["TopSites"],
     expires: getDateStr(tomorrow),
   },
-  { name: "metric.camel", expires: getDateStr(tomorrow) },
+  { name: "metric.camel", expires: getDateStr(tomorrow), origin: "glean-core" },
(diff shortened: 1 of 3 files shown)


tests/state.filter.test.js
```diff
@@ -0,0 +1,85 @@
...
+    expect(getNames(filterItems(items, "best sites", true))).toEqual([
+      "metric.bestsitez",
+    ]);
+    expect(getNames(filterItems(items, "best    sites", true))).toEqual([
+      "metric.bestsitez",
+    ]);
+  });
+});
+
+describe("filters on tag", () => {
```
Iinh, line 62: should we add a test for filtering on origin too?

Iinh (approved):
Looks good to me!

tests/state.filter.test.js
wlach: I figured that it was the same test case as tag, but on reflection we should probably cover all the bases. Will add before landing

PR #702, opened 2021-06-23
Persist variant state for metric and ping pages (fixes #693)

### LINKED ISSUE #693: Glean dictionary should store page state in URL
Sometimes I'm investigating a metric and would like to share the URL to the glean dictionary page. For example, working with the sync team, they wanted to look at [this metric](https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/creditcards_sync_failure_reason).

The problem is that I wanted to link them to the pre-release "version" of that metric (for nightly), but I can't. I can only link them to the page and tell them to switch to nightly so they can actually see the data in Looker and GLAM.

Iinh:
Fixes #693.

Examples:
- Metric page: https://deploy-preview-702--glean-dictionary-dev.netlify.app/apps/fenix/metrics/glean_error_invalid_label?channel=org.mozilla.fenix&ping=bookmarks-sync
- Ping page: https://deploy-preview-702--glean-dictionary-dev.netlify.app/apps/fenix/pings/events?channel=org.mozilla.firefox_beta&showExpired=1

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/VariantSelector.svelte
+++ b/src/components/VariantSelector.svelte
@@ -1,5 +1,6 @@
 <script>
   import { createEventDispatcher } from "svelte";
+  import { updateURLState, pageState } from "../state/stores";
 
   export let name;
   export let label;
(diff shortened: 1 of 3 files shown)


src/components/VariantSelector.svelte
```diff
@@ -12,7 +13,19 @@
 
   const change = () => {
     selectedVariant = variants.find((v) => v.id === selectedId);
-
+    $pageState = {
+      ...$pageState,
+      // because both app channel and ping use `id` as the identifier
+      // we check if selectedVariant.channel exists to differentiate
+      // between setting channel and ping state
```
wlach, line 20: This is a little confusing. I feel like maybe we should handle this in the parent? Perhaps via some helper functions to update the store? We could probably use the dispatch function to get notified when this changes.

wlach (changes requested):
Thanks for doing this! This works (mostly) correctly but I'd like to try an approach which doesn't put so much logic in the VariantSelector (which we might want to use in other contexts).

wlach (approved):
Ah right, we can just reactively respond to changes in the variables themselves, since we have a binding on them. Nicely done!

PR #746, opened 2021-07-23
Be more specific when a metric is not supported in GLAM

Iinh:
Recently a user asked in an internal slack channel about why the dictionary doesn't link out to GLAM for `event` metrics. GLAM doesn't support visualizing `event`, but as this is a gotcha that not a lot of people know about, I think it'd be more user-friendly to specifically state that a metric is not supported in GLAM, rather than not showing anything.

Before:
CleanShot 2021-07-23 at 14 52 21@2x

After:

CleanShot 2021-07-23 at 14 51 41@2x

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -220,7 +220,7 @@
               {params.metric}
             </AuthenticatedLink>
           {:else}
-            GLAM doesn't support <code>{metric.type}</code> metrics yet.
+            Currently GLAM doesn't support <code>{metric.type}</code> metrics.
           {/if}
         </td>
       </tr>
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -206,22 +206,24 @@
     <table>
       <col />
       <col />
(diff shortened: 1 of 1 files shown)


src/pages/MetricDetail.svelte
```diff
@@ -206,22 +206,24 @@
...
+        <td>
+          {#if selectedAppVariant.etl.glam_url}
             <AuthenticatedLink href={selectedAppVariant.etl.glam_url}>
               {params.metric}
             </AuthenticatedLink>
-          </td>
-        </tr>
-      {/if}
+          {:else}
+            GLAM doesn't support <code>{metric.type}</code> metrics yet.
```
robhudson, line 223: "yet" implies that GLAM will support events soon. But as far as I know that's not true since events can hold many things and we don't know how to aggregate events in a general case. Do we have a plan for events?

Iinh: you're right, I don't think supporting events is on the roadmap. I'll remove it

robhudson (approved):

Issue #224, opened 2020-11-24
Implement skeleton of search bar


wlach:
The search bar is going to be a complicated beast to get right, but we have to get started somewhere. [GLAM](https://github.com/mozilla/GLAM)'s search bar looks pretty nice and seems like a reasonable starting point for our own implementation:

![image](https://user-images.githubusercontent.com/20569/100150123-2d39ae80-2e6d-11eb-879c-5bf326e1ee04.png)

It is defined here:

https://github.com/mozilla/glam/tree/main/src/components/search

For populating the actual content based on user input, I have started a skeleton prototype of a metadata-extraction script using [lunr.js](https://lunrjs.com/) [here](https://github.com/mozilla/glean-dictionary/blob/main/scripts/build-glean-metadata.js) -- it's very incomplete and probably slightly broken, but it illustrates how one would extract search metadata from ping files. 

On the client side, we would automatically load the lunr data on any page load (hopefully we can keep the size reasonable) which would include the search data + *just* enough metadata to populate the search bar. Typing stuff into the search bar would then activate lunr. You can see some very rough client side code which I prototyped for the schema dictionary here: https://github.com/wlach/schema-dictionary/commit/97b71d16b8872171ed1fe73c66e8f9a0d55d0b10

I suspect it might be on me to produce at least the first draft on this, but putting this up here in the interests of working in the open.

Ankita13-code:
Can I work on this? I've never implemented a search bar before but would be happy to learn and implement this under your guidance @wlach!

wlach:
Going to work on this in the new year -- assigning to myself to make this more clear.

wlach:
I worked a little bit on this, and it's becoming clear that this is going to be kinda confusing given the number of applications that most of our metrics are currently duplicated across. It's also not 100% clear to me that this (rather complex) feature is actually required at this juncture. I'm going to put this in the backlog, we can always return to it later.

wlach:
Now that we've consolidated metrics across applications, this feature is a bit more tractable.

Note to self: we could also consider using something like Algolia here: https://www.algolia.com/ -- I think rolling our own is *probably* the right call, but we should see how it goes.

PR #756, opened 2021-07-26
Allow search by description (fixes #641)

### LINKED ISSUE #641: Add ability to search by metric type or description
It's not clear to me how the UX should behave here; e.g. should searching by default be across types and descriptions, or should it be more like a [logs explorer](https://console.cloud.google.com/logs/query;cursorTimestamp=2021-06-02T13:48:19.209941Z?project=moz-fx-data-ingesti-prod-579d) which requires you to select which field you're searching on.

The use case here is to e.g. see which events are available (this makes the Funnel Analysis explore in Looker easier to use). I'd imagine there might be other use cases (e.g. do we already have a counter for some event?).

Iinh:
Use lunr.js to allow search by description. Fixes #641.

### Pull Request checklist

      fixes, if applicable)


src/components/ItemList.svelte
```diff
@@ -1,4 +1,5 @@
 <script>
+  import * as lunr from "lunr";
```
Iinh, line 2: We do this instead of `import lunr from "lunr"` to pass storybook storyshot test.

Iinh: this didn't work, had to revert

src/components/ItemList.svelte
```diff
@@ -1,4 +1,5 @@
 <script>
+  import lunr from "lunr";
```
Iinh, line 2: Hi @wlach I need some help on this. Storyshot doesn't like this import (see [CircleCI failed test](https://app.circleci.com/pipelines/github/mozilla/glean-dictionary/2126/workflows/0d515da1-5b19-417c-861b-45162d688d41/jobs/4073)). 
Changing this to `import * as lunr from "lunr"` doesn't work (though it would appease storybook) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#import_an_entire_modules_contents

I'm not sure how to work around this 😕

wlach:
@Iinh I suspect that the problem is that lunr is rather old and not really designed to be used in a rollup-based project like this one (see: https://github.com/olivernn/lunr.js/issues/401).

Maybe we should try flexsearch instead?

https://github.com/nextapps-de/flexsearch

Iinh:
I switched to using FlexSearch, thanks @wlach for the recommendation.

src/components/ItemList.svelte
```diff
@@ -38,7 +49,10 @@
   // (above)
   $: {
     // filter items by search terms and expiry state
-    filteredItems = filterItems(items, search, showExpired);
+    filteredItems =
+      (filterItems(items, search, showExpired).length &&
+        filterItems(items, search, showExpired)) ||
+      fullTextSearch(search);
```
Iinh, line 55: So the logic here is that if searching by name/type doesn't return anything, we then try searching by description. Should I combine the 2 results instead of doing this way?

src/components/ItemList.svelte
```diff
@@ -28,6 +29,16 @@
   let topElement;
   let scrollY;
 
+  function fullTextSearch(query) {
+    // create an index of item names and descriptions for full-text search
+    const searchIndex = new Index();
+    items.forEach((item) => searchIndex.add(item.name, item.description));
```
wlach, line 35: I think we should generate the index as early as possible. Re-generating it everytime there's a a search is unnecessary and might be slow.

src/components/ItemList.svelte
wlach: I think we should use what's called a "document" index in flexsearch, and search by name/type/description all at once (then filter by `showExpired` at the very end). I played around with this a bit, here's an example of a document index:

https://irydium.dev/repl?id=b9c36dec-0418-11ec-866c-0a4e9a1089db

wlach (changes requested):
This is so exciting! I'm suspecting we might be able to use flexsearch for searching across all the applications (#224), which I'm super stoked about.

Iinh:
Okay so this is the first draft of combing searching by name/type/description. It seems to work fine if we only search by those 3 dimensions. However, there's an issue if we click on any tag/origin label.
- before full-text search, clicking on a tag returns all metrics of that tag, for example: [engine-gecko](https://dictionary.telemetry.mozilla.org/apps/fenix?itemType=metrics&page=1&search=engine-gecko&showExpired=1) (because there's no name match so the filter function returns the origin/tag match instead)
- with full-text search, this doesn't work anymore: [engine-gecko](https://deploy-preview-756--glean-dictionary-dev.netlify.app/apps/fenix?page=1&search=engine-gecko)

So I tried separating text `search` and `origin` to differentiate full-text searching vs getting all metrics of a certain tag/origin. However, this system seems kinda awkward. Do you have any recommendations @wlach?

wlach:
@Iinh Could you try making the tags/origin another part of the document to be indexed? I believe flexsearch can handle properties in a document that are in an array.

Iinh:
Hi @wlach this still needs more cleaning up but could you review what I have so far? I added tags and origin to the indexing document, and highlight the result text to make it more visible which field matches the result. Example: https://deploy-preview-756--glean-dictionary-dev.netlify.app/apps/fenix?page=1&search=event

CleanShot 2021-09-08 at 16 39 22@2x

src/components/ItemList.svelte
```diff
@@ -28,6 +29,47 @@
...
+    });
+  }
+
+  function handleSearch(searchItems, text, expired) {
+    let searchResult = text ? fullTextSearch(text, searchItems) : searchItems;
+    return filterExpiredItems(searchResult, expired);
+  }
+
+  function highlightSearch(text, query) {
+    if (query !== "") {
```
wlach, line 65:
```suggestion
    if (query.length) {
```

src/components/ItemList.svelte
```diff
@@ -28,6 +29,47 @@
...
+      let re = new RegExp(query, "gi"); // search for all instances
+      let newText = text.replace(re, `<mark>${query}</mark>`);
+      return newText;
```
wlach, line 68: You can probably make this one line.

wlach (commented):
This looks great! I don't think there's too much that needs to change.

src/components/ItemList.svelte
```diff
@@ -28,6 +29,47 @@
...
+      results = [...results, ...result.result];
+    });
+
+    return results.map((result) => {
+      return searchItems.find((item) => item.name === result);
+    });
+  }
+
+  function handleSearch(searchItems, text, expired) {
+    let searchResult = text ? fullTextSearch(text, searchItems) : searchItems;
```
wlach, line 60:
```suggestion
    const searchResult = text ? fullTextSearch(text, searchItems) : searchItems;
```

src/components/ItemList.svelte
```diff
@@ -28,6 +29,47 @@
...
+  function fullTextSearch(query, searchItems) {
+    let results = [];
+    // combine results of all fields
+    searchIndex.search(query).forEach((result) => {
+      results = [...results, ...result.result];
+    });
+
+    return results.map((result) => {
+      return searchItems.find((item) => item.name === result);
+    });
```
wlach, line 56: I think we can simplify this a bit.

```suggestion
    // combine results of all fields
    return searchIndex.search(query).map((result) => {
      return searchItems.find((item) => item.name === result.result);
    });
```

wlach (approved):
Looks good, r+wc.

src/components/ItemList.svelte
Iinh: I don't think we can simplify this. An example of what `searchIndex.search(query)` returns: 

```
[
  {
    "field": "id",
    "result": [
      "about_page.libraries_tapped",
      "about_page.licensing_tapped",
      "about_page.privacy_notice_tapped",
      "about_page.rights_tapped",
      "about_page.support_tapped"
    ]
  },
  {
    "field": "description",
    "result": [
      "metrics.start_reason_process_error",
      "about_page.licensing_tapped",
      "about_page.privacy_notice_tapped",
      "about_page.support_tapped",
      "geckoview.page_reload_time",
      "about_page.libraries_tapped",
      "about_page.rights_tapped",
      "perf.startup.home_activity_on_start"
    ]
  }
]
```
Since there are duplicates in `result` array for each field, we need to first combine and deduplicate the results: `results = [...results, ...result.result]`. Then return only the items in `searchItems` that match the item names in `results`. Maybe there's a different way to simplify this that I'm not seeing?

wlach: Oh, right! Sorry, I misread your code. I still think we can use [flat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) here to concatenate the results, then uniquify the result using Set:

https://stackoverflow.com/questions/9229645/remove-duplicate-values-from-js-array

Iinh: thanks! I used set to uniqfy the result in the latest commit. Could you explain why using flat is better than the spread operator in this case?

wlach: I think you forgot to push your commit. 😄 

Wrt your question: this is a little subjective, but in general I think it's slightly better practice to use map() instead of forEach() + an accumulator (the spread operator in this case). It's a little closer to a pure functional style, which (all things being equal) tends to be easier understand and less prone to bugs. I like this stackoverflow answer on the subject: https://stackoverflow.com/a/43883957/295132

It occurs to me now that maybe you could use flatMap here, which might be even more elegant as well as (very slightly) more efficient:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

Iinh: These are great! Thank you @wlach!

PR #780, opened 2021-08-18
Don't use a reactive declarations to handle updating the URL

wlach:
It's unreliable during page navigation and often causes us to persist
unwanted state and weird bugs. The basic problem is that navigation
(and the subsequent updating of page state) can trigger a reactive
statement inside a component we're navigating *away* from.

Instead, let's explicitly make a call to update the page state when we need
to, which will also update the URL. It's a little less elegant
but at least it seems to work.

### DIFF
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -10,7 +10,7 @@
 
   import { filterItems } from "../state/filter";
   import { isExpired } from "../state/metrics";
-  import { pageState, updatePageState } from "../state/stores";
+  import { pageState, updateURLState } from "../state/stores";
 
   let DEFAULT_ITEMS_PER_PAGE = 20;
 
@@ -23,18 +23,22 @@
   let filteredItems = items.filter((item) => !isExpired(item.expires));
   let pagedItems;
   let paginated = true;
+  let search;
+  let showExpired;
   let topElement;
   let scrollY;
+
+  $: {
+    showExpired =
+      $pageState.showExpired === undefined ? true : $pageState.showExpired;
+    search = $pageState.search || "";
+  }
 
   // update pagedItems when either pagination changes or search text changes
   // (above)
   $: {
     // filter items by search terms and expiry state
-    filteredItems = filterItems(
-      items,
-      $pageState.search || "",
-      $pageState.showExpired
-    );
+    filteredItems = filterItems(items, search, showExpired);
 
     // update pagination
     const currentPage = $pageState.page || 1;
@@ -47,10 +51,10 @@
   }
 
   const updateSearch = (origin, type = undefined) => {
-    updatePageState(
+    updateURLState(
       type
-        ? { ...$pageState, search: origin, page: 1, itemType: type }
-        : { ...$pageState, search: origin, page: 1 },
+        ? { search: origin, page: 1, itemType: type }
+        : { search: origin, page: 1 },
       true
     );
     // reset scroll position if we've scrolled down
@@ -71,7 +75,15 @@
     {#if itemType === "metrics"}
       <span class="expire-checkbox">
         <label>
-          <input type="checkbox" bind:checked={$pageState.showExpired} />
+          <input
+            type="checkbox"
+            bind:checked={showExpired}
+            on:change={() => {
+              // the binding changes *after* this callback is called, so use
+              // the inverse value
+              updateURLState({ showExpired: !showExpired });
+            }}
+          />
           Show expired metrics
         </label>
         <label>
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -10,7 +10,7 @@
 
   import { filterItems } from "../state/filter";
   import { isExpired } from "../state/metrics";
-  import { pageState, updateURLState } from "../state/stores";
+  import { pageState, updatePageState } from "../state/stores";
 
   let DEFAULT_ITEMS_PER_PAGE = 20;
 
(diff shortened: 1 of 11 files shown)


wlach:
Example test case:

* Go to https://dictionary.telemetry.mozilla.org
* Click on Firefox for Android
* Search for Architecture, select it
* Select addresses-sync ping from "send in pings"

This hangs for me on Firefox, with this error message:

image

src/state/stores.js
```diff
@@ -1,21 +1,27 @@
...
-  const simplifiedState = mapValues(
-    pickBy(get(pageState), (v) => (typeof v !== "string" && v) || v.length > 0),
+// updates page state and synchronizes with the url
+export const updatePageState = (newState, push = false) => {
+  const mergedState = mapValues(
+    pickBy(
+      { ...get(pageState), ...newState },
+      (v) =>
+        (typeof v !== "string" && v) ||
+        (v && ((v.length > 0 && v !== "page") || v !== "1"))
```
Iinh, line 15: Maybe you've already explained this to me before (sorry!), but why do we need these conditions here, especially this part `(v && ((v.length > 0 && v !== "page") || v !== "1"))`?

src/components/ItemList.svelte
```diff
@@ -47,11 +47,13 @@
...
-    $pageState = type
-      ? { ...$pageState, search: origin, page: 1, itemType: type }
-      : { ...$pageState, search: origin, page: 1 };
+    updatePageState(
+      type
+        ? { ...$pageState, search: origin, page: 1, itemType: type }
+        : { ...$pageState, search: origin, page: 1 },
+      true
+    );
     // when the user clicks on an origin (library name), we want to persist a new state
```
Iinh, line 56: I think we can remove this comment now.

Iinh (approved):

src/state/stores.js
wlach: The page thing was a silly experiment of mine to reduce the length of the URLs (there's no need to include it the URL if page=1), but we should do that in a more principled way (and not as part of this PR). I'll take it out!

wlach:
I noticed some bugs in my earlier implementation while fixing the review comments. In particular, the back button didn't always work as expected. These changes are probably big enough that they merit another round of review. Sorry for the noise! Most of the changes are in https://github.com/mozilla/glean-dictionary/pull/780/commits/4a9334fd2378397755145ccc1f6c266a66ec1262 so you can just focus on that.

src/pages/AppList.svelte
```diff
@@ -12,11 +12,6 @@
 
   const URL = "data/apps.json";
 
-  $pageState = {
-    search: "",
-    ...$pageState,
-  };
-
```
Iinh, line 19: Looking at this again I don't think this should be removed, at least the resetting search part. I think it makes sense to keep resetting search state in this component: expected behavior is empty search whenever we click on the homepage right?

src/state/stores.js
```diff
@@ -1,21 +1,36 @@
...
+// before updating the store
+const setPageState = (newState) => {
+  pageState.set(mapValues(
+    pickBy(
+      { ...get(pageState), ...newState },
+      (v) => (typeof v !== "string" && v) || v.length > 0
+    ),
     (v) => (typeof v === "boolean" ? +v : v)
   );
+};
```
Iinh, line 18: Looks like we're missing a `)` here.

src/state/stores.js
```diff
@@ -1,21 +1,36 @@
...
+
+// replaces the existing page state
+export const replacePageState = (newState) => {
+  setPageState(newState);
+};
```
Iinh, line 23: I'm probably missing something but why do we need to create a new function here instead of just using `setPageState()` for `App.svelte`?

Iinh (changes requested):

src/pages/AppList.svelte
wlach: Not necessarily. If there is a search in the URL (e.g. `?search=android`) we want to bring that into the component. That's the purpose of the `replaceState` function I just added: make sure that *only* the URL parameters find their way into the component.

src/state/stores.js
wlach: You're right, I'll change this up.

src/state/stores.js
wlach: Also we shouldn't be doing the state simplification there! 🤦

wlach:
Found some more issues, working on them.

src/pages/AppList.svelte
Iinh: You're right. I was thinking if a user searches for something in the search bar in AppDetail and then clicks on home page, that search state will carry over to AppList which I'm not sure looks right. How should we handle that?

wlach: In the last two revisions, we synchronize pageState with the URL when navigating here:

https://github.com/mozilla/glean-dictionary/pull/780/files#diff-d68daa8b73cec8e4419759802bbd91e9f87f8ee02f25cdb1d12582ef6de848dbR63

This seems to take care of the problem.

wlach:
Ok, I think this is really working now. The model here is more complex than I'd like-- I think we could probably simplify this a lot by moving to SvelteKit (I'd probably also change the stores to context). I don't think page.js is really designed to be used in reactive application like this, so everything is kind of overly complex.

Iinh:
This looks good, the only issue left is `showExpired` state doesn't persist correctly. For example: https://deploy-preview-780--glean-dictionary-dev.netlify.app/apps/fenix?page=1&showExpired=0 

That link has showExpired=0, but it isn't reflected in the checkbox. But looks like the production site doesn't refresh showExpired state correctly either so the issue might not be related to this PR.

wlach:
> This looks good, the only issue left is `showExpired` state doesn't persist correctly. For example: https://deploy-preview-780--glean-dictionary-dev.netlify.app/apps/fenix?page=1&showExpired=0
> 
> That link has showExpired=0, but it isn't reflected in the checkbox. But looks like the production site doesn't refresh showExpired state correctly either so the issue might not be related to this PR.

Oh good catch! That's easily fixable but let's do that in a followup.

Iinh (approved):

PR #788, opened 2021-08-24
Restore boolean (and numeric) properties stored in URL state

wlach:
We were processing them as strings, which meant they weren't being
incorporated as page state correctly. Follow up for:
https://github.com/mozilla/glean-dictionary/pull/780

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -60,7 +60,7 @@
 
   function setComponent(c) {
     return function setComponentInner(ctx) {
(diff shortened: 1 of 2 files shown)


Iinh (approved):

Iinh:
We probably can also change the below to just `currentPage = $pageState.page || 1;` with this change.
https://github.com/mozilla/glean-dictionary/blob/681ae2b2d3b6f602cefbbbec1206353666d9961c/src/components/Pagination.svelte#L56

PR #789, opened 2021-08-24
Improve documentation and help in table view page (fixes #787)

### LINKED ISSUE #787: Provide sample queries on table page
On a BQ table page, we list "live data" and "stable view", each with a table/view qualified at the namespace level. Users may be lacking context for what these terms mean, so we should provide an info field that links to the "table layout and naming" section of DTMO (https://docs.telemetry.mozilla.org/cookbooks/bigquery/querying.html#table-layout-and-naming) and provide some immediate message about how the live tables are generally not considered user-facing and should usually only be used for debugging purposes tracking new telemetry.

Additionally, we could provide sample queries to provide more complete guidance about how to start investing these views and encouraging efficient use of partitioning, clustering, and limiting. For stable views, the example could be:

```
SELECT
  *
FROM
  mozdata.org_mozilla_firefox.events  # We'd inject the view name here
WHERE
  DATE(submission_timestamp) = (CURRENT_DATE() - 1)
LIMIT
  10
```

That codifies the pattern of looking at just the most recent day, limiting to just a few records, and fully qualifying the name with the `mozdata` project (so that this example should work regardless of which project it's launched from).

A live example could be:

```
SELECT
  *
FROM
  `moz-fx-data-shared-prod.org_mozilla_firefox_live.events_v1`
WHERE
  submission_timestamp > (CURRENT_TIMESTAMP() - INTERVAL 1 HOUR)
LIMIT
  10
```

Here, we pick just the most recent hour of data (since live tables are clustered on submission_timestamp) and again have a limit. Also note that we fully qualify with `moz-fx-data-shared-prod` which is necessary since we do not automatically publish views for live data in the `mozdata` project.

It's not entirely clear to me how it would be best to present example queries. It may be ideal to have these as some sort of expandable item alongside the "live data" and "stable view" listings.

wlach:


src/pages/TableDetail.svelte
```diff
@@ -19,22 +20,49 @@
...
+      <td>
+        <a href={table.bq_definition}>
+          {table.bq_definition.split("/").slice(-1)}
+        </a>
+      </td>
     </tr>
   </table>
+  <p>
+    To access BigQuery data directly, you need to write queries using SQL. For
+    more information on how to do this, see <a
```
wlach, line 59: I struggled to find a friendlier way of writing this, but came up blank today.

src/pages/TableDetail.svelte
```diff
@@ -19,22 +20,49 @@
...
+          link={"https://docs.telemetry.mozilla.org/cookbooks/bigquery/querying.html#table-layout-and-naming"}
+        />
+      </td>
       <td><code>{table.live_table}</code></td>
     </tr>
     <tr>
-      <td>Stable View</td>
-      <td><code>{table.stable_table}</code></td>
+      <td
+        >Schema
```
wlach, line 44: This seems less interesting to most people than the live/stable table information, so I moved it to the bottom.

src/pages/TableDetail.svelte
```diff
@@ -19,22 +20,49 @@
     <col />
     <col />
     <tr>
-      <td>BigQuery Definition</td>
-      <td>
-        <a href={table.bq_definition}>
-          {table.bq_definition.split("/").slice(-1)}
-        </a>
+      <td
+        >Stable
```
wlach, line 24: I took out "Data" here because it seemed redundant.

jklukas (approved):
r+wc

Thanks for quickly adding these! It'll be fun to test them out.

src/pages/TableDetail.svelte
```diff
@@ -19,22 +20,56 @@
...
     <col />
     <tr>
-      <td>BigQuery Definition</td>
       <td>
-        <a href={table.bq_definition}>
-          {table.bq_definition.split("/").slice(-1)}
-        </a>
+        Stable
+        <HelpHoverable
+          content={"BigQuery table containing stable ping data, rolled up once per day with additional deduplication applied as well as partitioning by sample id. Use this table for most analysis."}
```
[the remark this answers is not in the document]
wlach: Touche! Not sure how I managed to swap "partitioned" for "clustered" in my head.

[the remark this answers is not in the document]
jklukas: > Not sure how I managed to swap "partitioned" for "clustered" in my head.

Partioning/sorting/clustering mean a wide variety of related things in different data technologies, so confusion is to be expected.

PR #803, opened 2021-09-01
Show retention period in days

wlach:
This is also likely useful from a data stewardship and legal
perspective (and maybe even for engineering and product).

### DIFF
--- a/src/data/schemas.js
+++ b/src/data/schemas.js
@@ -27,6 +27,17 @@ export const APPLICATION_DEFINITION_SCHEMA = [
       "A list of email addresses to notify for important events with the application or when people with context or ownership for the application need to be contacted.",
     linkFormatter: getEmailLink,
   },
+  {
+    title: "Rentention Period",
+    id: "retention_days",
+    displayNull: true,
+    type: "value",
+    helpText:
+      "The number of days Mozilla retains raw decoded ping data received for this application. If not specified, retention will be unlimited.",
+    valueFormatter: (value) => {
+      return value !== null ? `${value} days` : "Unlimited";
+    },
+  },
 ];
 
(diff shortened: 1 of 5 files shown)


src/data/schemas.js
```diff
@@ -27,6 +27,17 @@ export const APPLICATION_DEFINITION_SCHEMA = [
       "A list of email addresses to notify for important events with the application or when people with context or ownership for the application need to be contacted.",
     linkFormatter: getEmailLink,
   },
+  {
+    title: "Rentention Period",
+    id: "retention_days",
+    displayNull: true,
+    type: "value",
+    helpText:
+      "The number of days Mozilla retains raw decoded ping data received for this application. If not specified, retention will be unlimited.",
```
wlach, line 36: This might need rephrasing, comments welcome.

chutten: "raw" is too fuzzy here since there is indeed processing (validating, copy-dedupe, etc) and "decoded" is fuzzy as well. I think there might be a wording around "full" or "unaggregated" or "individual" or "high-resolution"... but yeah, tricky.

Maybe...
```suggestion
      "The number of days Mozilla retains raw ping data received for this application. If not specified, retention will be unlimited. Processed data may be retained longer than specified.",
```

I'm not sure. Lemme noodle.

chutten: Yeah, no, in the morning light how you have it written is certainly clear enough for both our internal and external audiences. I do wish we could link to some default retention policy but I understand that's still a subject of discussion : |

chutten (approved):

wlach:
Sounds good @chutten. I'll most likely be looking into this area (data governance) more in the coming weeks / months, there will be opportunities to improve wording as we go.

PR #815, opened 2021-09-07
Make it more clear if metric is removed (fixes #705)

### LINKED ISSUE #705: Make it clear when a metric is no longer present in source code
We warn when a metric is expired, but we don't currently make it clear when a metric is no longer in the source code. Once https://github.com/mozilla/probe-scraper/issues/320 is implemented, we should do this.

Example of where this would be useful: https://github.com/mozilla-mobile/fenix/pull/20156#issuecomment-868492244

wlach:
"Expired" and "removed" are often the same, but not always (e.g.
https://glean-dictionary-dev.netlify.app/apps/fenix/metrics/credit_cards_autofill_card
is not due to expire for quite some time, but has been removed from the
source code)

### DIFF
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -9,7 +9,7 @@
   import Markdown from "./Markdown.svelte";
   import Label from "./Label.svelte";
 
-  import { filterExpiredItems } from "../state/filter";
+  import { filterUncollectedItems } from "../state/filter";
   import { isExpired } from "../state/metrics";
   import { pageState, updateURLState } from "../state/stores";
 
@@ -25,7 +25,7 @@
   let pagedItems;
   let paginated = true;
   let search;
-  let showExpired;
+  let showUncollected;
   let topElement;
   let scrollY;
 
@@ -55,7 +55,7 @@
 
   function handleSearch(searchItems, text, expired) {
     const searchResult = text ? fullTextSearch(text, searchItems) : searchItems;
-    return filterExpiredItems(searchResult, expired);
+    return filterUncollectedItems(searchResult, expired);
   }
 
   function highlightSearch(text, query) {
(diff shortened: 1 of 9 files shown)


src/pages/MetricDetail.svelte
```diff
@@ -78,10 +78,15 @@
 </script>
 
 {#await metricDataPromise then metric}
-  {#if isExpired(metric.expires)}
+  {#if !metric.in_source}
     <AppAlert
       status="warning"
-      message="This metric has expired: it may not be present in the source code, new data will not be ingested into BigQuery, and it will not appear in dashboards."
+      message={"This metric is no longer defined in the source code: new data will not be collected by the application."}
```
wlach, line 84: This felt like a good opportunity to simplify the messaging here.

src/components/ItemList.svelte
```diff
@@ -149,7 +151,9 @@
                       />
                     {/each}
                   {/if}
-                  {#if isExpired(item.expires)}
+                  {#if item.in_source === false}
```
Iinh, line 154:
```suggestion
                  {#if !item.in_source}
```

src/state/filter.js
```diff
@@ -18,7 +18,8 @@ export function filterItems(items, search, showExpired) {
 
   return items.filter(
     (item) =>
-      (showExpired || !isExpired(item.expires)) &&
+      (showUncollected ||
+        (!isExpired(item.expires) && item.in_source !== false)) &&
```
Iinh, line 22:
```suggestion
        (!isExpired(item.expires) && item.in_source)) &&
```

Iinh (approved):

src/components/ItemList.svelte
Iinh: i'm wondering if it's possible for `item.in_source` to ever be undefined or null?

wlach: It shouldn't be, no. This code is a holdover from an earlier approach that I abandoned.

Issue #697, opened 2021-06-22
Add some kind of end-to-end testing


wlach:
#695 happened because we didn't manually test a case (has an annotation but no tags).

Ideally we would have some end-to-end tests that verified this type of case, I think puppeteer is the most popular standard for doing this (and has Firefox support):

https://github.com/puppeteer/puppeteer
https://jestjs.io/docs/puppeteer

wlach:
Playwright looks like another interesting option: https://playwright.dev/

Might be inclined to try that

wlach:
FWIW I asked about this on webcompat a while back and got this explanation from jgraham:

> wlach: So that's a slightly exciting question. Playwright ships a fork of Firefox with a custom automation protocol. Cypress uses WebExtensions to implement its own protocol. Basically only WebDriver things like Selenium are really natively supported in Firefox, but we're working on a standardised protocol to enable Playwright/Puppeteer/Cypress like functionality in Firefox. But that's a way off being useful. In the meantime we have basic support for enough of the Chrome DevTools Protocol to support basic Puppeteer use cases.
> So that was arguably a totally unhelpful answer since the question was presumably more about which has the best API &c. I think people who use Playwright and Puppeteer tend to be very happy with them; if you don't need advanced features like request interception WebDriver-based things like Selenium might be easier.

wlach:
#863 fixed this

PR #828, opened 2021-09-13
Add tooltips for library and feature metadata

wlach:
* Add tooltips for library and feature metadata and use them
  consistently throughout he application
* Add a "page header" component to be used on all pages with tags
  beneath them
* Remove "warning" banners for libraries on metric and ping pages in
  favour of using tags (with these descriptions)
* Fix some bugs in our usage of tippy.js which made descriptions
  not work with the itemlist


Iinh:
I'm testing with the deploy preview and at first glance it seems that AppDetail page for apps other than fenix is broken. Examples: 
- https://deploy-preview-828--glean-dictionary-dev.netlify.app/apps/bergamot
- https://deploy-preview-828--glean-dictionary-dev.netlify.app/apps/burnham

With the below error:
CleanShot 2021-09-14 at 15 58 40@2x

wlach:
Argh! Thanks @Iinh. More ammunition to spend some time on #697

scripts/build-glean-metadata.py
```diff
@@ -227,6 +241,9 @@ def get_app_variant_description(app):
     if all([app_id.get("prototype") for app_id in app_group["app_ids"]]):
         app_data["prototype"] = True
 
+    app_annotation = _get_annotation(annotations_index, app_name, "app")
+    app_tags = app_annotation.get("tags")
```
Iinh, line 245: Do we need to provide a fall back value for `app_annotation.get("tags")` or does `tags` always exist?

src/components/tippy.js
```diff
@@ -1,5 +1,14 @@
...
+  let instance = props && tippy(node, props);
+  return {
+    update(newProps) {
+      if (instance) instance.destroy();
+      if (newProps) instance = tippy(node, newProps);
+    },
+    destroy() {
+      if (instance) instance.destroy();
+    },
+  };
```
Iinh, line 13: I wasn't able to understand why this change was necessary. According to the PR description, is it to "fix some bugs in our usage of tippy.js which made descriptions not work with the itemlist"?

src/pages/AppDetail.svelte
```diff
@@ -72,7 +74,9 @@
   >
     <Tab key="metrics">Metrics</Tab>
     <Tab key="pings">Pings</Tab>
-    {#if app.tags && app.tags.length}
+
+    {#if false && app.tags && app.tags.length}
+      <!-- commenting this out for now, see: https://github.com/mozilla/glean-dictionary/issues/832 -->
```
Iinh, line 79: 👍

Iinh (approved):
This looks great to me! I only have some clarifying questions below.

scripts/build-glean-metadata.py
wlach: If the app has no tags then there shouldn't be any annotations with them: it may be best to fallback to `{}` in this case just for clarity though. I'll change this.

src/components/tippy.js
wlach: In the metric list group, labels may be updated or destroyed as the user paginates through the content. This could result in stale content appearing for a label.

Iinh: ah I see what you mean. Thanks for the explanation!

PR #829, opened 2021-09-13
Fix incorrectly applied "removed" status for pings

wlach:
Instead, we'll just say that they're always in source until we can
determine otherwise (see https://github.com/mozilla/probe-scraper/issues/341)


wlach:
This bug is very visible if you look at the pings page for Fenix:

https://dictionary.telemetry.mozilla.org/apps/fenix?itemType=pings&page=1

image

Iinh (approved):

wlach:
Yuck, this actually only half-fixes the problem since app_ids suffer from the same thing. I think the right solution here is to only consider an item removed if the in_source annotation is *actually* false. Not all items have the "in source" annotation. Will update PR.

src/state/items.js
```diff
@@ -0,0 +1,12 @@
+export const isRemoved = (item) => {
+  // only consider an item expired if it has an in_source property
+  // and it is false
+  return item.in_source === false;
+};
+
+export const isExpired = (item) => {
```
wlach, line 7: It felt like it would be best if these two methods took the same argument.

Iinh (approved):
Thanks for your patience, this looks good!

PR #833, opened 2021-09-14
Update package lock format to v2

wlach:
Also specify that npm v7 (earliest version that uses this new format) and nodejs 12.0+ (what we are testing with in ci) are required.

This will hopefully avoid footguns if people are checking out and developing the Glean Dictionary for the first time.


wlach:
Oops! Forgot to ask for review

Iinh (approved):

Iinh:
@wlach why do we call this version of package-lock "v2"? Is it merely because it's a different format than before?

wlach:
> @wlach why do we call this version of package-lock "v2"? Is it merely because it's a different format than before?

Yup, pretty much. This is an npm change. More detail here, if you care to look:

https://nitayneeman.com/posts/catching-up-with-package-lockfile-changes-in-npm-v7/

PR #834, opened 2021-09-15
Netlify-based search service for both legacy and current glean metrics

### LINKED ISSUE #926: Scrape legacy Firefox metadata (for searching + API queries only)
As a follow-on to #834, we might consider scraping legacy Firefox desktop probe metadata just to create a search index + a set of json files corresponding to the data for this platform. This would let us deprecate the probe search service in GLAM, replacing it entirely with the Glean Dictionary.

wlach:
Fixes #926 and a precondition for mozilla/glam#1678


scripts/build-search-index.js
```diff
@@ -0,0 +1,36 @@
...
+    const metricSummary = {
+      id: metric.name,
+      type: metric.type,
+      tags: (metric.tags || []).map((tag) => tag.name),
+      app,
+      description: metric.description,
+    };
+    metricIndex[globalId] = metricSummary;
+  });
+});
```
wlach, line 27: This step could probably be handled entirely within the existing `build-glean-metadata.py` script, but wrote things up this way for ease of prototyping.

README.md
```diff
@@ -50,6 +50,38 @@ Android) only, try:
 ./scripts/build-glean-metadata fenix
 ```
 
+For the search service _only_ (see below), the Glean Dictionary also indexes
+legacy Firefox telemetry metadata. To build this index (not necessary for most
+testing), you need to run a seperate script:
+
+```bash
+./scripts/build-legacy-metadata
+```
```
wlach, line 59: I would prefer not to have two scripts here, but it seems like the cleanest approach for now until we can further refactor the code in `__main__.py`  to be a bit modular. Will address in a followup.

etl/firefox_legacy.py
```diff
@@ -0,0 +1,74 @@
...
+from .search import create_metrics_search_js
+
+OUTPUT_DIRECTORY = os.path.join("public", "data")
+FUNCTIONS_DIRECTORY = ".netlify"
+PROBES_URL = "https://probeinfo.telemetry.mozilla.org/firefox/all/main/all_probes"
+
+# a bit of a hack, we would ideally make this public data. this data is already implicitly
+# public via the probe search service, so putting this here. probably best to make this
+# a public data set produced by bigquery-etl in the long run though
+PROBE_RECORDED_IN_PROCESSES_URL = "https://sql.telemetry.mozilla.org/api/queries/83152/results.json?api_key=Z1WhC2sRmLXmSwZyUwhIOFQQQAUK3JvGNLZgtZGW"  # noqa
```
wlach, line 15: This should be replaced with a link to the public data json once https://github.com/mozilla/bigquery-etl/pull/2555 lands

README.md
```diff
@@ -50,6 +50,38 @@ Android) only, try:
...
+
+## Search Service
+
+The Glean Dictionary also includes a search service which enables searching
+through active metrics. Under the hood, this service is implemented with
+[netlify functions]. For example:
+
+https://dictionary.telemetry.mozilla.org/.netlify/functions/metrics_search_burnham?search=techno
```
Iinh, line 67: something we might consider in a follow up is to simplify the api path [with redirects](https://www.jamesqquick.com/blog/how-to-redirect-netlify-functions-to-a-simpler-path).

etl/firefox_legacy.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def get_legacy_firefox_metric_summary():
+    # pull down the recorded in process information, which we use as the
```
Iinh, line 19: out of curiosity, is there a reason for when you use docstrings for function description ([example](https://github.com/mozilla/glean-dictionary/blob/ce84fc6b33bc58127792db0b74d46d9217bb03c6/etl/__main__.py?_pjax=%23js-repo-pjax-container%3Afirst-of-type%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%3Afirst-of-type%2C%20%5Bdata-pjax-container%5D%3Afirst-of-type#L76)) and when you don't? I'm trying to learn what's the good practice for Python function documentation.

.netlify/search.js
```diff
@@ -0,0 +1,71 @@
...
+          score += item.active || item.active === undefined ? 1 : 0;
+          return score;
+        }
+      : (item) => {
+          let score = 0;
+          // sort expired metrics below non-expired ones
+          score += !isExpired(item) ? 1 : 0;
+          // if in glam mode, we want to move unsupported metrics right to the bottom
+          score +=
+            glamMode && SUPPORTED_GLAM_METRIC_TYPES.has(item.type) ? 10 : 0;
```
Iinh, line 54: why is it 10 here?

Iinh (approved):
This looks great!! Using this new API with GLAM I noticed the search now returns much better and faster results 🎉

.netlify/search.js
wlach: Because we want to prioritize expired metrics above ones that glam doesn't support at all. Or at least that's what I infer! I'll add a comment.

README.md
wlach: Definitely. 👍 Would you be interested in taking that on?

etl/firefox_legacy.py
wlach: Unfortunately I don't have a great answer here. Ideally we would have docstrings for every function.

https://github.com/mozilla/lookml-generator is a much better example to work from than the Glean Dictionary (I believe @relud even hooked up CI to make sure that every public function has a docstring).

I'll update at least this one to have a docstring.

PR #837, opened 2021-09-16
Better ordering of applications on main page (fixes #456)

### LINKED ISSUE #456: Better ordering of applications on main (application) page
Currently we display the applications in sort-order, which is not ideal. Ideally we would have some heuristic based on an application's relative popularity (e.g. number of ping submissions in the last week), though a manually curated ordering might do in a pinch.

Iinh:
We display a manually-curated order of applications for now:

1. Firefox Desktop
2. Firefox Android
3. Firefox iOS

CleanShot 2021-09-16 at 13 49 31@2x
### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/pages/AppList.svelte
+++ b/src/pages/AppList.svelte
@@ -25,9 +25,14 @@
     filteredApps = apps;
   });
 
+  const appOrder = ["firefox_ios", "fenix", "firefox_desktop"];
+
   $: {
     // update page state when user filters something
     if (apps) {
+      apps.sort(
+        (a, b) => appOrder.indexOf(b.app_name) - appOrder.indexOf(a.app_name)
+      );
       filteredApps = apps.filter((appItem) =>
         appItem.canonical_app_name
           .toLowerCase()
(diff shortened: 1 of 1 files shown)


src/pages/AppList.svelte
```diff
@@ -25,9 +25,14 @@
     filteredApps = apps;
   });
 
+  const appOrder = ["firefox_ios", "fenix", "firefox_desktop"];
```
wlach, line 28: I would uppercase this constant (and put it inline in the onMount method, see below)

```suggestion
  const APP_SORT_ORDER = ["firefox_ios", "fenix", "firefox_desktop"];
```

src/pages/AppList.svelte
```diff
@@ -25,9 +25,14 @@
...
+
   $: {
     // update page state when user filters something
     if (apps) {
+      apps.sort(
```
wlach, line 33: I would move this to the `onMount` call, just below the other sort.

wlach (changes requested):
Looks good, just some minor quibbles about where the sorting is happening (I would do it at definition time).

wlach (approved):
👍

PR #838, opened 2021-09-17
Distinguish libraries from other tags / annotations

wlach:
Use the "package" emoji, which seems to get the point across pretty
well.

### DIFF
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -22,6 +22,7 @@
     getLibraryDescription,
   } from "../data/help";
   import { stripLinks } from "../formatters/markdown";
+  import { getLibraryName } from "../formatters/library";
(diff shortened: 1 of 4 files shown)


wlach:
Example:

image

src/pages/MetricDetail.svelte
```diff
@@ -92,7 +93,7 @@
         {#if metric.origin !== params.app}
           <a href={getMetricSearchURL(params.app, metric.origin)}
             ><Label
-              text={metric.origin}
+              text={getLibraryName(metric)}
               description={getLibraryDescription("metric", metric.origin)}
```
wlach, line 97: The amount of repetition in this PR indicates we might want to create a higher-level abstraction for displaying tags, but let's leave that for later.

Iinh (approved):
👍

PR #849, opened 2021-09-20
Make it clear if a metric/ping was recently added (fixes #836)

### LINKED ISSUE #836: Add some metadata to tell if a metric was recently added
During the fenix event explore demo yesterday, the question came up on whether it was possible to see when a metric was first added. It currently *isn't* but we the probe info service does provide this information if you look at the revision information.

We should add a metadata column to the table saying when it was first seen and maybe also provide some warning text under "Access" saying that it was recently added.

If an app has more than one channel:

"This metric was recently added. It may take some time before it is available on all channels."

If an app has only one channel (e.g. mozilla vpn):

"This metric was recently added. Data may not be available until a new version is released with this metric and clients are updated to it."

(or something like the above, we can wordsmith in the PR)

Iinh:
fixes #836.

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/data/schemas.js
+++ b/src/data/schemas.js
@@ -104,6 +104,12 @@ export const METRIC_METADATA_SCHEMA = [
     helpLink: OPTIONAL_METRIC_PARAMS_DOCS,
   },
   {
+    title: "First Seen",
+    id: "date_first_seen",
+    type: "value",
+    helpText: "The date when this metric was first seen.",
+  },
+  {
     title: "Time Unit",
     id: "time_unit",
     type: "value",
@@ -208,6 +214,12 @@ export const PING_SCHEMA = [
     type: "link",
     helpText: "Where the source definition of the ping may be found.",
     valueFormatter: getSourceUrlTitle,
+  },
+  {
+    title: "First Seen",
+    id: "date_first_seen",
+    type: "value",
+    helpText: "The date when this ping was first seen.",
   },
   {
     title: "Includes Client Identifier",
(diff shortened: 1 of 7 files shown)


src/state/items.js
```diff
@@ -10,3 +10,9 @@ export const isExpired = (item) => {
   }
   return new Date() > new Date(item.expires);
 };
+
+export const isRecent = (item) => {
+  const daysDifference =
+    (new Date() - new Date(item.date_first_seen)) / (1000 * 3600 * 24);
+  return daysDifference < 30;
```
Iinh, line 17: can we assume that "recent" means within the last 30 days, or should I change this?

src/pages/PingDetail.svelte
```diff
@@ -65,7 +70,11 @@
     title={"Access"}
     helpText={"Ways to access this metric in Mozilla's data warehouse."}
   />
-
+  {#if isRecent(ping)}
+    <AppAlert
+      status="warning"
```
wlach, line 75: I think it would be nice to have an "info" app alert with this type of information, since the warning one is a little alarming. Definitely something to do in a followup though.

src/data/schemas.js
```diff
@@ -103,6 +103,12 @@ export const METRIC_METADATA_SCHEMA = [
       "Defines the lifetime of the metric. Different lifetimes affect when the metrics value is reset.",
     helpLink: OPTIONAL_METRIC_PARAMS_DOCS,
   },
+  {
+    title: "First Seen",
+    id: "date_first_seen",
+    type: "value",
+    helpText: "The date when this metric was first seen.",
```
wlach, line 110: Part of me feels like we should format the date a bit more, but I don't want to overcomplicate this PR. Let's put this in front of people and see what their feedback is.

wlach (approved):
🎉

PR #861, opened 2021-09-27
Fix Firefox for Android Logo(Fixes #850)

### LINKED ISSUE #850: Firefox for Android shouldn't use Fenix logo
@magorlick mentioned this. It's confusing that we're using the Fenix (nightly) logo for Firefox for Android. We should just use the same logo we use for Firefox for Desktop and Firefox for iOS.

image

The logic we need to change is around here: https://github.com/mozilla/glean-dictionary/blob/67e48bc4a0aedd3fdea32f22c43ff54080edfe9a/src/pages/AppList.svelte#L59

chichi012:
Fixes #850

### DIFF
--- a/src/pages/AppList.svelte
+++ b/src/pages/AppList.svelte
@@ -18,6 +18,7 @@
 
   onMount(async () => {
     apps = await fetchJSON(URL);
+    console.log("App name", apps)
     apps.sort((a, b) =>
       a.canonical_app_name.toLowerCase() > b.canonical_app_name.toLowerCase()
         ? 1
@@ -57,7 +58,7 @@
 
   function getAppLogo(app) {
     if (app.match(/fenix/) || app.match(/nightly/)) {
-      return appLogos.nightly;
+      return appLogos.browser;
     }
     if (app.match(/beta/)) {
       return appLogos.beta;
--- a/src/pages/AppList.svelte
+++ b/src/pages/AppList.svelte
@@ -18,7 +18,7 @@
 
   onMount(async () => {
     apps = await fetchJSON(URL);
-    console.log("App name", apps)
+    console.log("App name", apps);
     apps.sort((a, b) =>
       a.canonical_app_name.toLowerCase() > b.canonical_app_name.toLowerCase()
         ? 1
--- a/src/pages/AppList.svelte
+++ b/src/pages/AppList.svelte
@@ -18,7 +18,6 @@
 
   onMount(async () => {
     apps = await fetchJSON(URL);
-    console.log("App name", apps);
     apps.sort((a, b) =>
       a.canonical_app_name.toLowerCase() > b.canonical_app_name.toLowerCase()
         ? 1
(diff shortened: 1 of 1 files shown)


chichi012:
@wlach A couple of observations

1. From line 13 in the `AppList.svelte` file ,`const URL = "data/apps.json"` I saw we're pulling the apps from that URL. In my local `data/apps.json` file, I noticed that `app_name` for Firefox for Android is `fenix`. We should consider(if possible) renaming `app_name`  to `firefox_android` to be consistent with `firefox_ios`, `firefox_desktop`, etc.

2. Line 59 in the `AppList.svelte`, I made any app that matches with `nightly` or `fenix` to use the `browser` logo. Currently it looks like only Firefox for Android satisfies that but another app can easily match `nightly` or `fenix` and this can mask bugs. Hence my suggestion above.

src/pages/AppList.svelte
```diff
@@ -57,7 +58,7 @@
 
   function getAppLogo(app) {
     if (app.match(/fenix/) || app.match(/nightly/)) {
-      return appLogos.nightly;
+      return appLogos.browser;
```
Iinh, line 61: We used to index Firefox Nightly but not anymore, so I think we can safely remove line 59. Since currently the only product that has "fenix" in its name is Firefox for Android, I think it might be better to add this condition to line 81, something like:

```
if (app.match(/firefox/) || app.match(/fenix/)) {
      return appLogos.browser;
}
```

src/pages/AppList.svelte
```diff
@@ -18,6 +18,7 @@
 
   onMount(async () => {
     apps = await fetchJSON(URL);
+    console.log("App name", apps);
```
Iinh, line 21: I think you forgot to take this out?

Iinh (changes requested):
Hi @chichi012, thank you for taking this up! I think I should review this PR since I'm the one who worked on adding app logos.

> From line 13 in the AppList.svelte file ,const URL = "data/apps.json" I saw we're pulling the apps from that URL. In my local data/apps.json file, I noticed that app_name for Firefox for Android is fenix. We should consider(if possible) renaming app_name to firefox_android to be consistent with firefox_ios, firefox_desktop, etc.

Admittedly I used to not know what fenix refers to when I first started working on the Glean Dictionary, but after a while, I learned that fenix is our [codename for Firefox for Android](https://github.com/mozilla-mobile/fenix). You can see this mentioned in the description for fenix in `apps.json`: 

```
"app_description": "Firefox for Android (Fenix)"
```
I think leaving it as `fenix` is okay as internally people are familiar with this name.

> Line 59 in the AppList.svelte, I made any app that matches with nightly or fenix to use the browser logo. Currently it looks like only Firefox for Android satisfies that but another app can easily match nightly or fenix and this can mask bugs. Hence my suggestion above.

I addressed this in a suggestion below.

src/pages/AppList.svelte
chichi012: oh yeah! worked on it now. Thank you @Iinh

Iinh (approved):
Looks good! Thank you @chichi012!

PR #862, opened 2021-09-29
Remove last of markdown reformatting of HTML (fixes #642)

### LINKED ISSUE #642: Probe description does not support <br> HTML or list item markdown
The old metrics.md supported html tags like `<br>` and list items (lines starting with `-`) ([example](https://github.com/mozilla-mobile/fenix/blob/releases_v89.0.0/docs/metrics.md):
image

However, the probe dictionary doesn't support it ([example](https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/events_app_opened_all_startup)):
image

Without lists or the ability to add paragraphs, it makes it difficult (for me) to write information about complex probes clearly.

wlach:
To make it less likely that people will accidentally throw in an HTML tag,
we could possibly check inside the glean parser for valid HTML, where it
is specified: https://bugzilla.mozilla.org/show_bug.cgi?id=1733256

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -19,8 +19,6 @@
   export let text;
   // if inline is set, do not wrap the markdown in a paragraph -- useful for short snippets
   export let inline = true;
-  console.log(text);
-  console.log(parse(text));
 </script>
 
 {#if inline}
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -13,17 +13,14 @@
         start ? `start="${start}"` : ""
       } class="mzp-u-list-styled">${body}</${outerEl}>`;
     },
-    html(html) {
-      // convert markdown with angle brackets to html,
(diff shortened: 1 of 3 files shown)


src/components/Markdown.svelte
```diff
@@ -13,17 +13,14 @@
...
-      return `${html.replace("/<[^mark]+/g", "&lt;")}`;
-    },
   };
   use({ renderer });
 
   export let text;
   // if inline is set, do not wrap the markdown in a paragraph -- useful for short snippets
   export let inline = true;
+  console.log(text);
+  console.log(parse(text));
```
Iinh, line 23:
```suggestion
```

Iinh (approved):

src/components/Markdown.svelte
wlach: 🤦 Sorry for not catching this!

PR #882, opened 2021-10-05
Set CORS headers to allow all for /data/*

wlach:
This will allow GLAM to use the Glean Dictionary as an API (see: https://github.com/mozilla/glam/issues/1581)


Iinh (approved):
lgtm!

PR #884, opened 2021-10-06
Render event extra descriptions as non-inline

wlach:
They can contain markdown which gets stripped unless we display them as
"full". We have the space for the full description.


wlach:
Before:
image


After:
image

PR #899, opened 2021-10-12
Reference ping count for metrics of ping that does not send a client id (fixes #873)

### LINKED ISSUE #873: Links to looker explores should not reference client count for pings without a client id
A good example would be the explore for the background update ping: https://mozilla.cloud.looker.com/explore/firefox_desktop_background_update/background_update?qid=XMTEmtkFtlBVT9sCPQ9pSh&toggle=fil,vis

The client count has no meaning, because a client id is not sent with these pings:

https://dictionary.telemetry.mozilla.org/apps/firefox_desktop_background_update/pings/background-update

For these cases, we should specify the ping count measure instead of client count. The offending line is here:

https://github.com/mozilla/glean-dictionary/blob/c7a542e9f214b3b200bb35f82b8f2201818645b0/scripts/build-glean-metadata.py#L429

However, you will need to find out if the ping that a metric is a part of includes a client identifier for each case (note that metrics might be sent by more than one ping, some of which may have a client identifier and some of which may not).

@Iinh can you take this one when you have time? This is more of an edge case than #871 but we should still fix it.

See also: https://github.com/mozilla/lookml-generator/issues/253

Iinh:
For pings that do not send a client id, the Looker links for their metrics now reference measure `ping_count` instead of `client` count.

Example of pings that do not send a client id:

- [address-sync](https://deploy-preview-899--glean-dictionary-dev.netlify.app/apps/fenix/pings/addresses-sync)
https://deploy-preview-899--glean-dictionary-dev.netlify.app/apps/fenix/metrics/app_build?ping=addresses-sync

- [background-update](https://deploy-preview-899--glean-dictionary-dev.netlify.app/apps/firefox_desktop_background_update/pings/background-update)
https://deploy-preview-899--glean-dictionary-dev.netlify.app/apps/firefox_desktop_background_update/metrics/background_update_client_id

After:

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -258,7 +258,7 @@ for (app_name, app_group) in app_groups.items():
             json.dumps(app.app)
         )
 
-        pings_with_client_id = []
+        pings_with_client_id = set()
         # ping data
         for ping in app.get_pings():
             if ping.identifier not in ping_identifiers_seen:
@@ -280,7 +280,7 @@ for (app_name, app_group) in app_groups.items():
             ping_data = next(pd for pd in app_data["pings"] if pd["name"] == ping.identifier)
 
             if ping_data["include_client_id"]:
-                pings_with_client_id.append(ping_data["name"])
+                pings_with_client_id.add(ping_data["name"])
 
             # write table description (app variant specific)
             ping_name_snakecase = stringcase.snakecase(ping.identifier)


scripts/build-glean-metadata.py
```diff
@@ -258,6 +258,87 @@ def get_app_variant_description(app):
             json.dumps(app.app)
         )
 
+        pings_with_client_id = []
```
wlach, line 261: It might make slightly more sense to have a dictionary of ping names -> ping data, which we could use for both identifying client id information as well as whether a ping has been seen yet. But maybe not worth worrying about too much for now. However, at the very least we *should* use a set here (like we already do for `ping_identifiers_seen`).

```suggestion
        pings_with_client_id = set()
```

scripts/build-glean-metadata.py
```diff
@@ -258,6 +258,87 @@ def get_app_variant_description(app):
...
+                        ),
+                        app_tags,
+                        full=False,
+                    )
+                )
+
+            ping_data = next(pd for pd in app_data["pings"] if pd["name"] == ping.identifier)
+
+            if ping_data["include_client_id"]:
+                pings_with_client_id.append(ping_data["name"])
```
wlach, line 283:
```suggestion
                pings_with_client_id.add(ping_data["name"])
```

PR #903, opened 2021-10-13
Bump netlify python version to 3.8

wlach:
Since we're using the 20.04 image now, which doesn't have python 3.7.


wlach:
Just going to merge this, it's effectively a dependency update.

PR #937, opened 2021-11-02
Improve looker links for events (fixes #936)

### LINKED ISSUE #936: Looker links for VPN don't quite work
(reported by @birdsarah)

e.g. https://dictionary.telemetry.mozilla.org/apps/mozilla_vpn/metrics/sample_app_permissions_view_opened
has this link: https://mozilla.cloud.looker.com/explore/mozilla_vpn/main?qid=2SKYHvTFTv489Dc16VGFa9&toggle=vis

Which has a bunch of fields that Looker doesn't know about:

image

I think there are some Firefox for iOS / Android specific assumptions in the Looker URL generation. Probably not too difficult to fix.

wlach:
* Use funnel_analysis explore when event_counts isn't present
* Don't use channel specifier when there is only one channel
* Other minor refactoring

### DIFF
--- a/scripts/looker.py
+++ b/scripts/looker.py
@@ -118,6 +118,10 @@ def get_looker_explore_metadata_for_metric(
                     + f"&f[step_1.event]=%22{metric_name}%22"
                     + f"&f[step_1.category]=%22{metric_category}%22"
                 )
+            else:
+                # this should never happen (unless we made a mistake in getting the
+                # base looker explore link)
+                raise Exception(f"Unexpected base looker explore {base_looker_explore['name']}")
         # for counters, we can use measures directly
         if metric_type == "counter":
             looker_metric_link = (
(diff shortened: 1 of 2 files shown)


scripts/looker.py
```diff
@@ -19,95 +19,109 @@
...
-                + f"&f[events.event_name]=%22{metric_name}%22"
-                + f"&f[events.event_category]=%22{metric_category}%22"
-            )
+            if base_looker_explore["name"] == "event_counts":
+                looker_metric_link = (
+                    base_looker_explore["url"]
+                    + f"&f[events.event_name]=%22{metric_name}%22"
+                    + f"&f[events.event_category]=%22{metric_category}%22"
+                )
+            else:
```
Iinh, line 115: small nit: if there's no other event explores other than event_counts and funnel_analysis, should we be more specific here?
```suggestion
            elif base_looker_explore["name"] == "funnel_analysis":
```

Iinh (approved):
lgtm!

scripts/looker.py
wlach: We should actually assert on the `else`, now that I think about it. Will fix.

Issue #561, opened 2021-04-29
Expired metrics can be too difficult to find


wlach:
When we were annotating things today, @irrationalagent initially had trouble finding this metric:

https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/places_manager_write_query_time

This metric recently expired (about 60 days ago) so it wasn't displaying. Our default is not to show expired metrics but this might be the wrong approach-- they are already relatively clearly marked in the UI.

We discussed various approaches (e.g. saying "Not showing XX expired" metrics) but the simplest solution is probably just to display expired metrics by default in the UI again. My experience with the UI is that filtering usually filters things down to what you care about pretty fast.

PR #946, opened 2021-11-08
Improve handling of uncollected metrics

wlach:
* *Don't* show them by default
* However, when they do exist and would otherwise have matched a search term, provide some helpful text explaining what's going on and a button to reveal them

### DIFF
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -37,6 +37,7 @@
   let showUncollected;
   let topElement;
   let scrollY;
+  let totalItems;
 
   const searchIndex = new Document({
     tokenize: "forward",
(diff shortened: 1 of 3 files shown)


wlach:
Some screenshots:

image
image

wlach:
@Iinh I think this gets around the previous pitfalls that we had around uncollected metrics (people getting lost because things expired or removed). Let me know what you think!

src/components/ItemList.svelte
```diff
@@ -121,34 +122,51 @@
...
+        </p>
+        <button
+          class="mzp-c-button mzp-t-secondary mzp-t-md"
+          on:click={() => {
+            showUncollected = true;
+            updateURLState({ showUncollected });
+          }}>Show</button
+        >
+      {:else}
+        <p>Try entering less specific or alternate search terms.</p>
```
Iinh, line 166: small nit: 
```suggestion
        <p>Try entering less specific or alternative search terms.</p>
```

Iinh (approved):
I like this new way of handling uncollected metrics. We decided to show all metrics by default after this issue https://github.com/mozilla/glean-dictionary/issues/561, but this looks like a much better approach. Thanks for doing this @wlach!

PR #949, opened 2021-11-09
Fix Looker link for pings

wlach:
We accidentally turned a data structure into a list in a previous PR. See #950 for a test to make sure this doesn't happen again.


Iinh (approved):

Iinh:
I'm merging this now to review #950

PR #950, opened 2021-11-09
Add a playwright end-to-end test

wlach:
Just starting with the metrics ping for now (motivated by #949).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ orbs:
   node: circleci/node@2.0.1
   python: circleci/python@0.2.1
 jobs:
-  build-and-test:
+  frontend-build-and-test:
     executor:
       name: node/default
       tag: "12.16"
@@ -15,8 +15,13 @@ jobs:
       - run: npm install
       - run: npm run build
       - run: npm run build-storybook
+      # for playwright integration tests, we want the data files we
+      # collected in the previous step
+      - attach_workspace:
+          at: /tmp/workspace
+      - run: cp -r /tmp/workspace/data/* public/data/
       - run: npm test
-  python-build-and-test:
+  backend-build-and-test:
     executor: python/default
     steps:
       - checkout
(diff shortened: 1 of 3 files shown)


wlach:
CI should start passing once #949 is merged.

playwright.config.js
```diff
@@ -1,6 +1,6 @@
 const config = {
   webServer: {
-    command: "npm run start",
+    command: "npm run start -- --single",
```
Iinh, line 3: what is this command for?

wlach: This puts sirv into "single page application" mode, which we need for URLs like "/apps/fenix/pings/metrics" to work. I'll add a comment

https://github.com/lukeed/sirv/tree/master/packages/sirv-cli

.circleci/config.yml
```diff
@@ -15,8 +15,13 @@ jobs:
       - run: npm install
       - run: npm run build
       - run: npm run build-storybook
+      # for playwright integration tests, we want the data files we
+      # collected in the previous step
+      - attach_workspace:
+          at: /tmp/workspace
+      - run: cp -r /tmp/workspace/data/* public/data/
```
Iinh, line 22: this question is just out of curiosity: since the backend job runs before frontend, as another way of sharing data, is it possible to  `attach_workspace` in backend after we run the python script (so it should have all the files we need), then we can just run `cp -r /tmp/workspace/data/* public/data/` in the frontend job without needing to `persist_to_workspace`?

Iinh (approved):
lgtm! do you know why the 2 CI tests below are still hanging? I checked the CI log and looks like they already passed? https://app.circleci.com/pipelines/github/mozilla/glean-dictionary?branch=playwright-e2e

.circleci/config.yml
wlach: I'm mostly just following [the approach recommended by circleci](https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-data-among-jobs). I assume that files written to an attached workspace are not themselves persisted by default otherwise they would probably have suggested doing exactly what you suggest here. :) I'm guessing on circleci's side having an explicit persist step is easier for them to implement (and arguably maybe more robust, since it's more obvious when you're writing to persistent storage).

wlach: > lgtm! do you know why the 2 CI tests below are still hanging? I checked the CI log and looks like they already passed? https://app.circleci.com/pipelines/github/mozilla/glean-dictionary?branch=playwright-e2e

Those are some old jobs which we replaced in the circle config in this PR. I'm guessing circleci will stop looking for them once we merge this to main.

wlach: Oh, actually just checked and we also need to update the branch protection rules: https://github.com/mozilla/glean-dictionary/settings/branch_protection_rules

I'll do that after merging this (and ci passes)

PR #962, opened 2021-11-18
Show the "deletion-request" ping last under access

wlach:
This is specially relevant for Mozilla VPN, where there are only
two pings: main and deletion-request (and notably, no metrics ping
which would come first). 99% of the time it's the former that people
are interested in.


wlach:
Example where this is a bit frustrating: https://dictionary.telemetry.mozilla.org/apps/mozilla_vpn/metrics/app_build

Iinh (approved):

PR #967, opened 2021-11-22
Handle in-app defined tags

wlach:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1734011

### DIFF
--- a/scripts/build-glean-metadata.py
+++ b/scripts/build-glean-metadata.py
@@ -52,34 +52,37 @@ def _get_annotation(annotations_index, origin, item_type, identifier=None):
     return annotations_index.get(origin, {}).get(item_type, {}).get(identifier, {})
 
 
-def _incorporate_annotation(item, item_annotation, tag_descriptions, app=False, full=False):
+def _incorporate_annotation(item, item_annotation, app=False, full=False):
     incorporated = dict(item, has_annotation=len(item_annotation) > 0)
 
     # handle tags, which are in a slightly different format than other types
     # of annotations (and we want to apply to both full and non-full)
-    if not app:
-        if full:
-            # we want an extended tag definition for a full description
-            incorporated.update(
-                {
-                    "tags": [
-                        {"name": tag, "description": tag_descriptions[tag]}
-                        for tag in item_annotation.get("tags", [])
-                    ]
-                }
-            )
-        else:
-            # for the non-full description, we just want the names
-            incorporated.update({"tags": item_annotation.get("tags", [])})
+    if not app and item_annotation.get("tags"):
+        # annotation tags always take precedence over any tags defined in
+        # metrics.yaml
+        incorporated.update({"tags": item_annotation["tags"]})
 
-    # other annotations are only applied to the full version (not the
-    # summaries we list out in various places)
     if full:
+        # other annotations are only applied to the full version (not the
+        # summaries we list out in various places)
         for annotation_type in ["commentary", "warning"]:
             if item_annotation.get(annotation_type):
                 incorporated[annotation_type] = item_annotation[annotation_type]
 
     return incorporated
+
+
+def _expand_tags(item, tag_descriptions):
+    """
+    Expand the tags into full name/description objects (for full definitions)
+    """
+    return dict(
+        item,
+        tags=[
+            {"name": tag_name, "description": tag_descriptions[tag_name]}
+            for tag_name in item["tags"]
+        ],
+    )
 
 
 def get_resource_path(line: str) -> str:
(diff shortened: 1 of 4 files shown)


wlach:
Example: https://deploy-preview-967--glean-dictionary-dev.netlify.app/apps/burnham?itemType=metrics&page=1&search=tags%3Ascience

Old, annotations-based tags still work. E.g.: https://deploy-preview-967--glean-dictionary-dev.netlify.app/apps/fenix?page=1&search=tags%3AWebExtensions

scripts/build-glean-metadata.py
```diff
@@ -173,6 +176,12 @@ def get_app_variant_description(app):
         app = next(app for app in apps if app.app_id == app_id)
         app_is_deprecated = app.app.get("deprecated")
 
+        # app-id tags: tags specified in the annotations (and or more recent versions of an app)
+        # will always override older ones
+        for tag in app.get_tags():
+            if not app_tags.get(tag.identifier):
+                app_tags[tag.identifier] = tag.description
```
Iinh, line 183: what would be an example of this?

src/pages/PingDetail.svelte
```diff
@@ -50,6 +51,14 @@
           /></a
         >
       {/if}
+      {#each ping.tags as tag}
```
Iinh, line 54: should we check if `ping.tags` exists first?

scripts/build-glean-metadata.py
wlach: Say you had the same tags specified for Firefox Android beta and Firefox Android nightly. You would want the definition for the latter to be the one we show in the Glean Dictionary, as it's more up to date:

https://github.com/mozilla/probe-scraper/blob/main/repositories.yaml#L81

scripts/glean.py
```diff
@@ -134,11 +132,30 @@ def _set_definition(self, full_defn: dict):
...
+            sorted(
+                full_defn[self.HISTORY_KEY],
+                key=lambda x: datetime.fromisoformat(x["dates"]["last"]),
+                reverse=True,
+            )
+        )
+
+        # The canonical definition for up-to-date schemas
+        self.definition = self.definition_history[0]
+        self.definition["name"] = full_defn[self.NAME_KEY]
```
Iinh, line 157: looking at the screenshot, I saw that the new "science" tag for burnham is lower case. Since I'm used to seeing fenix tags being upper case and lower case for origin labels, UI-wise, I didn't immediately associate "science" to be a tag. I think it's probably a good idea to keep the formatting consistent here, i.e. capitalize the tag label where we can.

scripts/build-glean-metadata.py
```diff
@@ -381,8 +393,8 @@ def get_app_variant_description(app):
         )
 
     # write tag metadata (if any)
-    if app_annotation and app_annotation.get("tags"):
-        tags = [{"name": k, "description": v} for (k, v) in app_annotation["tags"].items()]
+    if app_tags:
```
Iinh, line 396: i'm not following this part very well, why do we use `app_tags` as a condition check here instead of `app_annotation and app_annotation.get("tags")`?

scripts/build-glean-metadata.py
Iinh: that makes sense. thank you for the example!

src/pages/PingDetail.svelte
wlach: I don't think such a check is needed: we should always have this property (although sometimes it will be an empty array)

scripts/glean.py
wlach: Interesting, do you think we should suggest that at the glean level? I think it will be hard to enforce conventions on tags. Another alternative would be to add a 🏷️ emoji to the end of tags (just like we do with origin labels and the 📦 emoji). I'd probably do this as a seperate PR

scripts/build-glean-metadata.py
wlach: Because app_tags can now come from two places: both the annotations (e.g. https://mozilla.github.io/glean-annotations/api.json) as well as from probeinfo.tmo (e.g. https://probeinfo.telemetry.mozilla.org/glean/burnham/tags). We merge the two representations above

scripts/glean.py
Iinh: tbh this is not a big issue, it's probably just my own mental model (uppercase = tags, lowercase = origin) and no one else's. I still think we should experiment with adding a tag emoji to make it clearer that a label is a tag. Will do it in a separate PR.

src/pages/PingDetail.svelte
wlach: The loop will just not run if array is empty. You can see an example of this behaviour here: https://deploy-preview-967--glean-dictionary-dev.netlify.app/apps/mozilla_vpn/pings/main

Iinh: ah you're right!

Iinh (approved):
LGTM! Thanks for your patience in answering my questions as I tried to understand the code.

PR #970, opened 2021-11-24
Fix generation of Looker URLs for apps with only one app_id

wlach:
There was a broken assumption in the way we were generating the
looker URLs that caused invalid content to be generated. While we're
at it, let's use furl to generate/augment our URLs: this should make
it harder to make this kind of mistake in the future.

### DIFF
--- a/scripts/looker.py
+++ b/scripts/looker.py
@@ -1,8 +1,8 @@
 import json
-import urllib.parse
 
 import stringcase
 from bigquery import get_bigquery_column_name, get_bigquery_ping_table_name
+from furl import furl
 from glean import GLEAN_DISTRIBUTION_TYPES
 
 SUPPORTED_LOOKER_METRIC_TYPES = GLEAN_DISTRIBUTION_TYPES | {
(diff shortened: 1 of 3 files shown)


scripts/looker.py
```diff
@@ -32,32 +33,30 @@ def _get_looker_ping_explore(
 ):
     ping_name_snakecase = stringcase.snakecase(ping_name)
     if _looker_explore_exists(looker_namespaces, app_name, ping_name_snakecase):
-        url = f"https://mozilla.cloud.looker.com/explore/{app_name}/{ping_name_snakecase}"
+        url = furl("https://mozilla.cloud.looker.com/explore/{app_name}/{ping_name_snakecase}")
         # if there are multiple channels, we need a channel identifier
         if len(app_group["app_ids"]) > 1 and app_channel:
-            channel_identifier = "mozdata." + table_name.replace("_", "%5E_")
-            url += f"?f[{ping_name_snakecase}.channel]={channel_identifier}"
```
wlach, line 39: This was the source of the problem: later in the code we had the assumption that this *was* the case, and we were just appending further url parameters onto the end with `&`. Using furl fixes that assumption.

wlach:
Example of this failing before:

https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/browser_link_open_newwindow_restriction

Fixed:

https://deploy-preview-970--glean-dictionary-dev.netlify.app/apps/firefox_desktop/metrics/browser_link_open_newwindow_restriction

I have a PR in progress to add tests so this doesn't recur

Iinh (approved):

PR #972, opened 2021-11-25
Refactor ETL project structure and add tests (fixes #680)

### LINKED ISSUE #680: Consider a better internal structure than just dumping stuff in scripts/
We might be getting to the point where we want a proper python project for the ETL code, with unit tests and a sensible directory structure. [lookml-generator](https://github.com/mozilla/lookml-generator/) might be a good model to use here.

This is something a contributor could potentially take on, if they're interested. Main skills required would be python.

wlach:
This turns the Glean Dictionary into more of a standard python
package, like lookml-generator. Also add a single test for the Looker
ETL (will add more in the future, but this PR is big enough as it is)


etl_tests/test_looker.py
```diff
@@ -0,0 +1,41 @@
+import pytest
```
wlach, line 1: @Iinh could you review this part? Ideally we would have more tests here, but I thought it would be better to do that as a followup (since this is already a fairly large change).

Iinh: this part looks good to me!

Iinh (approved):

relud (approved):
r+wc

Issue #682, opened 2021-06-17
We should stop using Protocol as our main styling library


Iinh:
See [@wlach's comments](https://github.com/mozilla/glean-dictionary/pull/556#issuecomment-863300778):

> I'm concerned about the interaction of protocol and this site -- based on the error output I'm seeing (e.g. complaints about unused css selectors) it looks like it's importing a very large chunk of protocol for each svelte component. I'm beginning to think we should probably stop using protocol as a library-- while the idea of sharing styles across Mozilla properties is appealing, it's proving difficult to adapt the assets (which were created to serve the needs of a completely different site-building technology) to the needs of our particular application and I would expect ongoing issues from continuing to attempt to do this. 
> 
> I'd probably vote to keep using scss, the protocol tokens and whatever styles make sense to copy from the library but yeah, removing our explicit dependency on the package makes sense to me.

Dexterp37:
@Iinh does this still apply? Looks like we can now [only get what we're interested in](https://github.com/mozilla/protocol#usage)

PR #973, opened 2021-11-26
Protocol/SCSS refactoring

wlach:
* Move away from making scss the default, instead specify it manually
  ("defaults" are now discouraged: https://github.com/sveltejs/svelte-preprocess/issues/362)
* Import only the parts of protocol we actually need for each component,
  this speeds up the build by a fair chunk (50.25 -> 18.85 seconds on
  my MacBook) and helps pave the way to #682.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -14,11 +14,8 @@ module.exports = {
       ...svelteLoader.options,
       preprocess: sveltePreprocess({
         postcss: true,
-        defaults: {
-          style: "scss",
-        },
         scss: {
-          prependData: `@import '@mozilla-protocol/core/protocol/css/protocol.scss';`,
+          prependData: `@import 'src/protocol-tokens.scss';`,
         },
         postcss: {
           plugins: [autoPrefixer],
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -71,11 +71,8 @@ export default {
       // a separate file - better for performance
       preprocess: sveltePreprocess({
         postcss: true,
-        defaults: {
-          style: "scss",
-        },
         scss: {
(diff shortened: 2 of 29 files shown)


wlach:
This also seems to reduce the size of the generated CSS from 250k uncompressed (15k compressed) to 50k uncompressed (7.5k compressed). Not a gigantic improvement, but still nice.

wlach:
@Iinh Now that I think about it, I think this might duplicate some of the work you did when working on the sveltekit port last summer. Hopefully this will make it easier to reactivate that when the time comes.

Iinh (approved):
This is great, thanks @wlach! We should probably refactor svelte-preprocess scss options in Storybook [main.js](https://github.com/mozilla/glean-dictionary/blob/275df98780ac80b9308b5e125f0822555ce030b0/.storybook/main.js#L15) too, but maybe in a separate PR.

wlach:
> This is great, thanks @wlach! We should probably refactor svelte-preprocess scss options in Storybook [main.js](https://github.com/mozilla/glean-dictionary/blob/275df98780ac80b9308b5e125f0822555ce030b0/.storybook/main.js#L15) too, but maybe in a separate PR.

Oh good catch! Let's do that as part of this one, it's a straightforward extension of what we're doing here. I'll merge after doing that (and testing it) if that's ok.

wlach:
The storybook change reduced build times on netlify from ~6 minutes to ~4 minutes, and further improved the speed of the circleci job (from ~1 minute to ~30 seconds) 🎉

wlach:
Storybook still looks good: https://deploy-preview-973--glean-dictionary-dev.netlify.app/storybook/

PR #999, opened 2021-12-06
Show expiry information for Firefox desktop (fixes #984)

### LINKED ISSUE #984: Show expiry information for Firefox
Firefox desktop uses version numbers instead of dates for expiry. For example:

https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/browser_engagement_uri_count

@badboy and I discussed this a bit on #glean-dictionary:

@wlach:

> It looks like Firefox metrics have a different notion of expiry (by "release") than that of other applications e.g. https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/browser_link_open_newwindow_restriction
> 
> How should the Glean Dictionary handle these? The simplest solution I can think of would be to use the product details API to map releases to dates (as we can): https://product-details.mozilla.org/1.0/firefox.json

@badboy:

> Yes. build date is not good for firefox, glean_parser allows to change how expiry is defined and in fog we use the firefox version, just as legacy telemetry does (also ensures that tooling keeps working, in both directions of history, that is you can build old firefox without expiring metrics and you can test newer versions and see which ones expire)

> I like the idea of mapping it to actual releases.
> It has some problems though:
> - metrics exist in nightly, so there they will expire ~8 weeks before they do in release (because that's the time it takes for the versions to rollover). product details has only release details
> - product details has releases for the past, not the future.

--

Tentative proposal:

* In our python scripts, get a list of release dates from the above mentioned product details API (https://product-details.mozilla.org/1.0/firefox.json)
* Add an `expiry_date` field to the outputted JSON which is guaranteed to be a date. For most applications, this will just be the `expires` field. For Firefox, we will use the mapping above.
* Replace "expires" with a markdown string that we can render in the Glean Dictionary. For most applications, keep this the same (since they're just dates), but for Firefox we can replace it with something like: `100 (latest release is [94](https://wiki.mozilla.org/Release_Management/Calendar))`. Hopefully that will be enough information for people to reason about nightly and beta releases as well.

Thoughts?

Iinh:
Fixes #984.

Example: https://deploy-preview-999--glean-dictionary-dev.netlify.app/apps/firefox_desktop/metrics/browser_engagement_uri_count

Screen Shot 2021-12-06 at 3 29 19 PM

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/etl/__main__.py
+++ b/etl/__main__.py
@@ -9,6 +9,7 @@ import stringcase
 import yaml
 
 from .bigquery import get_bigquery_column_name, get_bigquery_ping_table_name
+from .expiry import get_expiry_date, get_expiry_text
 from .glam import SUPPORTED_GLAM_METRIC_TYPES, get_glam_metadata_for_metric
 from .glean import GleanApp
 from .looker import get_looker_explore_metadata_for_metric, get_looker_explore_metadata_for_ping
@@ -290,7 +291,7 @@ def main():
                             if "extra_keys" in metric.definition
                             else None,
                             type=metric.definition["type"],
-                            expiry_date=get_expiry_date(
+                            expires=get_expiry_date(
                                 metric.definition["expires"], app_name, product_details
                             ),
                             expiry_text=get_expiry_text(
@@ -322,7 +323,7 @@ def main():
                                 send_in_pings=list(metric.definition["send_in_pings"]),
                                 repo_url=app.app["url"],
                                 variants=[],
-                                expiry_date=base_definition["expiry_date"],
+                                expires=base_definition["expires"],
                                 expiry_text=base_definition["expiry_text"],
                             ),
                             metric_annotation,
--- a/etl/__main__.py
+++ b/etl/__main__.py
@@ -22,6 +22,10 @@ ANNOTATIONS_URL = os.getenv(
 NAMESPACES_URL = os.getenv(
     "NAMESPACES_URL", "https://raw.githubusercontent.com/mozilla/looker-hub/main/namespaces.yaml"
 )
+FIREFOX_PRODUCT_DETAIL_URL = os.getenv(
+    "FIREFOX_PRODUCT_DETAIL_URL",
+    "https://product-details.mozilla.org/1.0/firefox_history_major_releases.json",
+)
 
 # Priority for getting metric data (use the later definitions of nightly over release)
 METRIC_CHANNEL_PRIORITY = {"nightly": 1, "beta": 2, "release": 3, "esr": 4}
@@ -107,6 +111,7 @@ def main():
     # Pull down the annotations
     annotations_index = requests.get(ANNOTATIONS_URL).json()
     looker_namespaces = yaml.safe_load(requests.get(NAMESPACES_URL).text)
+    product_details = requests.get(FIREFOX_PRODUCT_DETAIL_URL).json()
 
     # Then, get the apps we're using
     apps = [app for app in GleanApp.get_apps()]
@@ -285,7 +290,12 @@ def main():
                             if "extra_keys" in metric.definition
                             else None,
                             type=metric.definition["type"],
-                            expires=metric.definition["expires"],
+                            expiry_date=get_expiry_date(
+                                metric.definition["expires"], app_name, product_details
+                            ),
+                            expiry_text=get_expiry_text(
+                                metric.definition["expires"], app_name, product_details
+                            ),
                         ),
                         metric_annotation,
                     )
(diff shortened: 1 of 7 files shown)


etl/__main__.py
```diff
@@ -101,6 +105,19 @@ def get_app_variant_description(app):
     return description
 
 
+def get_expiry_date(expiry, app_name):
```
wlach, line 108: This would probably be better named `get_expiry_text`.

Could you move this to a separate file called `expiry.py` and write some quick tests for it? You can probably model off of https://github.com/mozilla/glean-dictionary/blob/main/etl_tests/test_looker.py

(we won't want to query the network API in the test, but I can help with that part)

etl/__main__.py
```diff
@@ -101,6 +105,19 @@ def get_app_variant_description(app):
...
+    latest_release_version = [*requests.get(FIREFOX_PRODUCT_DETAIL_URL).json()][-1]
```
wlach, line 109: We should cache the results of the request so we only need to call it once.

etl/__main__.py
```diff
@@ -101,6 +105,19 @@ def get_app_variant_description(app):
...
+
+    desktop_expiry_description = f"{expiry}. Latest release is \
+    [{latest_release_version}](https://wiki.mozilla.org/Release_Management/Calendar)"
+
+    if expiry == "never":
+        return expiry
```
wlach, line 115: Pretty sure this isn't needed.

```suggestion
```

etl/__main__.py
```diff
@@ -101,6 +105,19 @@ def get_app_variant_description(app):
     return description
 
 
+def get_expiry_date(expiry, app_name):
+    latest_release_version = [*requests.get(FIREFOX_PRODUCT_DETAIL_URL).json()][-1]
+
+    desktop_expiry_description = f"{expiry}. Latest release is \
+    [{latest_release_version}](https://wiki.mozilla.org/Release_Management/Calendar)"
```
wlach, line 112: I would move this into the `if` statement below

etl/__main__.py
```diff
@@ -283,7 +300,7 @@ def main():
                             if "extra_keys" in metric.definition
                             else None,
                             type=metric.definition["type"],
-                            expires=metric.definition["expires"],
+                            expires=get_expiry_date(metric.definition["expires"], app_name),
```
wlach, line 303: I think we also want to calculate an explicit `expiry_date` property which we can use to determine when something is expired. Probably worth writing a seperate method (distinct from `get_expiry_text`) to do just that.

wlach (changes requested):
Great start! Unfortunately we won't get expiry warnings for firefox desktop as things stand. Have some suggestions on how to make that happen, let me know if you have any questions.

etl/expiry.py
```diff
@@ -0,0 +1,29 @@
...
+    except ValueError:
+        return False
+
+
+def get_expiry_date(expiry, app_name, product_details):
+    if app_name == "firefox_desktop":
+        return product_details.get(f"{expiry}.0")
+    if validate(expiry):
+        return expiry
+    return None
```
wlach, line 17:
```suggestion
    return expiry if _validate(expiry) else return None
```

etl/expiry.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def get_expiry_text(expiry, app_name, product_details):
+    latest_release_version = [*product_details][-1]
+
+    if expiry == "never" or expiry is None:
+        return expiry
+    if app_name == "firefox_desktop":
```
wlach, line 25:
```suggestion
    elif app_name == "firefox_desktop":
```

wlach (approved):

etl/expiry.py
```diff
@@ -0,0 +1,27 @@
...
+        datetime.datetime.strptime(date_text, "%Y-%m-%d")
+        return True
+    except ValueError:
+        return False
+
+
+def get_expiry_date(expiry, app_name, product_details):
+    if app_name == "firefox_desktop":
+        return product_details.get(f"{expiry}.0")
+    return expiry if _validate(expiry) else return None
```
wlach, line 15:
```suggestion
    return expiry if _validate(expiry) else None
```

etl_tests/test_expiry.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+
+from etl.expiry import get_expiry_date, get_expiry_text, validate
+
+
+def test_validate():
+    assert validate("2021-12-31") is True
```
wlach, line 7: We can probably skip this method, since it's not a public method, just a helper function

PR #1001, opened 2021-12-10
Privacy updates

wlach:
* Be certain that we're not loading or using google analytics if
  the user specifies Do Not Track (in accordance with our privacy
  policies)
* Update link to privacy policy

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -82,7 +82,9 @@ export default {
     // only use google analytics on production builds
     replace({
       __GOOGLE_ANALYTICS_ID__:
-        process.env.CONTEXT === "production" && process.env.GOOGLE_ANALYTICS_ID,
+        (process.env.CONTEXT === "production" &&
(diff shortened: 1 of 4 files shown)


rollup.config.js
```diff
@@ -82,7 +82,9 @@ export default {
     // only use google analytics on production builds
     replace({
       __GOOGLE_ANALYTICS_ID__:
-        process.env.CONTEXT === "production" && process.env.GOOGLE_ANALYTICS_ID,
+        (process.env.CONTEXT === "production" &&
+          process.env.GOOGLE_ANALYTICS_ID) ||
+        "",
```
Iinh, line 87: why do we need to add `|| ""` here?

wlach: Otherwise `__GOOGLE_ANALYTICS_ID__` gets replaced as `false`, which is not what we want here (since the conditional logic to turn it off altogether for a build looks for an empty string).

PR #1031, opened 2021-12-29
Use Glean Annotations to get app logos and tags

wlach:
This replaces a bunch of hardcoded logic inside the Glean Dictionary
and gives us more flexibility in the future.
Depends on https://github.com/mozilla/glean-annotations/pull/98

PR #1032, opened 2021-12-30
Slightly reduce the size of generated json by omitting spaces

wlach:
This brings down the size of the firefox legacy metrics search
function from 1053888 bytes down to 1017557 bytes. Not a huge
improvement, but why not.

### DIFF
--- a/etl/glean_etl.py
+++ b/etl/glean_etl.py
@@ -1,4 +1,3 @@
-import json
 import os
 
 import requests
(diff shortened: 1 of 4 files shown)


wlach:
Marking this one as a draft, as I'd rather not have it block #1031

wlach:
Let's pick this up after #1033 lands

wlach:
I can't push to this branch anymore, going to close and refile

PR #1086, opened 2022-01-19
Make sure latest metric definition in history is always used

wlach:
We were using the *earliest*. For some reason, this was still giving
correct looking results in production although I'm not sure why (I can
easily reproduce the problem in a unit test).

This allows tags to appear for Firefox Desktop.

### DIFF
--- a/etl/glean.py
+++ b/etl/glean.py
@@ -77,6 +77,7 @@ class GleanMetric(GleanObject):
             self.definition["send_in_pings"] = set(pings)
 
     def _set_definition(self, full_defn: dict):
+        # sort from latest to earliest
         self.definition_history = list(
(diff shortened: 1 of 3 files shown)


wlach:
> We were using the _earliest_. For some reason, this was still giving correct looking results in production although I'm not sure why (I can easily reproduce the problem in a unit test).

I'm *really* baffled by this, but have run out of time to look into why. I think this should be safe to merge (assuming things like right in the deploy preview)

wlach:
This includes #1087, just to validate the PR. Merge that one before this one.

Iinh (approved):
LGTM 👍 Thanks @wlach for working on this!

PR #1098, opened 2022-01-28
Slightly reduce the size of generated json by omitting spaces

wlach:
This brings down the size of the firefox legacy metrics search
function from 1053888 bytes down to 1017557 bytes. Not a huge
improvement, but why not.

### DIFF
--- a/etl/glean_etl.py
+++ b/etl/glean_etl.py
@@ -1,4 +1,3 @@
-import json
 import os
 
 import requests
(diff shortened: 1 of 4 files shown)


wlach:
@Iinh Ok, I think this one is ready for review, this was the last of my unlanded PRs :)

Iinh (approved):
👍

PR #1142, opened 2022-02-16
Make sure that the expiry input is always a string

Iinh:
Our ETL script currently fails with this error (you can also see the CI failing error [here](https://app.circleci.com/pipelines/github/mozilla/glean-dictionary/3250/workflows/1c9f2e12-8d35-4042-842a-1b259ac1cf80/jobs/6262)):

```
  File "/root/project/etl/expiry.py", line 6, in _validate
    datetime.datetime.strptime(date_text, "%Y-%m-%d")
TypeError: strptime() argument 1 must be str, not int

Exited with code exit status 1
```

which is probably from a malformed input for the expiry information. As a quick fix for now, let's make sure that we always convert the date input to string before we turn it into a datetime object. 

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/etl/expiry.py
+++ b/etl/expiry.py
@@ -3,7 +3,7 @@ import datetime
 
 def _validate(date_text):
     try:
-        datetime.datetime.strptime(date_text, "%Y-%m-%d")
+        datetime.datetime.strptime(str(date_text), "%Y-%m-%d")
         return True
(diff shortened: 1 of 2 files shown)


alekhyamoz (approved):
lgtm

PR #1162, opened 2022-03-02
Handle stripLinks edge case (when input is undefined)

Iinh:
This URL, https://dictionary.telemetry.mozilla.org/apps/fenix?page=1&search=fog, which searches for any metric names with "fog", failed. The reason is because we currently don't provide any FOG tag description in Fenix metadata yet, so certain input for the stripLinks method (which is used to strip URLs from the tag description) was `undefined`. This PR adds a condition to make sure that we only strip links if the input exists.

Working example: https://deploy-preview-1162--glean-dictionary-dev.netlify.app/apps/fenix?page=1&search=fog

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/formatters/markdown.js
+++ b/src/formatters/markdown.js
@@ -1,3 +1,3 @@
 export function stripLinks(text) {
-  return text.replace(/\[(.*)\]\(.*\)/, "$1");
+  return (text && text.replace(/\[(.*)\]\(.*\)/, "$1")) || text;
 }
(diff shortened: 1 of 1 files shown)


src/formatters/markdown.js
```diff
@@ -1,3 +1,3 @@
 export function stripLinks(text) {
-  return text.replace(/\[(.*)\]\(.*\)/, "$1");
+  return (text && text.replace(/\[(.*)\]\(.*\)/, "$1")) || text;
```
brizental, line 2:
```suggestion
  return text?.replace(/\[(.*)\]\(.*\)/, "$1"));
```

brizental (approved):

PR #1163, opened 2022-03-02
Consolidate legacy telemetry and FOG search API

Iinh:
The Glean Dictionary search service currently powers the search bar in GLAM. We have one search API per product:

- the legacy telemetry API [dictionary.telemetry.mozilla.org/api/v1/metrics_search_firefox_legacy?search=](https://dictionary.telemetry.mozilla.org/api/v1/metrics_search_firefox_legacy?search=techno) is used for desktop search
- [dictionary.telemetry.mozilla.org/api/v1/metrics_search_fenix?search=](dictionary.telemetry.mozilla.org/api/v1/metrics_search_fenix?search=) for Fenix search

As we're looking into adding FOG to GLAM, we need a way to let users search through FOG and legacy probes. One way to do it is to have 2 API calls each time a desktop search is performed, one to the legacy telemetry API and one to the FOG API. However, I propose that it would be better that on top of the existing APIs that we use per-product, we also create a single API that includes data for both legacy and FOG probes. This API will be used by the GLAM search for all desktop probes (legacy + FOG included). 

Working example of this new API:

https://deploy-preview-1163--glean-dictionary-dev.netlify.app/api/v1/metrics_search_fog_and_legacy?search=fog (searching for "fog" returns results of both legacy telemetry and FOG probes)

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/etl/firefox_legacy_etl.py
+++ b/etl/firefox_legacy_etl.py
@@ -78,4 +78,3 @@ def write_firefox_legacy_metadata(output_dir, functions_dir):
     open(os.path.join(functions_dir, "metrics_search_fog_and_legacy.js"), "w").write(
         create_metrics_search_js(probe_summary.values(), app_name="fog_and_legacy", legacy=True)
     )
-
--- a/etl/firefox_legacy_etl.py
+++ b/etl/firefox_legacy_etl.py
@@ -69,7 +69,13 @@ def write_firefox_legacy_metadata(output_dir, functions_dir):
         with open(os.path.join(probe_output_directory, f"data_{probe_name}.json"), "w") as f:
             json.dump(probe_metadata, f)
 
-    # write a search index for the app
+    # write a search index for legacy telemetry data
     open(os.path.join(functions_dir, "metrics_search_firefox_legacy.js"), "w").write(
         create_metrics_search_js(probe_summary.values(), legacy=True)
     )
+
+    # write a search index for legacy telemetry + FOG data
+    open(os.path.join(functions_dir, "metrics_search_fog_and_legacy.js"), "w").write(
+        create_metrics_search_js(probe_summary.values(), app_name="fog_and_legacy", legacy=True)
+    )
+
--- a/etl/glean_etl.py
+++ b/etl/glean_etl.py
@@ -483,7 +483,7 @@ def write_glean_metadata(output_dir, functions_dir, app_names=None):
         )
 
         # export FOG data to a separate file for the FOG + legacy search index
-        if app_name =="firefox_desktop": 
+        if app_name == "firefox_desktop":
             open(os.path.join(functions_dir, f"metrics_search_fog.js"), "w").write(
                 create_metrics_search_js(app_metrics.values(), app_name="fog", legacy=False)
             )
--- a/etl/glean_etl.py
+++ b/etl/glean_etl.py
@@ -479,8 +479,14 @@ def write_glean_metadata(output_dir, functions_dir, app_names=None):
 
         # write a search index for the app
         open(os.path.join(functions_dir, f"metrics_search_{app_name}.js"), "w").write(
-            create_metrics_search_js(app_metrics.values(), legacy=False)
+            create_metrics_search_js(app_metrics.values(), app_name, legacy=False)
         )
+
+        # export FOG data to a separate file for the FOG + legacy search index
+        if app_name =="firefox_desktop": 
+            open(os.path.join(functions_dir, f"metrics_search_fog.js"), "w").write(
+                create_metrics_search_js(app_metrics.values(), app_name="fog", legacy=False)
+            )
 
     # Write out a list of app groups (for the landing page)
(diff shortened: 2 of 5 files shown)


etl/glean_etl.py
```diff
@@ -479,9 +479,15 @@ def write_glean_metadata(output_dir, functions_dir, app_names=None):
 
         # write a search index for the app
         open(os.path.join(functions_dir, f"metrics_search_{app_name}.js"), "w").write(
```
brizental, line 481: nit: Avoid repeting this below by doing something like: 

```python
file_suffix = "fog" if app_name == "firefox_desktop" else app_name
open(os.path.join(functions_dir, f"metrics_search_{file_suffix}.js"), "w").write(
```

etl/firefox_legacy_etl.py
```diff
@@ -69,7 +69,12 @@ def write_firefox_legacy_metadata(output_dir, functions_dir):
         with open(os.path.join(probe_output_directory, f"data_{probe_name}.json"), "w") as f:
             json.dump(probe_metadata, f)
 
-    # write a search index for the app
+    # write a search index for legacy telemetry data
```
brizental, line 72: Why do we need to need keep the search that only includes legacy telemetry data?

etl/search.py
```diff
@@ -22,4 +28,19 @@ def create_metrics_search_js(metrics, legacy=False):
...
 
+    if app_name == "fog":
+        # since the FOG data will be consolidated with legacy data as one API
+        # we add a "fog" property to each metric to differentiate between
+        # FOG and legacy probes, for easier visual treatment implementation in GLAM
+        for metric_val in metric_data.values():
+            metric_val["fog"] = True
+        return FOG_DATA_JS_TEMPLATE.render(
+            metric_data=dump_json(metric_data), legacy=dump_json(legacy)
+        )
```
brizental, line 39: What if the user is searching for FOG only data? Wouldn't this break that use case? IIUC we are not creating a search template for that case.

brizental (commented):
Hey Linh, I have many questions as I am completely new to this part of the codebase. 

I do agree with your assesment that consolidating the API into one is better than making two API calls everytime.

etl/firefox_legacy_etl.py
Iinh: We're creating the legacy + fog API for GLAM only because of its specific use case. I wanted to still keep the APIs for individual products for other use cases as that was the original intention behind the Glean Dictionary search service: to let users search through metrics for each product.

etl/search.py
```diff
@@ -8,8 +8,14 @@
...
 
+FOG_AND_LEGACY_SEARCH_JS_TEMPLATE = jinja2.Template(
+    open(Path(__file__).resolve().parent / "fog_and_legacy.js.tmpl").read()
+)
+
+FOG_DATA_JS_TEMPLATE = jinja2.Template(open(Path(__file__).resolve().parent / "fog.js.tmpl").read())
+
 
-def create_metrics_search_js(metrics, legacy=False):
+def create_metrics_search_js(metrics, app_name=None, legacy=False):
```
brizental, line 18: Can we add a docstring here explaining what this function does and calling out that "fog" and "fog_an_legacy" are special values and will result in different behaviour of the function? 

Another option would be to create separate functions for the special cases instead of having the special value pattern. (And add docstrings for each of them).

etl/firefox_legacy_etl.py
brizental: > that was the original intention behind the Glean Dictionary search service: to let users search through metrics for each product.

Is this documented somewhere? If not, please file a follow-up bug to do that.

brizental (changes requested):
Thanks for bearing with me in the review here. I am confident I understand the changes now and have just left some comments to try to make the code a bit more accessible for people with less context. Otherwise, LGTM.

etl/search.py
```diff
@@ -22,4 +28,19 @@ def create_metrics_search_js(metrics, legacy=False):
         if metric_val.get("expires") == "never":
             del metric_val["expires"]
 
+    if app_name == "fog":
+        # since the FOG data will be consolidated with legacy data as one API
+        # we add a "fog" property to each metric to differentiate between
+        # FOG and legacy probes, for easier visual treatment implementation in GLAM
+        for metric_val in metric_data.values():
+            metric_val["fog"] = True
```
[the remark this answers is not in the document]
Iinh: `glean` is definitely more semantic than `fog`. I'll make this change!

etl/firefox_legacy_etl.py
Iinh: It's documented [here](https://github.com/mozilla/glean-dictionary#search-service), albeit not very thorough.

etl/glean_etl.py
Iinh: Yes that's exactly right. Thank you for your patience reading through this work and trying to understand it!

PR #1227, opened 2022-04-16
Fix bug where non-app metric annotations were not recognized

wlach:
Example: https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/glean_error_preinit_tasks_overflow


Iinh (approved):
Thanks @wlach!

PR #1461, opened 2022-10-17
Ensure jwe, labeled_rate, url, text fields show e.g. `url`, not `url2`

Iinh:
This fix addresses:
1. For metric types `jwe`, `labeled_rate`, `url`, `text`, show e.g. `url` for metric field, not `url2`
2. Since for live and stable tables, the metric field has the suffix 2, add a note to clarify so.


Iinh:
> This is exactly what we want, but unfortunately the stable table views don't do this mapping like I expected. Let me throw up a PR with that change and then we can proceed with this one.

Hi @fbertsch, any updates on ⬆️?

Dexterp37:
> > This is exactly what we want, but unfortunately the stable table views don't do this mapping like I expected. Let me throw up a PR with that change and then we can proceed with this one.
> 
> Hi @fbertsch, any updates on ⬆️?

I talked to @fbertsch , the relevant information are in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1741487#c0). Unfortunately the old PR is no longer available, so we need somebody to push this work forward before tackling this.

PR #1466, opened 2022-10-20
Enable filtering by metric types

### LINKED ISSUE #1462: Enable filtering by Glean metric type
When I use the Glean dictionary and I don't have a specific name for a metric, I find myself wanting to restrict skimming through the metrics only to specific types, e.g. events.

It would be useful to have some way to expose such a functionality. 

"As a user, I want to be able to only see metrics of a given type (e.g. events) to search more effectively if I don't know specifically what I'm looking for"

Iinh:
Allow users to search explicitly by metric types using special syntax `type:<metric type>` (similar to how we filter by `tags:` or `origin:`).  Fixes #1462.

For example, `type: event` in the search bar will return only results of metric type `event`. See: https://deploy-preview-1466--glean-dictionary-dev.netlify.app/apps/fenix?page=1&search=type%3Aevent

Also, make it more visible that this functionality is available by adding instructional text in the search bar placeholder.

CleanShot 2022-10-20 at 18 46 33@2x

### DIFF
--- a/src/components/FilterInput.svelte
+++ b/src/components/FilterInput.svelte
@@ -29,4 +29,7 @@
     border-radius: 15px;
     border: 1px solid #000;
   }
+  input::placeholder {
+    font-size: 13px;
+  }
 </style>
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -142,7 +142,9 @@
     </span>
   {/if}
   {#if showFilter}
-    <FilterInput placeHolder="Search {itemType}" />
+    <FilterInput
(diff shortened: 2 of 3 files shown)


Iinh:
Hi @Dexterp37 would you be able to review this short PR? Thanks in advance!

Dexterp37 (approved):
I absolutely ❤️ this. Amazing work! :shipit: 

A follow-up to this would be to make the placeholder text somehow available as a tooltip or something, so that I don't have to delete the text if I forget the syntax :-P

PR #1512, opened 2022-11-14
Fix displaying expiry information when expiry is a version number

### LINKED ISSUE #1497: Support version-based as well as date expiry
Currently [expiry version numbers are supported for `Firefox Desktop` but other apps are assumed to have `expiry` as a date. ](https://github.com/mozilla/glean-dictionary/blob/cbda56f858db72b0cfce4eaa493843a459c4ffc6/etl/expiry.py#L13) However the current state is that `expiry` can be one or another:

>by @chutten: 
 _`expiry` is a version in many (perhaps even most) cases, but can also be a date (and these can be mixed within the same project)_

Lack of such support results in the `expiry` field being `null` for certain metrics such as `tracking_protection.tcp_cfr_shown`, and because GLAM defaults to `expire: never` in the absence of the field, we end up with inconsistencies between [GLAM](https://glam.telemetry.mozilla.org/fenix/probe/tracking_protection_tcp_cfr_shown/explore?aggType=avg&currentPage=1&ping_type=*) and the [metric's definition ](https://github.com/mozilla-mobile/fenix/blob/ac64ec9efffcf15ff0019a59f8e6a7b18fab5f70/app/metrics.yaml#L4862-L4877)

Iinh:
The dictionary is marking some metrics as expired when they're not. For example, searching for `tracking_protection` metrics only yields one metric that is currently active, when there are at least 13 metrics that match the search term.

https://glean-dictionary-dev.netlify.app/apps/fenix?page=1&search=tracking_protection

CleanShot 2022-11-14 at 07 56 15@2x

To fix this, we need to address the case when expiry information is a version number.

This PR includes:
- Get the latest release version from the product API
- Then, if the expiry is an integer, compare it to the latest release version to determine if it has expired.

Examples

Correctly showing all the metrics that have not expired: [link](http://deploy-preview-1512--glean-dictionary-dev.netlify.app/apps/fenix?page=1&search=tracking_protection)

CleanShot 2022-11-14 at 08 12 27@2x

On the metric page, the correct expiry info is shown: [link](http://deploy-preview-1512--glean-dictionary-dev.netlify.app/apps/fenix/metrics/preferences_etp_custom_cookies_selection)

CleanShot 2022-11-14 at 08 12 44@2x

Fixes #1497.

### DIFF
--- a/etl/expiry.py
+++ b/etl/expiry.py
@@ -15,7 +15,7 @@ def get_expiry_text(expiry, app_name, product_details):
 
     if expiry == "never" or expiry is None:
         return expiry
-    elif app_name == "firefox_desktop":
+    if type(expiry) == int and (app_name in ["firefox_desktop", "fenix", "firefox_ios"]):
         return f"{expiry}. Latest release is \
      [{latest_release_version}](https://wiki.mozilla.org/Release_Management/Calendar)."
 
--- a/etl/expiry.py
+++ b/etl/expiry.py
@@ -16,6 +16,9 @@ def get_expiry_text(expiry, app_name, product_details):
     if expiry == "never" or expiry is None:
         return expiry
     if type(expiry) == int and (app_name in ["firefox_desktop", "fenix", "firefox_ios"]):
+        # We only have expiry versions for Firefox Desktop and Mobile currently.
+        # Let's update this when version-based expiry are available for other apps.
+        # See: https://github.com/mozilla/glean-dictionary/issues/1513
         return f"{expiry}. Latest release is \
      [{latest_release_version}](https://wiki.mozilla.org/Release_Management/Calendar)."
 
--- a/etl/glean_etl.py
+++ b/etl/glean_etl.py
@@ -127,6 +127,7 @@ def write_glean_metadata(output_dir, functions_dir, app_names=None):
     annotations_index = requests.get(ANNOTATIONS_URL).json()
     looker_namespaces = yaml.safe_load(requests.get(NAMESPACES_URL).text)
     product_details = requests.get(FIREFOX_PRODUCT_DETAIL_URL).json()
+    latest_fx_release_version = list(product_details)[-1]
 
     # Then, get the apps we're using
     apps = [app for app in GleanApp.get_apps()]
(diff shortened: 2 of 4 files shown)


etl/expiry.py
```diff
@@ -15,7 +15,7 @@ def get_expiry_text(expiry, app_name, product_details):
 
     if expiry == "never" or expiry is None:
         return expiry
-    elif app_name == "firefox_desktop":
+    if type(expiry) == int and (app_name in ["firefox_desktop", "fenix", "firefox_ios"]):
```
badboy, line 18: Does this mean we need to update it here when other apps switch to version based expiry?

etl/expiry.py
```diff
@@ -15,7 +15,7 @@ def get_expiry_text(expiry, app_name, product_details):
...
         return f"{expiry}. Latest release is \
      [{latest_release_version}](https://wiki.mozilla.org/Release_Management/Calendar)."
```
badboy, line 20: That link is not really perfectly accurate for mobile.

etl/glean_etl.py
```diff
@@ -127,6 +127,7 @@ def write_glean_metadata(output_dir, functions_dir, app_names=None):
     annotations_index = requests.get(ANNOTATIONS_URL).json()
     looker_namespaces = yaml.safe_load(requests.get(NAMESPACES_URL).text)
     product_details = requests.get(FIREFOX_PRODUCT_DETAIL_URL).json()
+    latest_fx_release_version = list(product_details)[-1]
```
badboy, line 130: Oh.
While mobile apps follow _roughly_ the same schedule, they don't _exactly_ match.
Let me find out if this is good enough for now or if we have something else to use.

badboy (changes requested):

etl/glean_etl.py
badboy: at least for android we would have https://product-details.mozilla.org/1.0/mobile_android.json
releases are usually off by a day for reasons™️.
It's a bit more work to parse that, so I'd be ok to just go with the desktop version for now.

We don't have data for iOS anyway, but releases should be similar, so it's ok as well. 
Let's file a followup to gather better data for this.

badboy (approved):

etl/expiry.py
Iinh: yep. Will file an issue to keep track of this.

PR #1517, opened 2022-11-17
Show expiration information with metric search

Iinh:
Recently, in our internal Slack, some users asked to show the expiration information in the search view, rather than having it only on the MetricDetails page.

See the [deploy preview](https://deploy-preview-1517--glean-dictionary-dev.netlify.app/apps/fenix) for the UI change.

CleanShot 2022-11-17 at 10 04 15@2x


Iinh:
@badboy hey Jan-Erik would you be able to review this PR? Thanks in advance!

badboy (approved):
Looks simple enough.
Bit weird when there are both versions and date expiry visible, but I don't have a good idea how to visually separate those.

PR #1518, opened 2022-11-17
Enable filtering by expiration information

### LINKED ISSUE #1467: Enable filtering by expiry date
I would love a way to filter by expiry date and to be able to choose a range.

Every 6 months I do an audit of our telemetry and renew metrics that are expiring soon. Currently I have to search through the metrics.yml.

It would be great to have a filter on the glean dictionary to see, for example, all metrics expiring in the next 3 months, or next 6 months, etc.

Iinh:
Let users search for metrics that will expire n number of months from now, by querying e.g. `expires:3` i.e. metrics that will expire 3 months/versions from now. 

Fixes #1467.

### DIFF
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -277,7 +277,9 @@
                 </td>
                 <td style="text-align: center;">
                   <div class="item-property">
-                    {item.expires ? item.expires : "never"}
+                    <Markdown
+                      text={item.expires ? item.expiry_text : "never"}
+                    />
                   </div>
                 </td>
               {:else if itemType === "tags"}
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -144,7 +144,7 @@
   {#if showFilter}
     <FilterInput
       tooltipped={true}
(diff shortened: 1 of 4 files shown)


badboy:
I don't think this is working as expected right now.
No matter what value I add, e.g. https://deploy-preview-1518--glean-dictionary-dev.netlify.app/apps/fenix?page=1&search=expires%31, I see the `never` expiring metrics.

With version expires it can also get a bit confusing, but I'm not sure how to solve this nicely. Due to our nice release cycle this might not be too bad though.

E.g. https://deploy-preview-1518--glean-dictionary-dev.netlify.app/apps/fenix?page=1&search=expires%3A6 shows nothing (except all the `never`-ending metrics),
https://deploy-preview-1518--glean-dictionary-dev.netlify.app/apps/fenix?page=1&search=expires%3A7 shows some expiring in 118.
But this lacks context, as one would need to know what version we have right now (112).

So I think the check for that should be inclusive? (I would have expected it to show the 118-metrics for `expires:6`)

badboy (approved):

Issue #1516, opened 2022-11-16
Tippy (tooltip library) doesn't work in production


Iinh:
It seems that Tippy no longer works on dictionary.telemetry.mozilla.org

<img width="440" alt="CleanShot 2022-11-16 at 13 23 20@2x" src="https://user-images.githubusercontent.com/28797553/202262095-6f0ee705-847a-4d14-a83e-58dd16b6b185.png">

<img width="634" alt="CleanShot 2022-11-16 at 13 23 12@2x" src="https://user-images.githubusercontent.com/28797553/202262112-a3e401fa-c7a2-45a0-be98-8946962f099e.png">

Locally on my machine it still works as expected.

<img width="500" alt="CleanShot 2022-11-16 at 13 21 38@2x" src="https://user-images.githubusercontent.com/28797553/202261796-712c4154-8564-478b-b529-8f2351223ee0.png">
<img width="810" alt="CleanShot 2022-11-16 at 13 21 31@2x" src="https://user-images.githubusercontent.com/28797553/202261804-130e6e6e-3d16-4d0a-b282-0948b744449f.png">

wlach:
Noticed this myself. This one's weird. Looking at the generated CSS it seems like it's not processing/including the tippy css that's being included in global styles: 
![image](https://user-images.githubusercontent.com/20569/202922326-5aba0d28-e4ea-4d9b-bae6-0174f11cbb3a.png)

https://github.com/mozilla/glean-dictionary/blob/c8c2a42fba76225827c08d816322abe7ff783c6f/src/GlobalStyles.svelte#L11

Weirdly, it looks like it has no problem with the other includes in the same file. I can't reproduce this problem at all locally. Perhaps the full log of the netlify deploy (which I can't see myself) might provide a hint.

Dexterp37:
I stumbled upon this myself as well :\ This is fairly annoying :-D

PR #1520, opened 2022-11-20
Update Glean Dictionary's lockfile to the latest version

wlach:
I suspect this might fix some tooltip styling issues I'm seeing in
production, let's see!


wlach:
I thought this might fix #1516, since it's the only difference between what's on my machine and what's being used in production. But I see the same problem here. Closing.

PR #1598, opened 2023-02-06
Add a copy button to the metric detail page (closes #1515)

### LINKED ISSUE #1515: Feature Request: An easy-to-copy column value for each metric
I often use the Glean Dictionary while writing queries on Glean tables. One small hindrance is that the column is down below:
Screen Shot 2022-11-16 at 10 26 23 AM

So not only do I have to scroll down, it's hard to copy+paste that value because it's a link.

Could we add a field up above in `Metadata` that has the column value not as a link? In this case maybe `Column: metrics.string.metrics_install_source`

wlach:
Probably could look a little nicer, but it works!

### DIFF
--- a/src/components/CopyButton.svelte
+++ b/src/components/CopyButton.svelte
@@ -1,0 +1,19 @@
+<script>
+  import tippy from "./tippy";
+  import CopyIcon from "./icons/CopyIcon.svelte";
+
+  export let textToCopy;
+
+  function updateClipboard() {
+    navigator.clipboard.writeText(textToCopy);
+  }
+</script>
+
+<button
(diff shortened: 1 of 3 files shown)


wlach:
image

wlach:
No idea about the circleci issue, but I don't think it has anything to do with the PR.

badboy:
The deploy key was cycled. CI should now work again.

badboy (approved):

badboy:
```
W: Some index files failed to download. They have been ignored, or old ones used instead.
```

-_-
Different error. CI is starting correctly. 

By the power of my admin capabilities I'll still land this now.

PR #1681, opened 2023-04-24
Provide link to start an STMO query for metrics

### LINKED ISSUE #1644: Make Glean dictionary provide stub SQL as well as allowing to copy-paste names
It would be great for the Glean dictionary to make it super easy to start an STMO (or BQ) query for searched fields.  It's possible to inject SQL into STMO (Redash) using an `{{ input }}`, but perhaps there's a better way with an API call.  I'm thinking of something like:
```sql
SELECT
    m.metrics.datetime.background_update_targeting_env_profile_age
FROM
    firefox_desktop_background_update.background_update AS m
WHERE
    DATE(submission_timestamp) = CURRENT_DATE()
```

Iinh:
Following the examples from this [guide](https://docs.telemetry.mozilla.org/cookbooks/accessing_glean_data.html), I created 2 starter queries: one for event metrics and one for the other metric types. Then link to them from the Glean Dictionary.

UI change:

CleanShot 2023-05-01 at 15 28 58@2x

Fixes #1644

### DIFF
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -42,6 +42,19 @@
   let selectedPingVariant;
   let pingData = {};
 
+  const STMO_GLEAN_QUERY_URL =
+    "https://sql.telemetry.mozilla.org/queries/91729/source";
+  const STMO_GLEAN_EVENT_QUERY_URL =
+    "https://sql.telemetry.mozilla.org/queries/91780/source";
+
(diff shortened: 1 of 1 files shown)


badboy (approved):
Nice.
Can you add a comment to the queries like:

> Fork this query to modify and save it.

or something along those lines, so people who are new to this know what to do.

PR #1694, opened 2023-05-01
Work around uBlock Origin privacy filters

### LINKED ISSUE #1682: uBlock Origin blocks parts of the Glean Dictionary website
The text of "ad_impression" seems to trigger uBlock to prevent the following page from loading. https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/serp_ad_impression

Iinh:
Fixes #1682.

See deploy preview to verify [serp_ad_impression](https://deploy-preview-1694--glean-dictionary-dev.netlify.app/apps/firefox_desktop/metrics/serp_ad_impression) is no longer blocked by uBlock Origin.

### DIFF
--- a/src/state/api.js
+++ b/src/state/api.js
@@ -22,12 +22,10 @@ export async function getMetricData(appName, metricName) {
   let updatedMetricName = metricName;
 
   Object.keys(UBLOCK_ORIGIN_PRIVACY_FILTER).forEach((filter) => {
-    if (updatedMetricName.includes(filter)) {
-      updatedMetricName = updatedMetricName.replace(
-        filter,
-        UBLOCK_ORIGIN_PRIVACY_FILTER[filter]
-      );
-    }
+    updatedMetricName = updatedMetricName.replace(
+      filter,
+      UBLOCK_ORIGIN_PRIVACY_FILTER[filter]
+    );
   });
 
   // we added data to metric names to avoid the JSON resource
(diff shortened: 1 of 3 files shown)


src/state/api.js
```diff
@@ -18,9 +19,20 @@ export async function getPingData(appName, pingName) {
 }
 
 export async function getMetricData(appName, metricName) {
+  let updatedMetricName = metricName;
+
+  Object.keys(UBLOCK_ORIGIN_PRIVACY_FILTER).forEach((filter) => {
+    if (updatedMetricName.includes(filter)) {
```
mikewilli, line 25: This check should not be needed -- if `updatedMetricName` does not contain `filter` then the replace will just return the original string.

src/state/api.js
```diff
@@ -18,9 +19,20 @@ export async function getPingData(appName, pingName) {
...
+      updatedMetricName = updatedMetricName.replace(
```
mikewilli, line 26: One note since I looked it up: JS `replace` will only replace the first occurrence, whereas the Python `replace` used in `etl/glean_etl.py` will replace all occurrences by default. I don't think it's worth worrying about (seems pretty unlikely that there will be a metric with the same blocked string twice), but just so you're aware on the off chance it does come up in the future. Possible solutions would be to use regex (with /g) or nested loops to ensure you get all the instances of the filter key.

mikewilli (approved):
Looks good to me!

PR #1706, opened 2023-05-12
Allow copying generated SQL code

Dexterp37:
In #1644 we added a way for folks to create a dummy query in STMO. That's indeed a good start, although it has a couple of shortcomings:

- the link to STMO takes to a query that uses variables and how to use it it's not immediately clear;
- the usage of `CURRENT_DATE` makes the query return no result (as stable tables might not be populated for the current day)
- if user wants to query in the BigQuery console, he has to follow the STMO link and copy query, the paste it in BQ and customize it because the fields are using STMO variables;

This PR adds a new button to the "Access" section, for both metrics and pings, to copy pre-generated SQL code that could be easily pasted and used both in STMO and BQ.

### Pull Request checklist

      fixes, if applicable)


src/pages/MetricDetail.svelte
```diff
@@ -327,16 +374,19 @@
               >
                 {selectedAppVariant.etl.bigquery_column_name}
               </a>
-              <CopyButton
-                textToCopy={selectedAppVariant.etl.bigquery_column_name}
-              />
```
wlach, line 332: A bit sad to lose this as I think this was added for another use case.

wlach (commented):
It's a bit confusing to me what the copy button beside the BigQuery table actually means:

image

I'd personally add another row to the access table with its own explanation, maybe something like this (in retrospect I wouldn't capitalize "copy sql" but you get what I mean):

image

src/pages/MetricDetail.svelte
Dexterp37: @wlach nice to see you! 👋  Do you recall which use-case?

wlach: Yup! See https://github.com/mozilla/glean-dictionary/issues/1515 and https://github.com/mozilla/glean-dictionary/pull/1598

wlach:
> > It's a bit confusing to me what the copy button beside the BigQuery table actually means:
> 
> Heh, yeah. I had an async conversation with @Iinh where I mentioned that my front-end skills are negligible and this is more to get the conversation started :-)

Makes sense. 😄 

> > I'd personally add another row to the access table with its own explanation, maybe something like this (in retrospect I wouldn't capitalize "copy sql" but you get what I mean):
> 
> Fair. Let's see what @Iinh thinks! In the end I have no real preference, but I'd really appreciate having some way to copy pre-baked sql

Agreed, this seems like a nice feature to have.

src/pages/MetricDetail.svelte
Iinh: I agree that we shouldn't change the functionality of the copy button, considering it was added originally for another use case that is still needed as wlach pointed out.

Iinh:
> I'd personally add another row to the access table with its own explanation, maybe something like this (in retrospect I wouldn't capitalize "copy sql" but you get what I mean):
> 
> > Fair. Let's see what @Iinh thinks! In the end I have no real preference, but I'd really appreciate having some way to copy pre-baked SQL


CleanShot 2023-05-16 at 08 01 54@2x

UX-wise, I'd prefer the UI to make it clearer to users -- before they leave the site -- what's being copied here and where to use it: pre-generated SQL that can be used in STMO and BigQuery. What I have in mind is similar to GLAM's approach, where the SQL content is in a pop-up component where it can still be quickly copied, along with more details on where and how to use it.

Since that requires more of a front-end lift, and if that approach seems reasonable, I'd be happy to pick up the work from here if you don't mind @Dexterp37 :)

Dexterp37:
> > I'd personally add another row to the access table with its own explanation, maybe something like this (in retrospect I wouldn't capitalize "copy sql" but you get what I mean):
> > > Fair. Let's see what @Iinh thinks! In the end I have no real preference, but I'd really appreciate having some way to copy pre-baked SQL
> 
> CleanShot 2023-05-16 at 08 01 54@2x
> 
> UX-wise, I'd prefer the UI to make it clearer to users -- before they leave the site -- what's being copied here and where to use it: pre-generated SQL that can be used in STMO and BigQuery. What I have in mind is similar to GLAM's approach, where the SQL content is in a pop-up component where it can still be quickly copied, along with more details on where and how to use it.
> 
> Since that requires more of a front-end lift, and if that approach seems reasonable, I'd be happy to pick up the work from here if you don't mind @Dexterp37 :)

I don't mind at all! I'm sorry that I have pushed more work you're way, though! :-E

Iinh:
See #1717 for the continuation of this work.

PR #1725, opened 2023-05-22
Expose all the UNION-ed tables

### LINKED ISSUE #1479: Report the 'union' tables for products that have them
Some products (e.g. [Focus Android](https://github.com/mozilla/probe-scraper/blob/926f0af1fa2a5e145fc0d7221905f72bba549279/repositories.yaml#L670-L685)) have multiple channels, with one app id per channel (that's the case mostly for mobile products). These products generally have a dataset that essentially UNIONs the data of all the channels.

For example, Focus Android has the `focus_android`  dataset that is the union of `org_mozilla_focus`, `org_mozilla_focus_beta` and `org_mozilla_focus_nightly`. The availability of such a UNION'd dataset is reported by the probe-scraper. It would be nice to mention that where we link to the tables.

Iinh:
Fixes #1479. If there's more than one app_id available, there's always a UNION table, so essentially all Glean apps have it. 

Since Data Catalog is available now, we should link to this table in Data Catalog.

CleanShot 2023-05-23 at 10 40 09@2x

### DIFF
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -320,6 +320,15 @@
                 selectedPingVariant.id
               )}>{pingData.bigquery_table}</a
             >
+            and <AuthenticatedLink
+              href={`https://mozilla.acryl.io/dataset/urn:li:dataset:(urn:li:dataPlatform:bigquery,moz-fx-data-shared-prod.${params.app}.${selectedPingVariant.id},PROD)`}
+            >
+              {params.app}.{selectedPingVariant.id}
+              <HelpHoverable
+                content={"The result table of UNION-ing all the per-app_id datasets."}
+                link={"https://github.com/mozilla/bigquery-etl/tree/main/sql_generators/glean_usage#glean-usage"}
+              />
+            </AuthenticatedLink>
             <!-- Skip search string for event metrics as we can't directly lookup the columns in events tables -->
             {#if metric.type !== "event"}
               as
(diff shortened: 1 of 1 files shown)


scholtzan (approved):

src/pages/MetricDetail.svelte
```diff
@@ -343,6 +343,19 @@
               >
               <CopyButton textToCopy={metric.event_info.name} />)
             {/if}
+            {#if metric.variants.length}
+              <p>
+                (also in <AuthenticatedLink
+                  href={`https://mozilla.acryl.io/dataset/urn:li:dataset:(urn:li:dataPlatform:bigquery,moz-fx-data-shared-prod.${params.app}.${selectedPingVariant.id},PROD)`}
```
Dexterp37, line 349: @Iinh should this actually be more explicit and have its own entry as "Data Catalogue" (e.g. below Big Query)? If you think that doesn't help, what about listing the tables as bullets instead of having this long text?

For example

```markdown
* `org_mozilla_ios_firefox.events` (bla bla bla)
* `firefox_ios.events` (bla bla bla)
```

My concern is that the discovery of this information is hindered by the density :)

Iinh: Since this is for the Metric page, and the Data Catalog link is of a ping dataset, I don't think it'd make sense to have its own separate entry yet (we're reserving that spot for when Glean metrics become available in Data Catalog 🙂)

f you think that doesn't help, what about listing the tables as bullets instead of having this long text?

Dexterp37: > > If you think that doesn't help, what about listing the tables as bullets instead of having this long text?
> 
> Makes sense! I'll change it :)

Thanks!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Iinh: I simplified it to just one line, instead of using bullet points:

CleanShot 2023-05-23 at 21 55 04@2x

What do you think?

Dexterp37 (approved):

PR #1881, opened 2023-08-31
Link to metrics in Data Catalog

Iinh:
Link to metrics now that column-level lineage is available in Data Catalog.

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/state/urls.js
+++ b/src/state/urls.js
@@ -23,3 +23,11 @@ export function getBigQueryURL(appName, appId, pingName, metricName) {
 export function getMetricSearchURL(app, search) {
   return `/apps/${app}?search=${search}`;
 }
+
+export function getDataCatalogMetricURL(appId, pingId, bigQueryColumn) {
+  const DATA_CATALOG_URL = "https://mozilla.acryl.io";
+  return `${DATA_CATALOG_URL}/dataset/urn:li:dataset:(urn:li:dataPlatform:bigquery,moz-fx-data-shared-prod.${appId.replace(
+    /\./g,
+    "_"
+  )}.${pingId},PROD)/Lineage?column=${bigQueryColumn}`;
+}
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -32,7 +32,11 @@
     updateURLState,
     updateBreadcrumbs,
   } from "../state/stores";
-  import { getBigQueryURL, getMetricSearchURL } from "../state/urls";
+  import {
+    getBigQueryURL,
+    getDataCatalogMetricURL,
+    getMetricSearchURL,
+  } from "../state/urls";
   import { getAppBreadcrumbs } from "./AppDetail.svelte";
 
   import { isExpired, isRemoved, isRecent } from "../state/items";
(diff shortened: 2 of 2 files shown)


Dexterp37:
@Iinh is this something we still want? Do you want me to help poking people?

Iinh:
> @Iinh is this something we still want? Do you want me to help poking people?

Yes we still want this, but tabling this work for now due to a UI issue in Data Catalog that prevents metric column from being shown. Will pick it back up once that is fixed.

Dexterp37:
> > @Iinh is this something we still want? Do you want me to help poking people?
> 
> Yes we still want this, but tabling this work for now due to a UI issue in Data Catalog that prevents metric column from being shown. Will pick it back up once that is fixed.

Hey @Iinh , was the blocker fixed? If not, would you please block this against the ticket used to track the other problem?

Iinh:
thanks @Dexterp37! 

@ANich could you review this and see if it's good to go - thank you!

src/state/urls.js
```diff
@@ -23,3 +23,11 @@ export function getBigQueryURL(appName, appId, pingName, metricName) {
 export function getMetricSearchURL(app, search) {
   return `/apps/${app}?search=${search}`;
 }
+
+export function getDataCatalogMetricURL(appId, pingId, bigQueryColumn) {
+  const DATA_CATALOG_URL = "https://mozilla.acryl.io";
+  return `${DATA_CATALOG_URL}/dataset/urn:li:dataset:(urn:li:dataPlatform:bigquery,moz-fx-data-shared-prod.${appId.replace(
+    /\./g,
+    "_"
+  )}.${pingId},PROD)/Lineage?column=${bigQueryColumn}`;
```
ANich, line 32: It might be more useful to point to the `stable` or `live` version of this BigQuery table corresponding to the latest version since I believe that's what's mostly used in ETL downstream though I don't think that's consistent either. 

Also it looks like column lineage still doesn't support nested fields which these mostly are so the lineage view is empty when I click on it in the preview deploy. Maybe pointing to the top level column would be more useful? i.e. `metrics` instead of `metrics.x.y`

ANich (approved):
Left a few suggestions/thoughts. In the preview I'm not really seeing much though I only looked at a few examples

src/state/urls.js
Iinh: > i.e. metrics instead of metrics.x.y

For clarification do you mean to link to the ping level instead?

ANich: I mean the nested BQ columns but actually it might be fine. Here's an [example](https://mozilla.acryl.io/dataset/urn:li:dataset:(urn:li:dataPlatform:bigquery,moz-fx-data-shared-prod.firefox_desktop_stable.newtab_v1,PROD)/Lineage?column=%5Bversion%3D2.0%5D.%5Btype%3Dstruct%5D.%5Btype%3Dstruct%5D.metrics.%5Btype%3Dstruct%5D.uuid.%5Btype%3Dstring%5D.legacy_telemetry_client_id&filter_degree___false___EQUAL___0=1&page=1&unionType=0) with downstream column lineage that's okay. The first paragraph might still be relevant though (looking at `_stable`).

Iinh: Thanks for the clarification. 

> I mean the nested BQ columns but actually it might be fine. Here's an example with downstream column lineage that's okay.

Yea I saw some that worked and some that didn't show anything. Will leave it as is for now.

> The first paragraph might still be relevant though (looking at _stable).

I added:
- `_stable` and `_v1` with the assumption that's the latest version for most tables. Would this work?

ANich: > `_stable` and `_v1` with the assumption that's the latest version for most tables. Would this work?
LGTM 👍 would be nice if we could use some metadata to determine the version, not sure how much work that is  here though

Dexterp37:
Thank you all!

PR #1945, opened 2023-10-17
Enable showing only app metrics

### LINKED ISSUE #1944: Add an option to only show application metrics
Metrics added in the application are often buried under the set of metrics the application inherits from its dependencies, including Glean. It would be nice to have a way to only show metrics from the application itself.

immagine

Dexterp37:
This fixes #1944 .

It additionally breaks a circular dependency  detected at build time.

By ticking the box the Glean Dictionary entry shows this instead of two pages of less interesting metrics:

immagine

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -21,7 +21,7 @@
     getLibraryDescription,
   } from "../data/help";
 
-  import { adjustDataTypes } from "./iOSThirdPartyData.svelte";
+  import { adjustDataTypes } from "./AdjustDataTypes.svelte";
 
   import { fullTextSearch } from "../state/search";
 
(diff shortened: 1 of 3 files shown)


src/components/ItemList.svelte
```diff
@@ -155,7 +165,7 @@
       </h3>
       {#if totalItems > 0}
         <p>
-          {totalItems} expired or removed {itemType} found.
+          {totalItems} expired, removed or hidden {itemType} found.
```
Iinh, line 168: let's also add this to line 318 of this file

Iinh (approved):
LGTM, thanks for working on this!

src/components/ItemList.svelte
Dexterp37: Thanks, done!

PR #1946, opened 2023-10-18
Update some of the svelte, storybook dependencies

Dexterp37:
Note: this is based on #1945 and should be rebased if it gets merged.

This should fix some of the dependabot PRs that are currently failing, as some of the dependency had major breaking changes that I'm tackling here. This is not yet ready for review (hence the draft mode!) because I still need to:

- [ ] Understand why clicking on "Docs" in the storybook triggers a `Uncaught (in promise) TypeError: docsParameter.renderer is not a function`
- [x] Migrate over any other dependency marked as deprecated

### Pull Request checklist

      fixes, if applicable)


.storybook/main.js
```diff
@@ -4,7 +4,7 @@ const path = require("path");
 
 module.exports = {
   stories: ["../stories/**/*.stories.js"],
-  addons: ["@storybook/addon-knobs"],
+  addons: ["@storybook/addon-controls"],
```
Iinh, line 7: CleanShot 2023-10-26 at 13 30 05@2x

Looking at the error logs, seems like we also need to install the `"@storybook/addon-docs"` dependency and add it here for the autogenerated docs to work.

```suggestion
  addons: ["@storybook/addon-controls", "@storybook/addon-docs"],
```

package.json
```diff
@@ -30,10 +30,12 @@
...
-    "@storybook/addon-knobs": "6.4.0",
-    "@storybook/addon-storyshots": "6.5.10",
-    "@storybook/addon-storysource": "6.5.10",
-    "@storybook/svelte": "6.5.10",
+    "@storybook/addon-controls": "^7.5.0",
+    "@storybook/addon-storyshots": "7.4.6",
+    "@storybook/addon-storysource": "7.4.6",
+    "@storybook/cli": "7.4.6",
+    "@storybook/svelte": "7.4.6",
+    "@storybook/svelte-webpack5": "^7.4.6",
```
Iinh, line 38: I strongly suggest we keep all storybook dependencies at the same version number, maybe 7.4.6, due to experiences in the past where CI broke because they were at different versions.

Iinh:
```suggestion
    "@storybook/addon-controls": "7.4.6",
     "@storybook/addon-docs": "7.4.6",
    "@storybook/addon-storyshots": "7.4.6",
    "@storybook/addon-storysource": "7.4.6",
    "@storybook/cli": "7.4.6",
    "@storybook/svelte": "7.4.6",
    "@storybook/svelte-webpack5": "7.4.6",
```

Iinh (approved):
> Understand why clicking on "Docs" in the storybook triggers a Uncaught (in promise) TypeError: docsParameter.renderer is not a function

Left some comments below that could fix this error. I'd also be okay with setting `autodocs: false` in `main.js` for now since we don't utilize docs functionality much.

r+ with suggested changes. Thank you so much for working on this!

PR #1954, opened 2023-10-25
Update Looker url to event_counts to show event counts over time

scholtzan:
We'd like to show event counts over time when clicking on the Looker link for a specific event. Currently, it's just showing the overall counts which aren't really useful to users.

This PR updates the relevant Looker URLs.


quiiver (approved):
i have no idea what i'm looking at but it looks great

Iinh (approved):
Thanks, this is definitely more helpful than the current event links!

PR #1975, opened 2023-11-03
Add link to Looker event monitoring dashboard

scholtzan:
This adds a link to the [event monitoring dashboard](https://mozilla.cloud.looker.com/dashboards/1452?Event+Name=disqualification&App+Name=Firefox+for+iOS&Window+Start+Time=28+days&Channel=release) for Glean events.

cc @Dexterp37

### DIFF
--- a/src/pages/MetricDetail.svelte
+++ b/src/pages/MetricDetail.svelte
@@ -381,6 +381,28 @@
           />
         </td>
       </tr>
+      {#if pingData.event_monitoring}
+        <tr>
+          <td
+            >Monitoring <HelpHoverable
+              content={"Show real-time monitoring for this event in Mozilla's instance of Looker."}
+            />
+          </td>
+          <td>
+            <div>
+              In the
+              <AuthenticatedLink href={pingData.event_monitoring.base}>
+                Event Monitoring Dashboard
+              </AuthenticatedLink>
+              for the
+              <AuthenticatedLink href={pingData.event_monitoring.event.url}>
+                {pingData.event_monitoring.event.name}
+              </AuthenticatedLink>
+              event
+            </div>
+          </td>
+        </tr>
+      {/if}
     </table>
   {/if}
 {:catch}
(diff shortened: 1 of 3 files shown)


Dexterp37:
This is what this PR adds:

immagine

@scholtzan I tried this in the preview and, for the `page.load` event of the Glean Debug Ping Viewer it defaults to a non-existing channel for that product (`beta`). Any way to not have the dashboard default to a channel unless specified by the dictionary? The URL itself seems to not have anything about beta at all, `https://mozilla.cloud.looker.com/dashboards/1452?App+Name=Glean+Debug+Ping+Viewer&Event+Name=load`

src/pages/MetricDetail.svelte
```diff
@@ -381,6 +381,28 @@
...
+            >Monitoring <HelpHoverable
+              content={"Show real-time monitoring for this event in Mozilla's instance of Looker."}
+            />
+          </td>
+          <td>
+            <div>
+              In the
+              <AuthenticatedLink href={pingData.event_monitoring.base}>
+                Event Monitoring Dashboard
+              </AuthenticatedLink>
```
Dexterp37, line 396: I find linkifying this a bit confusing: we likely only want to showcase the direct link. @Iinh , thoughts?

Iinh: taking a closer look- I'm wondering if it might be a good idea to include this in the Looker row since it's also connected to a Looker dashboard. But there might be a specific reason we'd want to highlight this further by keeping it in a separate row that I'm not aware of? Wdyt?

Dexterp37: Good question. @scholtzan ?

scholtzan: Yeah, I can move it into that row

scholtzan:
I updated the dashboard so that the default Channel is set to "any channel" when not specified

Dexterp37 (approved):

PR #2001, opened 2023-11-09
Bug 1842150 - Add server knobs observability

travis79:
This adds information about the sampling of metrics via Server Knobs

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/etl/glean.py
+++ b/etl/glean.py
@@ -47,6 +47,7 @@ class GleanObject(object):
     ORIGIN_KEY = "origin"
     HISTORY_KEY = "history"
     IN_SOURCE_KEY = "in-source"
+    SAMPLING_INFO_KEY = "sampling_info"
 
 
 class GleanMetric(GleanObject):
@@ -91,6 +92,7 @@ class GleanMetric(GleanObject):
         self.definition["name"] = full_defn[self.NAME_KEY]
         self.definition["origin"] = full_defn[self.ORIGIN_KEY]
         self.definition["in_source"] = full_defn[self.IN_SOURCE_KEY]
+        self.definition["sampling_info"] = full_defn.get(self.SAMPLING_INFO_KEY)
 
         # first seen is the earliest date in the history
         self.definition["date_first_seen"] = self.definition_history[-1]["dates"]["first"]
--- a/etl/glean_etl.py
+++ b/etl/glean_etl.py
@@ -137,9 +137,8 @@ def _get_app_variant_description(app):
     return description
 
 
-def _get_metric_sample_data() -> dict:
+def _get_metric_sample_data(experiment_data) -> dict:
     # get experiment metric sampling data to enrich metric definitions
-    experiment_data = requests.get(EXPERIMENT_DATA_URL).json()
     interesting_experiments = [
         experiment for experiment in experiment_data if "glean" in experiment["featureIds"]
     ]
@@ -155,7 +154,7 @@ def _get_metric_sample_data() -> dict:
         bucket_config = experiment["bucketConfig"]
         sample_size = bucket_config["count"] / bucket_config["total"]
         channel = experiment["channel"]
-        sampling_data[app_name] = {}
+        sampling_data[app_name] = sampling_data.get(app_name, {})
         for branch in experiment["branches"]:
             feature_configs = branch["features"]
             filtered_configs = [
@@ -168,8 +167,10 @@ def _get_metric_sample_data() -> dict:
             ]
             for entry in metric_config:
                 for key in entry:
-                    sampling_data[app_name][key] = {}
-                    sampling_data[app_name][key][channel] = {}
+                    sampling_data[app_name][key] = sampling_data[app_name].get(key, {})
+                    sampling_data[app_name][key][channel] = sampling_data[app_name][key].get(
+                        channel, {}
+                    )
                     sampling_data[app_name][key][channel]["sample_size"] = sample_size
                     sampling_data[app_name][key][channel]["experiment_id"] = experiment["slug"]
                     sampling_data[app_name][key][channel]["start_date"] = experiment["startDate"]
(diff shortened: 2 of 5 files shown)


travis79:
Screenshot 2023-11-09 at 10 08 44 AM

Dexterp37:
@travis79 this is a good start. A few observations:

* the ETL is failing because of something being `None`, let's fix it so that we can see the live preview here.
* it might be wise to add tests for the ETL (see the other ETL tests).
* it's a bit unclear to me why we're mentioning "experiments": the fact that we're using experiments (rollouts?) is an implementation detail. What is that you're trying to present?

travis79:
>     * the ETL is failing because of something being `None`, let's fix it so that we can see the live preview here.

Ahh, dang. I thought I had found all of the places that this could happen. I did see some of this in my local testing but I must have missed some.
 
>     * it might be wise to add tests for the ETL (see the other ETL tests).

Agreed, and this is why this is still a draft as I definitely want test coverage, especially considering the `None` failure in CI.

>     * it's a bit unclear to me why we're mentioning "experiments": the fact that we're using experiments (rollouts?) is an implementation detail. What is that you're trying to present?

This was something that DS folks liked, mostly because it helps to document the source of the sampling. I'm happy to try and downplay the "experiment/rollout" terminology here in the Glean Dictionary, but knowing where to go to change a sampling configuration does seem useful, but I don't have any strong opinion about where this should live.

Dexterp37:
> > ```
> > * it's a bit unclear to me why we're mentioning "experiments": the fact that we're using experiments (rollouts?) is an implementation detail. What is that you're trying to present?
> > ```
> 
> This was something that DS folks liked, mostly because it helps to document the source of the sampling.

@travis79 what other alternatives were offered to DS?

> I'm happy to try and downplay the "experiment/rollout" terminology here in the Glean Dictionary, but knowing where to go to change a sampling configuration does seem useful, but I don't have any strong opinion about where this should live.

It's not about downplaying, it's more about setting the right context imho: how does the word "experiment" in this context interact with the notion of "experiments" we're going to show in other dashboards? Will this show a subset? The same set? If the latter, why are we showing experiments that have nothing to do with sampling? If the former, this could cause confusion as DS are not the only users of this and the fact that we're using "experiments" is an implementation detail.

This is one potential path (happy to discuss other options):

1. Replace the word "Experiment" with "Sampling"
2. Add an information box to "Sampling id" mentioning what this means (you would need this even if you stick with the current nomenclature I believe)
3. Linkify the experimenter link. Rename it to "Sampling monitoring dashboard" or something like that.
4. Instead of no value for "End Date", can we add something like "Currently active"?

Note that I think @quiiver needs to be a reviewer on this.

travis79:
@Dexterp37 Thanks for reviewing this while you are busy with your focused workweek. Please see my comments below, I have no complaints with your suggestions and will push a commit today with updates as requested.

> It's not about downplaying, it's more about setting the right context imho: how does the word "experiment" in this context interact with the notion of "experiments" we're going to show in other dashboards? Will this show a subset? The same set? If the latter, why are we showing experiments that have nothing to do with sampling? If the former, this could cause confusion as DS are not the only users of this and the fact that we're using "experiments" is an implementation detail.

Okay, fair enough. I don't want to create confusion and it's probably not clear to everyone that experiments are what powers this sampling anyway.

> This is one potential path (happy to discuss other options):
> 
>     1. Replace the word "Experiment" with "Sampling"

I'm good with this replacement

>     2. Add an information box to "Sampling id" mentioning what this means (you would need this even if you stick with the current nomenclature I believe)

What about a more descriptive label for the field? Something like "Sampling source id", and I could also see if I can add a tooltip or something for the on-hover that expands on that with: "The identifier of the sampling configuration for this metric"?

>     3. Linkify the experimenter link. Rename it to "Sampling monitoring dashboard" or something like that.

I already did this yesterday... perhaps you were looking at an earlier version of this from a preview link that I shared before I added the `<a href=>`?

>     4. Instead of no value for "End Date", can we add something like "Currently active"?

Yes, I was already thinking about doing something like this to fill in the blank field with something descriptive, "Active" or "Currently active" works just fine.

> Note that I think @quiiver needs to be a reviewer on this.

Yep, he's in a workweak I hear and I didn't want to interfere with that, but if you think he needs to review this before it lands on my proposed completion date of the 16th, then I'll go ahead and ping him. I'd like to get Daniel Berry to take a look before this lands and I've already had Charlie from the Nimbus team look at this also. Is there anyone else you can think of that should see this before it lands?

Dexterp37:
> > ```
> > 2. Add an information box to "Sampling id" mentioning what this means (you would need this even if you stick with the current nomenclature I believe)
> > ```
> 
> What about a more descriptive label for the field? Something like "Sampling source id", and I could also see if I can add a tooltip or something for the on-hover that expands on that with: "The identifier of the sampling configuration for this metric"?

Sounds good!

> > ```
> > 3. Linkify the experimenter link. Rename it to "Sampling monitoring dashboard" or something like that.
> > ```
> 
> I already did this yesterday... perhaps you were looking at an earlier version of this from a preview link that I shared before I added the `<a href=>`?
Whoops, sorry, yes, looking at the screenshot.
> > ```
> > 4. Instead of no value for "End Date", can we add something like "Currently active"?
> > ```
> 
> Yes, I was already thinking about doing something like this to fill in the blank field with something descriptive, "Active" or "Currently active" works just fine.

👍 

> > Note that I think @quiiver needs to be a reviewer on this.
> 
> Yep, he's in a workweak I hear and I didn't want to interfere with that, but if you think he needs to review this before it lands on my proposed completion date of the 16th, then I'll go ahead and ping him. I'd like to get Daniel Berry to take a look before this lands and I've already had Charlie from the Nimbus team look at this also. Is there anyone else you can think of that should see this before it lands?

Oh, yeah, let's not block on Wil. Let's merge and reference this to him so that he can review the landed thing.

rosahbruno (approved):
Besides a few small UI callouts, I think everything looks great!

All the text in the table is starting to look a little cramped. I had a few ideas of how we could maybe space things out a bit.

1. We could increase the height of each row in the tables from `50px` to `75px`. It doesn't really even look like that much of a difference but mostly everything is now visible without any scrolling or looking too cramped.
2. Make the font-size of all the table cells (besides the metric name) 14px instead of most of them being 16 and the description being 14. I think having these all be 14px might make things look a bit more uniform too.

Dexterp37 (approved):

Dexterp37:
> Besides a few small UI callouts, I think everything looks great!
> 
> All the text in the table is starting to look a little cramped. I had a few ideas of how we could maybe space things out a bit.
> 
>     1. We could increase the height of each row in the tables from `50px` to `75px`. It doesn't really even look like that much of a difference but mostly everything is now visible without any scrolling or looking too cramped.
> 
>     2. Make the font-size of all the table cells (besides the metric name) 14px instead of most of them being 16 and the description being 14. I think having these all be 14px might make things look a bit more uniform too.

I went ahead and merged this: @travis79 let's take the above to a follow-up PR.

PR #2078, opened 2024-01-12
Fix quotes in Events dashboard links

scholtzan:
The event names need to be wrapped in quotes for the Looker filter to consider it as string and not raw value.

### DIFF
--- a/etl/looker.py
+++ b/etl/looker.py
@@ -273,10 +273,11 @@ def get_looker_monitoring_metadata_for_event(app, app_group, metric, ping_name):
     if metric_type != "event":
         return None
 
-    (_, metric_name) = get_event_name_and_category(metric.identifier)
+    (metric_category, metric_name) = get_event_name_and_category(metric.identifier)
+    event_identifier = ".".join([metric_category, metric_name])
 
     url = furl(EVENT_MONITORING_DASHBOARD_URL).add(
-        {"App Name": app.app["canonical_app_name"], "Event Name": '"' + metric_name + '"'}
+        {"App Name": app.app["canonical_app_name"], "Event Name": '"' + event_identifier + '"'}
     )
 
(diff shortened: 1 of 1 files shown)


Dexterp37:
@scholtzan would you kindly provide an example of a page exhibiting the broken behaviour? (did you check if the deployed preview fixes it?)

scholtzan:
Any page shows this behaviour: https://mozilla.cloud.looker.com/dashboards/1452?Event+Name=new_private_tab_tapped&App+Name=Firefox+for+Android&Window+Start+Time=28+days&Channel=release
For some tiles it shows the error: `Cannot use non-literal string filter for parameter field "event_name"`

Another thing I just noticed and fixed, some of the tiles are filtered on the event names that consist of `event_category.event_name` and some are just filtered on `event_name`. We need to be consistent here, so filtering on `event_category.event_name` would make the most sense here. This will require a small update to the dashboard for the tiles that currently expect `event_name` only (just need to point the filters to a different field).

Dexterp37:
> Any page shows this behaviour: https://mozilla.cloud.looker.com/dashboards/1452?Event+Name=new_private_tab_tapped&App+Name=Firefox+for+Android&Window+Start+Time=28+days&Channel=release For some tiles it shows the error: `Cannot use non-literal string filter for parameter field "event_name"`

Woh, I never stumbled on this :|

> Another thing I just noticed and fixed, some of the tiles are filtered on the event names that consist of `event_category.event_name` and some are just filtered on `event_name`. We need to be consistent here, so filtering on `event_category.event_name` would make the most sense here. This will require a small update to the dashboard for the tiles that currently expect `event_name` only (just need to point the filters to a different field).

Oh! Good catch!

Dexterp37 (approved):
Stealing review from @Iinh :-) @scholtzan would you kindly also adjust the dashboard filters?

PR #2079, opened 2024-01-22
Add a header anchor link to every header rendered from commentary

badboy:
Fixes https://github.com/mozilla/glean-annotations/issues/217

Completely unstyled for now, because I have no idea how to actually apply styling here.
I tried inline `<style>` but nothing seems to work.
Probably should make the headings black and underlined on hover.

cc @aminomancer 

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -13,6 +13,15 @@
         start ? `start="${start}"` : ""
       } class="mzp-u-list-styled">${body}</${outerEl}>`;
     },
+    heading(text, level, raw) {
+      const id = raw
+        .toLowerCase()
+        .trim()
+        .replace(/<[!/a-z].*?>/gi, "")
+        .replace(/ +/g, "-");
+
+      return `<h${level} class="annotation-header-link" id="${id}"><a href="#${id}">${text}</a></h${level}>\n`;
+    },
   };
   use({ renderer });
 
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -33,5 +33,26 @@
 {#if inline}
   {@html parseInline(text)}
 {:else}
-  {@html parse(text)}
+  <div>
+    {@html parse(text)}
+  </div>
 {/if}
+
+<style lang="scss">
+  div :global(h1 a),
+  div :global(h2 a),
+  div :global(h3 a),
+  div :global(h4 a),
+  div :global(h5 a) {
+    color: black;
+    text-decoration: none;
+  }
+
+  div :global(h1 a:hover),
+  div :global(h2 a:hover),
+  div :global(h3 a:hover),
+  div :global(h4 a:hover),
+  div :global(h5 a:hover) {
+    text-decoration: underline;
+  }
+</style>


src/components/Markdown.svelte
```diff
@@ -13,6 +13,15 @@
         start ? `start="${start}"` : ""
       } class="mzp-u-list-styled">${body}</${outerEl}>`;
     },
+    heading(text, level, raw) {
+      const id = raw
+        .toLowerCase()
+        .trim()
+        .replace(/<[!\/a-z].*?>/gi, '')
```
Iinh, line 20: do we need the `\/` escape?

```suggestion
        .replace(/<[!a-z].*?>/gi, '')
```

src/components/Markdown.svelte
```diff
@@ -13,6 +13,15 @@
...
+        .replace(/ +/g, '-');
+
+      return `<h${level} class="annotation-header-link" id="${id}"><a href="#${id}">${text}</a></h${level}>\n`;
```
Iinh, line 23: > Probably should make the headings black and underlined on hover.

what do you think if we add inline styling here:

```suggestion
      return `<h${level} class="annotation-header-link" id="${id}"><a href="#${id}" style="color: black; text-decoration: none;" onmouseover="this.style.textDecoration='underline'" onmouseout="this.style.textDecoration='none'">${text}</a></h${level}>\n`;
```

src/components/Markdown.svelte
badboy: Should have noted that:
I took this code from the recommended marked addon here: https://github.com/markedjs/marked-gfm-heading-id/blob/main/src/index.js (no need for a new dep for such simple functionality).

This removes everything that looks roughly like an HTML tag, that's why `/` is matched there as well.

src/components/Markdown.svelte
Iinh: yeah it's not the prettiest, but unfortunately since this is a dynamically generated component, any content generated by this function and subsequently rendered in the markup (HTML) will not automatically inherit CSS styles, unless they are explicitly scoped, like inline styling as i suggested here.

Iinh (approved):
thanks for working on this! r+ pending tests passing.

src/components/Markdown.svelte
badboy: yeah ... I was not happy with that. We should need inline JavaScript for styling. So I added styling now.

src/components/Markdown.svelte
```diff
@@ -13,6 +13,15 @@
         start ? `start="${start}"` : ""
       } class="mzp-u-list-styled">${body}</${outerEl}>`;
     },
+    heading(text, level, raw) {
+      const id = raw
+        .toLowerCase()
+        .trim()
+        .replace(/<[!/a-z].*?>/gi, "")
```
aminomancer, line 20: Idk what this is meant to remove (html tags?), but I wonder if you might have meant to put `trim()` after this (or both before and after), since if there is a string like `Some text <something>` this will return `some-text-`

badboy: See https://github.com/mozilla/glean-dictionary/pull/2079#discussion_r1482633233
This essentially is what GitHub also does.
`some-text-` is fine. `trim()` only removes whitespace and has no effect on `some-text-`

aminomancer: Moving `trim()` to after line 20 would remove the trailing space before it becomes a dash. I suppose you're right that GitHub's implementation is the same. Just did [a test](https://github.com/aminomancer/uc.css.js/blob/e1dd07a1a191d48d46b8436897a631a63a107fa5/README.md#header-test-) and it does yield a trailing dash. It's just a nitpick though. Thanks for working on this in any case!

badboy: oh THAT space. Sorry, yes, that would be removed. I don't mind either way, this is just to get "clean" slugs to use. github-compat was just the first thing I encountered and it seemed good enough.

PR #2084, opened 2024-01-25
Enable automatic page load instrumentation

Dexterp37:
This additionally removes a debugging mechanism based on environment variables.

### Pull Request checklist

      fixes, if applicable)


rosahbruno (approved):
LGTM!

Looks like we can do the same thing that we did for the DPV. We can keep both page load events running at the same time, compare the counts, and then remove the original after validating.

PR #2151, opened 2024-03-05
fix: unbreak linting and pytests

### LINKED ISSUE #2150: pytest >= 8 breaks tests (`PytestDeprecationWarning`)
```
site-packages\_pytest\nodes.py:159: PytestDeprecationWarning: <class 'pytest_black.BlackItem'> is not using a cooperative constructor and only takes {'parent', 'fspath'}.
  See https://docs.pytest.org/en/stable/deprecations.html#constructors-of-custom-pytest-node-subclasses-should-take-kwargs for more details.
    warnings.warn(
```

This is because our codebase relies on pytest-black, which appears to be stale (unmaintained). As a company we hit the same snag on other projects (see mozilla/mozanalysis#210). A solution was deployed there (use ruff), so let's try to do the same thing.

Dexterp37:
Fixes #2150 

Sadly, we can't bump Python to a more recent version because Netlify doesn't support Python > 3.8.

### Pull Request checklist

      fixes, if applicable)


badboy (approved):
ruff ruff ruff!

PR #2158, opened 2024-03-20
[Bug 1885857] Remove non standard page load events

abhi-agg:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1885857

### Pull Request checklist

      fixes, if applicable)

### Test

Tested the change locally by running `npm run dev` step mentioned in https://github.com/mozilla/glean-dictionary/blob/main/docs/development.md#development and observing [log pings](https://mozilla.github.io/glean/book/reference/debug/logPings.html) on browser's console. I don't see non-standard page load events anymore.

https://github.com/mozilla/glean-dictionary/pull/1015 was very handy as a reference in doing this task.


src/telemetry/metrics.yaml
```diff
@@ -1,44 +0,0 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
----
-$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
```
Dexterp37, line 6: Please check if this is going to break the probe-scraper

Dexterp37 (approved):
Please check whether or not these changes will break the probe-scraper before merging and deployinh them

src/telemetry/metrics.yaml
abhi-agg: I have a bug for this already https://github.com/mozilla/probe-scraper/issues/716. I will resolve that bug first and then merge it.

Dexterp37: I'd recommend trying to dry-run the probe-scraper before merging this PR

rosahbruno (commented):
Hmm.. I understand that we no longer have any Glean metrics that we are recording, but I feel like I am unsure about removing the Glean instrumentation entirely. Unless we are wanting to showcase collecting telemetry without actually needing to do anything with the Glean files themselves.

@Dexterp37
What are your thoughts here?

Dexterp37:
> @Dexterp37 What are your thoughts here?

I'm torn about this: on one end, leaving code around for no purpose is bad. On the other end, we might end up adding more context to events soon-ish, so we might as well leave it around (we'll need to add most of this back if we want to add a lifetime: application metric)

rosahbruno (changes requested):
Let's continue to remove the code that calls the page load events, but for now lets not take out the Glean metrics and set up entirely.

abhi-agg:
@rosahbruno @Dexterp37 I have made the required changes as per the discussion above 👍🏾

src/telemetry/metrics.yaml
abhi-agg: As we are leaving the metrics and pings in the tree, I believe we don't need to do this.

src/telemetry/index.js
```diff
@@ -8,10 +8,6 @@ import GleanMetrics from "@mozilla/glean/metrics";
 
 import { googleAnalytics } from "./ga";
 
-// Glean generated metrics and pings files
-import * as pageMetrics from "./generated/page";
```
Dexterp37, line 12: We still need to remove the metrics and pings from the YAML files, but leave the yaml files there (without metrics)

Dexterp37 (changes requested):

Dexterp37 (approved):

PR #2164, opened 2024-03-26
[Bug 1883593] Instrumentation to capture more clicks via glean

abhi-agg:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1883593
### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/src/components/AuthenticatedLink.svelte
+++ b/src/components/AuthenticatedLink.svelte
@@ -2,9 +2,11 @@
   import tippy from "./tippy";
 
   export let href;
+  export let label = undefined; // a prop for glean click events
+  export let type = undefined; // a prop for glean click events
 </script>
 
-<a {href}><slot /></a>
+<a {href} data-glean-label={label} data-glean-type={type}><slot /></a>
 <a
   class="lock-link"
   href="https://docs.telemetry.mozilla.org/concepts/gaining_access.html"
--- a/src/components/AuthenticatedLink.svelte
+++ b/src/components/AuthenticatedLink.svelte
@@ -2,8 +2,8 @@
   import tippy from "./tippy";
 
   export let href;
-  export let label = undefined; // a prop for glean click events
-  export let type = undefined; // a prop for glean click events
+  export let label = ""; // a prop for glean click events
+  export let type = ""; // a prop for glean click events
 </script>
 
 <a {href} data-glean-label={label} data-glean-type={type}><slot /></a>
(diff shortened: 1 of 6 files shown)


src/components/AuthenticatedLink.svelte
```diff
@@ -2,9 +2,11 @@
   import tippy from "./tippy";
 
   export let href;
+  export let label = null; // a prop for glean click events
+  export let type = null; // a prop for glean click events
 </script>
 
-<a {href}><slot /></a>
+<a {href} data-glean-label={label} data-glean-type={type}><slot /></a>
```
rosahbruno, line 9: I am not 100% sure how Svelte works with props passing. But, if there is a way to only set those properties if we are guaranteed to have a value for them (so they aren't incorrectly set as undefined or something) then that would be good. Unless Glean.js is checking to make sure those values aren't undefined? I forget how the Glean.js implementation looks

abhi-agg:
Okay. I am marking this PR read to review. This PR adds instrumentation to capture clicks on:

1. Metric details page: Clicks on `Access` section (bottom most) and `Sampling Configuration Snippet ` row inside `Metric sampling` section

2. App details page: Clicks on `Metrics`, `Pings`, `Tags`, `Application IDs` and `Third-Party Data` tabs

I have tested this change locally and it works as expected.

src/pages/AppDetail.svelte
```diff
@@ -91,15 +91,17 @@
...
+      <Tab key="tags" label="AppDetail.TagsTab">Tags</Tab>
     {/if}
-    <Tab key="app_ids">Application IDs</Tab>
+    <Tab key="app_ids" label="AppDetail.ApplicationIdsTab">Application IDs</Tab>
     {#if app.app_name === "firefox_ios"}
       <!-- for now, only Firefox iOS uses third-party data collection -->
-      <Tab key="third_party_data">Third-Party Data</Tab>
+      <Tab key="third_party_data" label="AppDetail.ThirdPartyDataTab"
+        >Third-Party Data</Tab
+      >
```
travis79, line 104: Formatting here looks weird, is this what the linter/formatter wanted?

travis79 (approved):
Looks good to me, with one minor question/nit about formatting in one place.

src/components/AuthenticatedLink.svelte
rosahbruno: Yep, sounds good 👍

src/pages/AppDetail.svelte
abhi-agg: Yeah. This is the final output of running formatter and linter.

PR #2196, opened 2024-04-17
Bug 1887694 - Generate events_stream SQL for events

### LINKED ISSUE #2185: Change autogenerated SQL to use `events_stream` tables
Instead of using the raw events tables.

Dexterp37:
Fixes #2185 

### Pull Request checklist

      fixes, if applicable)


src/pages/MetricDetail.svelte
```diff
@@ -54,7 +54,11 @@
     additionalInfo
   ) {
     if (metricType === "event") {
-      return getGleanEventQuery(table, additionalInfo);
+      // Although events might come from other pings, we override that and
+      // generate SQL just for the `events_stream`. This changes
+      // `some_database.table` to `some_database.events_stream`.
```
Dexterp37, line 59: I'm not really sure if we want to retain the old SQL code generation. Does it make any sense? This is quite a hack @badboy

src/data/gleanSql.js
```diff
@@ -15,25 +15,23 @@ LIMIT 10`;
 
 export const getGleanEventQuery = (table, additionalInfo) => `
   -- Auto-generated by the Glean Dictionary.
-  -- https://docs.telemetry.mozilla.org/cookbooks/accessing_glean_data.html#event-metrics \n
+  -- https://docs.telemetry.mozilla.org/cookbooks/accessing_glean_data.html#event-metrics
 
 WITH events AS (
   SELECT
     submission_timestamp,
```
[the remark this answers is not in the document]
badboy: Yes. Seems like the better idea.

src/pages/MetricDetail.svelte
badboy: For events that are not in an events ping this will be incorrect with this change, because they don't end up in the stream just yet.

But otherwise it's fine; no point in keeping the legacy way around

badboy (approved):
small thing, we could fix that in a followup as well.

src/pages/MetricDetail.svelte
Dexterp37: Ok, what about this? We retain the old query for non 'events' pings until this gets fixed.

badboy (approved):

PR #2246, opened 2024-06-03
[DENG-3417] Show link to Looker event monitoring dashboard for all metrics with type=event

scholtzan:
See https://mozilla-hub.atlassian.net/browse/DENG-3417

We want to show links to the event monitoring dashboards for all metrics with type event, not just for those coming from the event ping.


Dexterp37 (approved):
Thank you very much for your work here, @scholtzan !

LadyE215 (commented):
[null]()

PR #2285, opened 2025-02-05
DENG-7692 - feat: Add support to auto-generated click events

edugfilho:
### Pull Request checklist

      fixes, if applicable)


edugfilho:
CI is not passing yet because circleci can't access BQ to run `build-metadata` (BQ is needed after this PR)

etl/glean_auto_events.py
```diff
@@ -23,9 +23,12 @@ def _query_bigquery(query):
...
 def _get_auto_events_names(app):
     """Get the automatic events names for the app"""
-    query = f"""SELECT *
-                FROM `moz-fx-data-shared-prod.glean_dictionary_derived.auto_events_metadata`
-                WHERE app = '{app}' """
+    query = f"""SELECT
+                    app,
+                    name
+                FROM `moz-fx-data-shared-prod.glean_dictionary_derived.auto_events_metadata_v1`
+                WHERE app = '{app}'
```
akkomar, line 30: Should this also filter by `submission_date`?

edugfilho: Not really, we want distinct pairs of (app, name) regardless of date. The table is clustered on those columns so I think it'll be fine

etl/glean_auto_events.py
```diff
@@ -0,0 +1,48 @@
...
+def get_auto_events_for_app(app):
+    """Get the automatic events for the app"""
+    event_names = _get_auto_events_names(app)
+    auto_events = []
+    for row in event_names:
+        auto_event_id = row.name.split(".")[-1]
+        event_template = _auto_event_template.copy()
+        event_template["name"] = row.name
+        event_template["description"] = (
+            f"A event triggered whenever the {auto_event_id} element is clicked on a page."
```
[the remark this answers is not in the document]
edugfilho: That will be added later because they are less straightforward. For example, we are considering a count threshold after which the pageload auto-event will be shown on Glean Dictionary.

.circleci/config.yml
```diff
@@ -28,24 +28,24 @@ jobs:
       - run: npm test
   backend-build-and-test:
     docker:
-      - image: python:3.8
+      - image: python:3.9
```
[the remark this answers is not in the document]
edugfilho: Yes I'd prefer doing that update to 12 on a different PR, after I see this working, in order to not have to re-test stuff to ship this. I preferred to update to 3.9 because it was easier to use the google libraries for BQ compared to 3.8. 
And thanks for the heads up, I'll update `netlify.toml` and other files that reference python3.8

etl/glean_auto_events.py
akkomar: I see a risk of some "trash" data surfacing here and staying forever. I haven't checked if this does happen, but we could for example limit this to last 30 days or use some count threshold. Can be changed later though, I agree.

Re. sanitization, I mentioned some in another PR: https://github.com/mozilla/bigquery-etl/pull/6987#pullrequestreview-2609373141

akkomar (approved):

PR #2295, opened 2025-05-20
Use a local git checkout of the MPS repo instead of fetching from the CDN

### LINKED ISSUE #2294: build-metadata runs into GitHub rate limit
The [new GitHub rate limits](https://github.blog/changelog/2025-05-08-updated-rate-limits-for-unauthenticated-requests/) are much more strict.

As per https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28 it's:

> The primary rate limit for unauthenticated requests is 60 requests per hour.

And I guess that applies to raw.githubusercontent.com too, as we're getting a 429 after a while.
Now there's only two separate repositories we get from the CDN:

* The looker namespaces: https://github.com/mozilla/glean-dictionary/blob/a5cc7c476b1e188bf751ef2407912e10eb7f5e88/etl/glean_etl.py#L26
* bq schemas for all pings: https://github.com/mozilla/glean-dictionary/blob/a5cc7c476b1e188bf751ef2407912e10eb7f5e88/etl/glean_etl.py#L358-L361

The first would be fine, we request it exactly once, early.
The latter ... is a lot.
But we don't _need_ a request per file! We could just clone the whole repository on that branch and read the file from disk.

badboy:
This avoids rate-limiting on the CDN and might even be faster for full app runs.

Fixes #2294

### Pull Request checklist

      fixes, if applicable)

### DIFF
--- a/etl/glean_etl.py
+++ b/etl/glean_etl.py
@@ -1,8 +1,12 @@
 import copy
 import os
 
+import git
+import json
 import requests
+import shutil
 import stringcase
+import tempfile
 import yaml
 
 from .bigquery import get_bigquery_column_name, get_bigquery_ping_table_name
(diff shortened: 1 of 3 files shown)


badboy:
This might actually be slightly faster on CI too: old runtime for the `Build metadata` step was always somethings between 2min 45s and 3min 05s. This run now is 2min 4s

edugfilho (approved):
LGTM. Thanks so much for this!

PR #2297, opened 2025-06-03
Improve filtering speed by caching the search index

Standard8:
Currently, on my Firefox, when I visit https://dictionary.telemetry.mozilla.org/apps/firefox_desktop?page=1 and filter the metrics, there's a noticeable lag between typing and the display updating. Profiling this shows jank that can cover half a second when typing multiple characters.

When the filtering was [first implemented](https://github.com/mozilla/glean-dictionary/commit/52c128a7a1ee7abf5f062a5acfe0de8d9ea13904), `ItemList.svelte` created its own `searchIndex` and used that its `fullTextSearch` function. However, when the code was re-arranged in 80cb86f2106a7d1e79607abd9aeb3663f498933f, the `fullTextSearch` started to create a new `searchIndex` for every search. It is now processing ~6400 items each time.

Hence, this PR is using the `searchIndex` created in `ItemList.svelte` as a cache to avoid the re-generation. With this change the jank is limited to an initial ~170ms, and then all interactions are very responsive, with no more jank.

### Pull Request checklist

- [X] The pull request has a descriptive title (and a reference to an issue it
      fixes, if applicable)
- [X] All tests and linter checks are passing
- [X] The pull request is free of merge conflicts

### DIFF
--- a/src/components/ItemList.svelte
+++ b/src/components/ItemList.svelte
@@ -1,6 +1,5 @@
 <script>
   import { chunk, escapeRegExp } from "lodash";
-  import { Document } from "flexsearch";
 
   import { getItemURL } from "../state/urls";
   import { stripLinks } from "../formatters/markdown";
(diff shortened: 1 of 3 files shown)


chutten (approved):
If this were mozbuild python, I'd suggest merely adding `@memoize`, but alas we must do this the hard way.

PR #2325, opened 2026-02-17
Resolve library dependency pings/metrics/tags in MCP server

akkomar:
This should fix the issue reported in https://mozilla.slack.com/archives/C07MNJ1Q3DW/p1770997453481839?thread_ts=1770294328.866299&cid=C07MNJ1Q3DW

The MCP server previously only queried an app's own probeinfo endpoints, missing metrics and pings defined by library dependencies (e.g. baseline ping from glean-core). This adds dependency resolution that fetches library-variants mappings, resolves each app's dependencies to v1 API names, and merges library data with app data.


edugfilho (approved):
lgtm

